PR #3, opened 2016-08-16
Easier way of localization languages can now easily be added

chuparCh0pper:
outsourced pokemon names into strings.xml


farkam135:
This is where knowing android comes in handy. A lot easier thanks!

PR #16, opened 2016-08-16
Refactored MainActivity

chuparCh0pper:
extraced method "checkPermissions" & finalized not changing parameters


chuparCh0pper:
pokefly & MainActivity need more outsourcing pokefly has over 800 lines :grimacing:, but I'm of for tonight, see you tomorrow

PR #18, opened 2016-08-16
fixed cn translation

chuparCh0pper:
I'm not sure if this will work as tessdata is "eng.traineddata".

Also this line has to be altered https://github.com/farkam135/GoIV/blob/master/app/src/main/java/com/kamron/pogoiv/MainActivity.java#L111


farkam135:
not sure if I want to include chinese since just the cn.traineddata is about 80mb. Thoughts?

chuparCh0pper:
maybe we can outsource the traineddata and people can download it in the app, would make the app smaller aswell

farkam135:
Alright sounds good.

Issue #24, opened 2016-08-16
Narrow down pokemon identification with candy


nahojjjen:
An idea I've discussed a few times on reddit, is to narrow down the possibilities of what pokemon the user is looking at by watching candy names.

By doing this, you can narrow down the pokemon-picker to a single evolution line.

Example: A user has renamed a weedle to "snores", and the app erroneously recognizes it as a snorlax. If the app instead read that the pokemon wants to eat "weedle candy", then the options of possible pokemon are instead weedle, kakuna and beedrill. The app could then instead pick whichever of those 3 alternatives are closest to "snores", and when the user then corrects the input with the dropdown menu, the menu should only contain the 3 possible options, instead of all 151 in alphabetical order.

What would need to be done? My suggestions:
The application would need to ocr the candy name. If we in the Pokemon object allow another field which stores the "next evolution", some logic could solve the rest.

For example
        pokemon.add(new Pokemon("pokemon name", "evolution name", 110, 76, 50));

So an algorithm that figures out an evolution chain could simply check the candy name, for example bulbasaur, and then see if that pokemon has a non-null evolution name.

agrajaghh:
what is with OCR errors? How to correct them without displaying the full 151?

nahojjjen:
Reopened because it's still not narrowing down the dropdown list

Eccenux:
I see there were some changes in `showInfoLayout`. It was better now its worse ;-).

If you changed the name of your pokemon this will not give you expected result:
```java
            if (possiblePoke[1] < pokeCalculator.get(possibleCandy[0]).getDistance(origPokemonName)) {
                pokemonList.setSelection(possiblePoke[0]);
            } else {
                pokemonList.setSelection(possibleCandy[0]);
            }
```
`possiblePoke` is best match to any pokemon in pokedex
`possibleCandy` is best match of candy name in pokedex 

For pokemon named "Blah", candy "Mr. Mime"
`possiblePoke` would be nonsens.
`possibleCandy` would be an actual match.

`possibleCandy` is good for all non-evolved pokemons. It might not be good for evolved pokemon, but still probably better if `possiblePoke[1]` (distance to any known pokemon) is high.

So this would be better:
```java
            if (possiblePoke[1] < possibleCandy[1]) {
                pokemonList.setSelection(possiblePoke[0]);
            } else {
                pokemonList.setSelection(possibleCandy[0]);
            }
```

Eccenux:
[...]

PR #25, opened 2016-08-16
Resources over Hard coding data

McAJBen:
I moved the hard coded pokemon base stats to be in a resource file.


nahojjjen:
I like the idea and i think the pull should be accepted, but it does slightly inconvenience #24 since that issue requires an extra field to be added

nahojjjen:
Okay why did you close this so suddently

McAJBen:
When I transfered to resource file I might have missed something, the values might be off by one

nahojjjen:
I see

McAJBen:
I'll re-open after I make sure it's fixed

PR #28, opened 2016-08-16
Resources over Hard coding data 2

McAJBen:
I moved the hard coded pokemon base stats to be in a resource file.
Fixed order of values.


farkam135:
Array names are mixed up, def is supposed to be atk, sta is def and atk is sta. Also any benefit in having the stats in a res? Sorta liked being able to see all the stats right next to the pokemon?

hlplulustg:
That british defence makes me crazy, replace with defense plz :-)

McAJBen:
Going from hard-coded to resources doesn't seem to give it any improvement or detriment.
I was trying to make it loop through all of the names, to remove that long list of pokemon### but all I could add was the integers since putting the strings in an array makes it a ton harder to translate.

Damn I did give them the wrong names...

Issue #21, opened 2016-08-16
outsource traineddata


chuparCh0pper:
traineddata should be outsource to the apk file is smaller

Eccenux:
Don't mind large apk that can be downloaded through wifi. Would mind large requests upon each app usage (if that is what you meant). Hence -1.

farkam135:
Not enough demand for this, closing for now.

PR #39, opened 2016-08-17
fixed translations

chuparCh0pper:
Sorry for submitting empty strings.xml last time


nahojjjen:
Just a question, isnt the tesseract told to only recognize the "normal" alphabet right now? 
`tesseract.setVariable(TessBaseAPI.VAR_CHAR_WHITELIST, "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789/♀♂");`

So even if there's japanese language support for pokemon names, it cant recognize them...?

chuparCh0pper:
Yes @nahojjjen you are right, right now only ASCII symbols (and ♀♂) are supported. But, as mentioned [here ](https://github.com/farkam135/GoIV/issues/21) we plan on downloading the specific traineddata for each language.

PR #44, opened 2016-08-17
regenerated gradle files and updated to the latest build tools

harkin:
Android Studio prompted for the update from `alpha-6` to `beta-1`

The gradle files were changed by the task I added, running `./gradlew wrapper` from the command line


farkam135:
Updated in Crashlytics merge

PR #45, opened 2016-08-17
added .idea to the .gitignore and cleaned up that folder

harkin:
This whole folder is auto generated and will vary on different machines. Best to leave it out entirely


harkin:
heh, should've checked the open PRs. Someone beat me to it https://github.com/farkam135/GoIV/pull/42

Issue #46, opened 2016-08-17
Support Android 4.4


amra:
Could you support Android version 4.4. I would like to use this great app.

hlplulustg:
Already underway :-)

farkam135:
Supported in latest version 2.0.0

PR #48, opened 2016-08-17
support Android 4.4

chuparCh0pper:
I think there is no reasin supporting lower than 4.4 as pokémon go itself requires android 4.4
https://github.com/farkam135/GoIV/issues/46

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,7 +6,7 @@ android {
     buildToolsVersion "24.0.1"
     defaultConfig {
         applicationId "com.kamron.pogoiv"
-        minSdkVersion 21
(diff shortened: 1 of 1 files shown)


hlplulustg:
There are mods (adjusted minsdk only) for pokemon go on older versions which are running fine. If lower support doesn't require further adjustments I'd consider it.

farkam135:
Currently the screen capture api is only supported by 5.0+, a commit is coming soon that adds support for 4.4 as well as a lot of changes to accommodate it. So i'm also going to wait before merging any pull requests since they might cause conflicts with the new version.

PR #52, opened 2016-08-17
Revert "Added resources across languages for future translations."

chuparCh0pper:
This reverts commit [88e93a959b991cd841e086e4e97035ee416830c1](https://github.com/farkam135/GoIV/commit/88e93a959b991cd841e086e4e97035ee416830c1).

> please revert this. It is unnecessary, Android pulls the information from string.xml if e.g. strings-de.xml does not contain the element.
> 
> Also there is a nice translation view where you can see all elements in every language.
> http://tools.android.com/_/rsrc/1408658785758/recent/androidstudio087released/translations4.png


farkam135:
Sorry, still new to android! I meant to put this just so it's setup for translations when the time comes. Are you saying remove it because with the translation view we are able to see what we are missing?

chuparCh0pper:
Exactly, the translation view shows you which strings have been translated and which haven't.

farkam135:
Got it, reverting now, thanks!

PR #72, opened 2016-08-18
strings.xml, eevee, README

Dreckiger-Dan:
put a lot of strings into the `strings.xml` file

changed the way eevee names are handled (replace Rainer etc. with Eevee)


nahojjjen:
`.replace("Sparky", getString(R.string.pokemon133)).replace("Rainer", getString(R.string.pokemon133)).replace("Pyro", getString(R.string.pokemon133));`

They shouldnt all replace with 133 (vaporeon?) I think it should be

`.replace("Sparky", getString(R.string.pokemon134)).replace("Rainer", getString(R.string.pokemon133)).replace("Pyro", getString(R.string.pokemon135));`

Dreckiger-Dan:
No, pokemon133 is eevee. I think it is better to use eevee instead of the evolved pokemon (vaperon etc.), because you rename your pokemon before evolving (while it is still eevee) so you don't forget renaming it (at least my colleagues and I do it like this) and then you want to check if it is worth evolving every once in a while. If we implement it like you suggested, then you had to change the selected name every time.

nahojjjen:
Okay, i guess that makes sense

PR #73, opened 2016-08-18
Refactoring to make pokefly more readable 

nahojjjen:
This commit takes the pokemon IV analysis logic from pokefly and moves it to a class "PokeInfoCalculator". 

So for example, pokefly no longer creates its IV text in a huuuge blob. It instead asks for an IVScanResult for a pokemon, and then populates the string with information from the scan result.


hlplulustg:
It's showing the wrong pokemon name for the scanned one (Mew). Calculations still correct.

hlplulustg:
[https://cdn.discordapp.com/attachments/215438329024348161/215884739129769984/Screenshot_20160818-192816.png](https://cdn.discordapp.com/attachments/215438329024348161/215884739129769984/Screenshot_20160818-192816.png)

hlplulustg:
Was a user error (renamed pokemon), seems to be fine.

nahojjjen:
I killed this branch to make a new one without conflicts, see #75

PR #76, opened 2016-08-18
Selection of the pokemon base on the candies instead of the name of the pokemon

oliverlj:
Hello,

This is a new request to select a pokemon base from the candies instead of the name of the pokemon when is not found.

- Please check the rectangle and color of the ocr (I let a TODO).
- I've check if the name of the pokemon is perfect to a pokemon name. will be better to allow a little of tolerance.

Oliver, Pogo trainer.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -83,6 +83,7 @@ public class MainActivity extends AppCompatActivity {
     private boolean readyForNewScreenshot = true;
 
     private String pokemonName;
+    private String candyName;
     private double estimatedPokemonLevel;
     private int pokemonCP;
     private int pokemonHP;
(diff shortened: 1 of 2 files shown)


farkam135:
Awesome! Before i merge this, what do you mean by check rectangle and color? Is this like a double check to make sure values are accurate?

oliverlj:
Yes, by the check, I mean values are accurate.

- I calculate rectangle size with a Pidgey, for the width, I put half of the screen, maybe it is oversizes. When I debug on my one plus one, it extract  "BONBONS ROOCOUL I" (I'm french).
- For the color, I copied paste code from pokemon name and didn't change it. I don't know what it is used for.

oliverlj:
I just notice a bug according to english version pictures : 

candy pokemon name is in first (DRATINI CANDY) so the split should choose the first word. In french, it is the second word (BONBONS MINIDRACO) !

PR #81, opened 2016-08-19
Added auto updater

jehy:


jehy:
Sorry, here was some trash, gonna clean up and resubmit

Issue #10, opened 2016-08-16
Feature request: refine / narrow down option


hlplulustg:
A refine option for narrowing down the possible iv combinations is needed to be independent from any web calculators.

I don't know what the best idea would be to implement this. As of now I simply thought of local storage with past calculations and adding a third button 'refine' under 'cancel' and 'check iv'. When clicking the 'refine' button a dropdown list of past calculations for that specific species appears and the apropriate one can be chosen crosschecking the results after. This method would lay the foundation of another possible feature in the future - an iv database of your pokemon within the app. May be a stupid solution :-D

nahojjjen:
Ive got a version somewhat working logic-wise, but since I have nowhere to put it in the GUI im gonna ask stopyourshenanigains about it.

Here's an image of its current implementation in action: https://cdn.discordapp.com/attachments/211357346910175232/216268156610543617/Screenshot_20160819-205209.png

hlplulustg:
Closing in favor of #153. May be reopened if solution is not satisfying.

PR #91, opened 2016-08-19
Adds IV possibility refinement logic by leveling up & some crash fixes

nahojjjen:
see issue #10 
The result is not displayed, i commented out the part which actually puts it in the result string, since we dont want to accidentally put this in release until the results can be displayed in a better way.

if you wanna see it in its current form, un comment the block under 
` //TODO if you wanna work on the placement of the refinement (issue #10) then un-comment this code!`
in pokefly


nahojjjen:
Ive also fixed 2 of the more common crashes according to crashlytics (of the ones i didnt solve in my last pull request)

sadyc1:
The gastly crash is really annoying, hopefully it gets merged soon

nahojjjen:
Im redoing the pull from a new fork to avoid messing with the merge conflicts

PR #97, opened 2016-08-20
Improve efficiency of evolution tracking

sarav:
Associating a pokemon and its evolutions was using an O(n^2) algorithm.
For 151 pokemon, that's 22801 iterations! Changing it into O(n)
algorithm. So, the pokemon list population should take less than 1% of
the previous time.

Since every evolution comes after the pokemon in the pokedex, we can
immediately associate the evolved pokemon to its devolution. To do this
the evolution list had to be converted into a Pokemon list. This has
some cascading code change impact, but those make the code more readable
and efficient too (removes a lot of unnecessary int to Pokemon lookups).

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
+++ b/app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
@@ -37,37 +37,19 @@ public class PokeInfoCalculator {
         pokedex = new ArrayList<>();
         int pokeListSize = names.length;
         for (int i = 0; i <= pokeListSize - 1; i++) {
-            pokedex.add(new Pokemon(names[i], i, attack[i], defense[i], stamina[i], devolution[i]));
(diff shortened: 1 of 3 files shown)


sarav:
I tested the changes with Charmander and Eevee and it works fine. Btw, no idea why the original code had the evolutions in reverse-alphabetical order when it generated the CP range text. I just made it alphabetical since it was simple code.

farkam135:
Awesome change, pretty bad I didn't bother looking into how the evolutions were implemented before merging so thank you for catching that and making the algorithm more efficient!

sarav:
Finally got around to making a change instead of just talking about it! Thanks for pulling it in quickly.

Next attempt is going to be Google Now integration. We'll see how much easier it is with the latest clean up.

hlplulustg:
> Since every evolution comes after the pokemon in the pokedex

Keep in mind that this changes with later generations. May be better to set evolutions in a seperate loop than fixing in when gen 2 kicks in.

PR #98, opened 2016-08-20
Add Google Now support when running in battery saver mode

sarav:
It should be possible to change this in the future (with some code
refactoring) to add the capability to use Google Now without actively
having to start/stop the GoIV service.

Technically any fullscreen viewer can be used to share and analyze
existing screenshot files, but the user will have to update the trainer
level in the app to match their level when the picture was taken.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/ShareHandlerActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/ShareHandlerActivity.java
@@ -1,0 +1,37 @@
+package com.kamron.pogoiv;
+
+import android.app.Activity;
+import android.content.Intent;
+import android.graphics.Bitmap;
+import android.net.Uri;
+import android.provider.MediaStore;
+import android.support.v4.content.LocalBroadcastManager;
+import android.os.Bundle;
+
+import java.io.IOException;
+
+public class ShareHandlerActivity extends Activity {
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_share_handler);
+        finish();
+
+        Intent intent = getIntent();
+        String action = intent.getAction();
+        String type = intent.getType();
+
+        if (Intent.ACTION_SEND.equals(action) && type != null && type.startsWith("image/")) {
+            Uri imageUri = (Uri) intent.getParcelableExtra(Intent.EXTRA_STREAM);
+            try {
+                Bitmap bitmap = MediaStore.Images.Media.getBitmap(this.getContentResolver(), imageUri);
+                Intent newintent = new Intent("process-bitmap");
+                newintent.putExtra("bitmap", bitmap);
+                LocalBroadcastManager.getInstance(ShareHandlerActivity.this).sendBroadcast(newintent);
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
+        }
+    }
+}
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -54,6 +54,14 @@
             </intent-filter>
         </activity>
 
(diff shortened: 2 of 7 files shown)


sarav:
It is finally here! Works like a charm! Maybe @nahojjjen can make one of those demo videos for this.

nahojjjen:
I tested this mode while GoIV was on & in battery saving mode, and it worked. I wonder about what you mentioned in the original comment though - 

> without actively having to start/stop the GoIV service.

I tried sharing the information to the PoGO app through google now on tap while the application was not running in the background, and did not get a result popup. Would it be possible to fix so that sending a picture to GoIV while it's off still returns a result?

app/src/main/java/com/kamron/pogoiv/ShareHandlerActivity.java
```diff
@@ -0,0 +1,37 @@
...
+        String action = intent.getAction();
+        String type = intent.getType();
+
+        if (Intent.ACTION_SEND.equals(action) && type != null && type.startsWith("image/")) {
+            Uri imageUri = (Uri) intent.getParcelableExtra(Intent.EXTRA_STREAM);
+            try {
+                Bitmap bitmap = MediaStore.Images.Media.getBitmap(this.getContentResolver(), imageUri);
+                Intent newintent = new Intent("process-bitmap");
+                newintent.putExtra("bitmap", bitmap);
+                LocalBroadcastManager.getInstance(ShareHandlerActivity.this).sendBroadcast(newintent);
```
harkin: Either here or when you try to deserialise it I'd expect you to encounter [this exception](https://developer.android.com/reference/android/os/TransactionTooLargeException.html) sometimes. The maximum transaction size is ~1mb and a bitmap will get quite large on, say, QHD phones

sarav: @nahojjjen, the problem is that a lot of the other data in MainActivity would not be initialized. Like trainerlevel, arc calculation caching, tesseract, etc. So, until that's refactored, I've decided not to support this while the app is not running.

Also, after thinking aboit it, I'm not sure we should support this as it quickly starts giving misinformation. I myself scanned old screenshots using the share and I was getting good numbers. I then remembered that those screenshots were when I had a lower trainer level. So. for each share from a saved file, we'll need to add trainer level input.

At least for now or this commit, I didn't feel up to it.

sarav: @harkin, thanks for the heads up. I initially didn't want to serialize the bitmap in this intent because that's  lot of data to serialize and unserialize, but wanted to keep the receiving Intenthandler generic (so that it can handle files, etc) and this seemed nice. But it might be better to just pass the Uri and let the final intent get the bitmap.

Any idea if the actual bitmap associated with the Uri will still be kept around if I finish handling the this intent from Google Now? From, Google Now's perspective, once this function returns, it could consider the Intent handled. Why would it still be holding on to the bitmap instead of throwing it away?

I think I have to extract the bitmap here. I'm considering a wait and watch approach on this one. Will fix it if anyone reports the issue.

sarav:
@farkam135, I'm hoping you'll pull this in before the next release.

nahojjjen:
Okay that makes sense for now, but the user should get an error or something when they share the image with GoIV and its not running, so they dont think the app is not working.

app/src/main/java/com/kamron/pogoiv/ShareHandlerActivity.java
harkin: > Any idea if the actual bitmap associated with the Uri will still be kept around if I finish handling the this intent from Google Now? 

No idea I'm afraid; I'm not familiar with how it works at all. 

When I ran across that exception the way I fixed it was to write the bitmap to disk and pass a URI.  Thinking about it, I was dealing with larger bitmaps, say 4,000*3,000 pixels. This could be fine.

PR #101, opened 2016-08-20
made the gradlew file executable by the user

harkin:
I do a lot of things on the command line and it's good practice to use the wrapper instead of gradle directly for tasks 

With this change you can now do things like `./gradlew clean build` to build the project


Blaisorblade:
👍 , instant merge!

PR #109, opened 2016-08-21
Improve efficiency of evolution tracking and Google now support

sarav:
Fixed up my earlier pull requests and pushing them again.

I also addressed comments made on the earlier pull request for the evolution tracking. The code no longer assumes any relation between pokemon and their devolutions.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -54,6 +54,14 @@
             </intent-filter>
         </activity>
 
+        <activity android:name=".ShareHandlerActivity" android:theme="@android:style/Theme.Translucent.NoTitleBar">
+            <intent-filter>
+                <action android:name="android.intent.action.SEND" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data android:mimeType="image/*" />
+            </intent-filter>
+        </activity>
+
         <service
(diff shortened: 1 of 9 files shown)


sarav:
Both changes have been tested and seem to work fine.

sarav:
@nahojjjen In regards to your earlier comment that the app should say something if an image is shared but the GoIV service isn't running, I'd like to show a Toast show some message along those lines. But I'm not sure what's the proper way to get the actual text from the XML files and put in translations. Do you want to do that change? It could say something like "The GoIV service needs to be started first".

I also want to warn users of not trying to share an old screenshot and get bad IV numbers, but I don't know how to enforce or warn for that.

nahojjjen:
I'm one of those dirty programmers that consistently hard-code strings, so I dont have much knowledge in this field

sarav:
Also, "Minor text fixes" :smiley:

sarav:
@farkam135 I might not be using pull requests correctly, but looks like you missed [this commit](https://github.com/farkam135/GoIV/pull/109/commits/5a7204e296e79064513252e6e31361e1c2e7046d) because I updated the pull request.

sarav:
Nevermind, looks like you did. I'm just confused by github interface.

Issue #129, opened 2016-08-22
Candy identification broke Pokemon identification


sarav:
I'm on a fairly [recent commit](https://github.com/farkam135/GoIV/commit/99a1bf7b1f2b8f515c1b2201c86bbc2e7fe55fa0) for this post. A lot of pokemon are being recognized as the lowest devolutions that aren't even close to the fully evolved names.

* Victreebel is being recognized as bellsprout
* Primeape as Mankey
* Vileplume as Oddish.

I think this needs to be fixed before the next release.

Blaisorblade:
I tried to improve things with https://github.com/farkam135/GoIV/pull/123, but that wasn't enough because of OCR errors—see #120.

I don't have those Pokemons, so it'd be good if somebody having them set a breakpoint in `Pokefly.showInfoLayout`, looked at what happens and posted the OCRed names.

Relatedly, #130 is about reducing the cost for misrecognition.

------
In case you are interested/have time to try debugging, resuming our convo: Why did Android Studio refuse to install the app—can you try and post the exact error? I had it once because (it told me) I had to first uninstall the existing app.
BTW, I'm using Android Studio 2.1.3.

sarav:
The candy name based detection logic was very broken. I fixed it in #131 

Beedrill is a separate issue. OCR just does badly and make the name length match with Weedle instead of Beedrill.

mattaereal:
Trying with Gloom, with latest repo, debugging `showInfoLayout`. 

`
candyName: Oddish
pokemonName: Gloom
`

But when the pop up appears with the pokemon name, Oddish prevails.

mattaereal:
The `levenshteinDistance` is failing, or something is happening inside the for loop at PokeFly.getPossiblePokemon.

When iterating through the pokemon list, it stucks with number 43 (Oddish), and number 44 (Gloom) gets passed by. This has to do with the distance calculation. I'm still debugging.

mattaereal:
I think I found the bug.

possiblePoke = {43, 0} // Gloom
possibleCandy = {42, 0} // Oddish

```

            if (possiblePoke[1] < possibleCandy[1]) {
                pokemonList.setSelection(possiblePoke[0]);
            } else {
                pokemonList.setSelection(possibleCandy[0]);
            }

```

[...]

Issue #130, opened 2016-08-22
Decide what should be the order of pokemons in the dropdown after recognition


Blaisorblade:
Because of candy-based detection and issues with it (#129), pokemons are sometimes misrecognized as their base form. That wouldn't be so bad, *except* that the right Pokemon is far away in the list, because the list is in alphabetical order. Using Pokedex numbers this would be solved.

Blaisorblade:
There's a bunch of different ideas, so I'll try collecting them together to get some consensus?

Blaisorblade:
First, #131 is improving the base logic.

- In #125 @Eccenux implements what I originally suggested here, and @sarav suggests

> limiting the list to relevant pokemon? And then listing the rest? I think the rest should still be listed in alphabetical order to make them easy to find.

Based on #131, the list of relevant pokemons would be detected based on candy, and then the rest could be sorted by how well they match with the name.

- In #24 @nahojjjen, @Eccenux and me also discuss variants; I think the latest proposal there matches the above idea.

Eccenux:
I think best way would be to have pokedex sorting for now as it is always in evolution order currently.

But it would be great to make a combobox to be able to filter the dropdown (e.g. type in "fairy" to quickly get to Clefairy)... Of top of my head I don't remember if there is a standard filterable combobox in Android.

hlplulustg:
> I think best way would be to have pokedex sorting for now as it is always in evolution order currently.

Still that would only postpone a future solution. Evolution order by pokedex will change with Gen 2.

sarav:
I have a suggestion.

1. First pokemon is the detected pokemon.
2. The next set of pokemon are the ones with similar candy, in pokedex or alphabetical order among them.
3. The rest are sorted by name,

(1) -- because I don't want to scroll up to find more pokemon. I quickly lose track if I need to find the original pokemon again.
(3) I think once we do (1) and (2), sorted alphabetically is the most user friendly option. No one remembers the pokedex order.

@hlplulustg Can you give an example just so I'm confident about your claim?

Eccenux:
Currently (1) and (2) are true (as long as pokedex is sorted by index).

Doing (3) would require creating a new sorted list (currently Spinner is fed directly with pokedex). This would make the app consume more memory (not sure how much more). Memory consumption is important as this is the most likely reason for GoIV/PokemonGO to crash (at least on my 2GB phone).

hlplulustg:
[...]

PR #131, opened 2016-08-22
Fix candy name based Pokemon search

sarav:
If the pokemon name match isn't good, we can't just pick the pokemon
name that matches the candy name perfectly. If the candy name was
detected perfectly, then we'll always end up picking the base pokemon.
Instead, once we figure out the base pokemon with the best candy name
match, we have to go through all the pokemon with that candy name and
see how well they match with the detected pokemon name.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -20,6 +20,7 @@ import android.provider.MediaStore;
 import android.support.v4.content.LocalBroadcastManager;
 import android.text.Html;
 import android.util.DisplayMetrics;
+import android.util.LruCache;
 import android.view.Gravity;
 import android.view.LayoutInflater;
 import android.view.MotionEvent;
@@ -36,6 +37,7 @@ import android.widget.TextView;
 
 import java.text.NumberFormat;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 
 import butterknife.BindView;
@@ -103,6 +105,10 @@ public class Pokefly extends Service {
     private int pokemonCP;
     private int pokemonHP;
     private double estimatedPokemonLevel = 1.0;
+
+    private HashMap<String, String> userCorrections;
+    /* We don't want memory usage to get out of hand for stuff that can be computed. */
+    private LruCache<String, String> cachedCorrections;
 
     private PokemonSpinnerAdapter pokeAdapter;
 
@@ -186,7 +192,8 @@ public class Pokefly extends Service {
                 getResources().getIntArray(R.array.defense) ,
                 getResources().getIntArray(R.array.stamina),
                 getResources().getIntArray(R.array.DevolutionNumber));
-
+        userCorrections = new HashMap<>(pokeCalculator.pokedex.size());
+        cachedCorrections = new LruCache<>(pokeCalculator.pokedex.size() * 2);
     }
 
     @Override
@@ -467,7 +474,6 @@ public class Pokefly extends Service {
                 pokemonList.setBackgroundColor(Color.parseColor("#ffcccc"));
             }
 
-            pokemonName = pokeCalculator.get(possiblePoke[0]).name;
             pokemonList.setSelection(possiblePoke[0]);
             pokemonHPEdit.setText(String.valueOf(pokemonHP));
             pokemonCPEdit.setText(String.valueOf(pokemonCP));
@@ -495,6 +501,17 @@ public class Pokefly extends Service {
         int bestMatch = 100;
         int bestCandyMatch = 100;
         Pokemon p;
+
+        /* If the user previous corrected this text, go with that. */
+        if (userCorrections.containsKey(poketext)) {
+            poketext = userCorrections.get(poketext);
+        }
+
+        /* If we already did similarity search for this, go with the cached value. */
+        String cached = cachedCorrections.get(poketext);
+        if (cached != null) {
+            poketext = cached;
+        }
 
         /* If the pokemon name was a perfect match, we are done. */
         p = pokeCalculator.get(poketext);
@@ -544,6 +561,9 @@ public class Pokefly extends Service {
             }
         }
 
+        /* Cache this correction. We don't really need to save this across launches. */
+        cachedCorrections.put(poketext, p.name);
+
         /* Adding the candy distance and the pokemon name distance gives a better idea of how much
          * guess is going on. */
         int[] result = {p.number, bestCandyMatch + bestMatch};
@@ -558,6 +578,16 @@ public class Pokefly extends Service {
     private String getIVText() {
         int selectedPokemon = pokemonList.getSelectedItemPosition();
         Pokemon pokemon = pokeCalculator.get(selectedPokemon);
+
+        /* TODO: Figure out the Android way to save the user corrections
+         * But we'll have to set a size limit on that and throw away LRU entries.
+         * We should add Google backup support if we do. */
+        /* TODO: Move this into an event listener that triggers when the user
+         * actually changes the selection. */
+        if (!pokemonName.equals(pokemon.name)) {
+            userCorrections.put(pokemonName, pokemon.name);
+        }
+
         String returnVal = String.format(getString(R.string.ivtext_title), estimatedPokemonLevel, pokemonCP, pokemonHP, pokemon.name);
         returnVal += "<br>"; //breakline
         IVScanResult ivScanResult = pokeCalculator.getIVPossibilities(selectedPokemon, estimatedPokemonLevel, pokemonHP, pokemonCP);
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -450,8 +450,7 @@ public class Pokefly extends Service {
         if (!infoShownReceived) {
 
             infoShownReceived = true;
-            int[] possiblePoke = getPossiblePokemon(pokemonName);
-            int[] possibleCandy = getPossiblePokemon(candyName);
+            int[] possiblePoke = getPossiblePokemon(pokemonName, candyName);
             ivText.setVisibility(View.GONE);
             pokemonInfoLayout.setVisibility(View.VISIBLE);
             initialButtonsLayout.setVisibility(View.VISIBLE);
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -491,17 +483,73 @@ else if (possiblePoke[1] < 2) {
...
+                pokeNumber = p.number;
+                bestMatch = dist;
+            }
+
+            /* We found a perfect match. We are done. */
+            if (bestMatch == 0) {
+                int[] result = {pokeNumber, bestMatch};
+                return result;
+            }
+        }
```
Blaisorblade: I'm trying to wrap my head around this logic, and this first loop seems unnecessary though—any match found here should be found by the 2nd loop.

The only differences are:
- this loop might be faster, but I don't think this (undemonstrated) performance difference matters enough for this complexity.
- if we ever detect the perfect pokemon name match but the wrong candy (does that ever happen?), removing this loop introduces a bug.

sarav: You need to consider the case where neither are OCRed correctly. In that case, I'd want to go by the closest match. This loop is needed in my opinion.The loop itself has nothing to do with performance.

Blaisorblade:
`if (possiblePoke[1] < possibleCandy[1])` was introduced by @Eccenux in #125, based on the argument in https://github.com/farkam135/GoIV/issues/24#issuecomment-241289970. I disagree with that, but I'd like to slow down the ongoing edit war on that, so can we sit down and discuss this together? 😄 

The question is what happens if a pokemon is renamed to something else entirely—candy names are more reliable. But IIUC your logic works better even there—at worst, it will pick the wrong pokemon in the evolution family.
We still probably need to sort the list better—let's discuss that in #130.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: The zero check is the performance optimization that says that if you already found a perfect match, the candy search can't do any better.

Also, the optimization isn't hypothetical. The distance check is very expensive and in a mobile app, you need to consider these things for the sake of power too.

Blaisorblade: OK, so we need a loop—the zero check doesn't increase code complexity too much.

Eccenux:
If you pick one that best matches detected name from evolutions then I think this is fine. Probably just skip this checking if candy matched pokemon doesn't have evolutions.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -491,17 +483,73 @@ else if (possiblePoke[1] < 2) {
...
+                bestCandyMatch = dist;
             }
+            /* We found a perfect match. We are done. */
+            if (bestCandyMatch == 0) {
+                break;
+            }
+        }
+
+        /* If the pokemon name match was better than the candy name match, don't use candy name */
+        if (bestMatch < bestCandyMatch) {
```
Blaisorblade: After @mattaereal's debugging in #129, we should think whether we want `<=` here instead of `<`. At least, the scenario where `bestMatch` and `bestCandyMatch` are both 0 (which caused the original bug) is impossible here.

mattaereal:
If you're going to focus on finding a perfect match first, wouldn't it be better to use a hashmap instead? Also, if it's a perfect match the loop should break when found.

Blaisorblade:
> If you're going to focus on finding a perfect match first, wouldn't it be better to use a hashmap instead?

+100, that changes the lookup cost from cubic (n. pokemon * length OCRed string * avg. length of pokemon name) to linear in the length of the OCR string.

> Also, if it's a perfect match the loop should break when found.

Well, that happens already where it should, in the first loop, through `return`.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -491,17 +483,73 @@ else if (possiblePoke[1] < 2) {
...
+
+        for (Pokemon p: pokeCalculator.pokedex) {
+            int dist = p.getDistance(poketext);
+            if (dist < bestMatch) {
+                pokeNumber = p.number;
+                bestMatch = dist;
+            }
+
+            /* We found a perfect match. We are done. */
+            if (bestMatch == 0) {
```
Blaisorblade: @mattaereal this is where the first loop breaks.

sarav:
> If you're going to focus on finding a perfect match first, wouldn't it be better to use a hashmap instead? Also, if it's a perfect match the loop should break when found.

Not really "focusing" on finding the perfect match first. I'm just saying that we don't need to proceed with more useless work if we found the perfect pokemon name match.

Regarding hashmap, that can be a separate optimization patch, however I don't think it's correct. The OCR'd value might change each time. So, we can't store it. At best a hashmap reduces the check for the "candy match list". Which is not going to be large anyway (4 max). Doesn't feel worth the complexity.

mattaereal:
I don't think it would be very complex, with a BidiMap we can sort it out. It's a bijective mapping, whereas a pokemon id can return its name and otherwise. We can add this as a future enhancement, we don't know how many pokemons Niantic is going to add in the future, and unnecessary looping must be taken out.

mattaereal:
Oh, I forgot to add that this pull request fixes #129, already tested it out and works fine.

Blaisorblade:
> Regarding hashmap, that can be a separate optimization patch, however I don't think it's correct. The OCR'd value might change each time.

The correct hashmap fix builds a hashmap mapping Pokémon names to the Pokémons. Once. Then you can look up the OCRed name into that map instead of the first loop here.

sarav:
Ah, for the perfect OCR case? Sure, seems like a decent idea. We could can do that when we populate the pokemon I guess.

sarav:
Actually, I got some additional ideas on how to optimize this and improve detection by a lot. I'll put up a patch on top of this one later today. So, it'd be nice if others don't work on the same thing.

The app seriously need some aesthetics improvement though. Some material design would be nice.

farkam135:
Waiting for the improvements before merge!

Eccenux:
@sarav Not sure what you have in mind be if we can trust candy OCR, then maybe just find candy and __then__ search for name in evolutions? That would be O(n) and would satisfy most I guess (would find at least first evolution and would also find a "Gyrados" for names like "Gyrados 97").

BTW. One more thing came to me. As most people are adding numbers to original names then maybe just filter OCR-ed name from numbers and spaces. Would solve common use-cases.

Blaisorblade:
> Not sure what you have in mind be if we can trust candy OCR, then maybe just find candy and then search for name in evolutions?

@sarav was concerned about errors with the candy. Never seen that but might still happen.

> As most people are adding numbers to original names then maybe just filter OCR-ed name from numbers and spaces. Would solve common use-cases.

That's probably a good idea. There's an issue with letters recognized as numbers, and there are often not enough chars for spaces, but still.

sarav:
Yeah, since others are saying that candy OCR works better, I plan to filter by that first and then pick the best name match. 

Btw, technically the Pokemon detection is already O(n) if you consider the distance calculation as one unit of work. We are just trying to improve detection and a constant factor (say divide by 3) reduction by doing candy first because there are less possibilities and because OCR does better with candy name.

The hashmap idea is about short circuiting some of the calculations.

sarav:
Yeah, I don't plan to remove the numbers because of letters being recognized as numbers. Not sure how it'll affect the distance calculations.

mattaereal:
> maybe just find candy and then search for name in evolutions?
I thought of this, but I'm not planning on refactoring such a thing yet. Having returned a list of possible evolutions by candyname, and iterate between them, calculating distance, whatever.

Getting pokemonid from candyName is O(1) if you're using a hashmap, and then from that all the possible evolutions O(n) being n < 3 for now (that's the case of an Eevee for example).

sarav:
@farkam135 I'm done with the improvements I had in mind. I might do more, but this part of the code shouldn't be affected. Please pull this in.

sarav:
Added a cool "learn from the user" feature. This should allow us to handle Beedrill issues if we can save the learning across app launches. I haven't figure out the right way to do it. See TODO. Anyone wanna give that a shot?

sarav:
@Eccenux Do you want to take care of [this TODO](https://github.com/farkam135/GoIV/blob/35de21f1a12c5cdf3d7796d6f5a68610d86be5a6/app/src/main/java/com/kamron/pogoiv/Pokefly.java#L585)?

You seem more familiar with this stuff. You just need to the stuff in the if block only if the user clicked the selection list.

Without that, if the user sees Beedrill as detected as "Weedle" and doesn't fix it the first time, this assume the user added that correction.

Eccenux:
Hm... If I understand this correctly you assume someone names his pokemons the same for each type.

Not sure if that is common but I add attack DPS to names, so both Golduck and Vaperon can be named `water15/30`. So I'm not sure if `userCorrections` thing is a good idea. But if those corrections only work for searching within evolutions then I guess it's fine.

sarav:
Nope. Even without any renames.

sarav:
@nahojjjen Hey, you forgot to mention the learning from the user part in the reddit post! Once user makes a correction the app will remember it, at least for that run.

nahojjjen:
Was that already implemented for 2.2.0?

sarav:
Yes, but it didn't have the saving across launches part.

Blaisorblade:
FWIW: to avoid forgetting changes in the release notes, in some projects PRs worth mentioning have to add a changelog entry, so at the end you don't need to try to remember all that happened.

sarav:
What do you mean by changelog entry?

Blaisorblade:
A changelog entry is a bullet point of the release notes—any bullet point in https://www.reddit.com/r/GoIV/comments/4z5qxz/goiv_220_released/ is one.

I found a website explaining the idea in more detail: http://keepachangelog.com/

sarav:
Yeah, I know what a changelog is. I thought there was already a scheme that was used here that I didn't know about.

I'm not a huge proponent of maintaining a separate changelog when git log --pretty=oneline works well enough to get a good idea if people put in proper git commit lines. Git recommended approach is that the first line be one sentence that fits in 75-80 chars and describes the commit. The rest of the commit text can come after the commit line to explain it in detail. I try to follow that in my git commits.

Blaisorblade:
I haven't seen coding standards around—I started https://github.com/farkam135/GoIV/blob/master/DEVELOPMENT.md. The only thing not mentioned is "hey, there's this Discord chatroom where devs hang out", the one linked on our Reddit (https://discord.gg/y6BvF5D0).

Re Git: even if reviewing enforced standards on all PRs, `git log --pretty=oneline 2.1.0..2.2.0` has 130 entries and many make sense for devs but not for users (and some msgs could be better, yes).  Though maybe that could be a starting point for a manual changelog.

Anyway, these are the options, it's not up to me to pick what matters most.

sarav:
Oh, I didn't know about the chatroom. That link seems to have expired. Can I get another one?

Blaisorblade:
Here's one from the bosses for devs, I've been asked to not reshare on Reddit: https://discord.gg/NcCT7

sarav:
I got into the #dev channel, but there's no history and my messages just vanish into the ether. Anyway, I'll  try again later today.

PR #132, opened 2016-08-22
Update README.md

jehy:
explain need for internet permission.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,11 +85,9 @@ It's 'possible', but we don't think it will happen. Niantic could alter their ap
 
 ###Why does GoIV require X permission?
 
-Storage:
-To store the OCR module. (OCR - Optical character recognition, the thing that converts images of text to text.)
-
-Screen capture:
-To identify when to add the overlay, and to read the pokemon information.
+ - Storage - To store the OCR module (OCR - Optical character recognition, the thing that converts images of text to text);
+ - Screen capture - To identify when to add the overlay, and to read the pokemon information;
+ - Internet - send crash reports and self update. You can disable it in Android aplication settings section.
 
 ###What are IVs?
 


README.md
```diff
@@ -85,11 +85,9 @@ It's 'possible', but we don't think it will happen. Niantic could alter their ap
...
 ###Why does GoIV require X permission?
 
-Storage:
-To store the OCR module. (OCR - Optical character recognition, the thing that converts images of text to text.)
-
-Screen capture:
-To identify when to add the overlay, and to read the pokemon information.
+ - Storage - To store the OCR module (OCR - Optical character recognition, the thing that converts images of text to text);
+ - Screen capture - To identify when to add the overlay, and to read the pokemon information;
+ - Internet - send crash reports and self update. You can disable it in Android aplication settings section.
```
Blaisorblade: aplication -> application. But IIUC no, you can't, which is why you can build a separate APK without Internet permission.

Blaisorblade:
BTW, can we please all stop titling commits with `Update README.md`? Better would be
`README.md: explain need for internet permission`.
Trying to look at the file history is too hard. https://github.com/farkam135/GoIV/commits/master/README.md

README.md
jehy: That's patrly wrong, you can disable it in Android 6+.

Blaisorblade: Got Android 6 on Nexus 5, and I was asked about everything but Internet access, because it's not a dangerous permission — @nahojjjen explained me so also the official docs seem to agree:
https://developer.android.com/training/permissions/requesting.html
https://developer.android.com/guide/topics/security/permissions.html#normal-dangerous

jehy: Yup, sorry, my mistake. The whole thing about users afraid of internet permission is a simple paranoia because you know when app is doing screen capture but we have to deal with it. What a pity.

jehy:
Wrong update, see discussion in line notes.

PR #133, opened 2016-08-22
disable update checking for offline version

jehy:


app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -340,11 +340,14 @@ public void onStop() {
...
         super.onResume();
 
-        //if (BuildConfig.enableUpdater) {
-            //if (currentSettings.isUpdatesEnabled()) {
-                new AppUpdateLoader().start();
-                checkingForUpdate = true;
-    //}
+        if(BuildConfig.FLAVOR.equals("offline"))
+        {
+            Toast.makeText(mContext, "Offline version does not auto update", Toast.LENGTH_SHORT).show();
```
Blaisorblade: IIUC, this is a message for the user of the offline version every time he resumes the GoIV activity. It's a floating message designed to be unobtrusive, but that still seems too much noise to me. Or am I missing sth.?

jehy:
Toast is short and non intrusive. Also I think that it is important to inform user that he can be missing new features and updates :)

rhari991:
Still, it will popup every time the user switches to and from the app.
And I took care of this in a commit in my pr.

PR #134, opened 2016-08-22
Spanish localization and minor fixes

oskomonzon:
- Added Spanish locale resource file.
- Switched equals() to equalsIgnoreCase() for better localization support.
- Added missing "Accept Screen Capture" string to resource files.
- Replaced hardcoded "Stop" literal with R.strings.main_stop at the start of Pokefly.


Blaisorblade:
> Switched equals() to equalsIgnoreCase() for better localization support.

You mentioned this on Discord... but now that I think of it, why would you need that? All calls to setText that I see preserve the case.

oskomonzon:
> You mentioned this on Discord... but now that I think of it, why would you need that? All calls to setText that I see preserve the case.

I think the conflict is no longer present due to the cleaning of hard-coded literals. (The issue arraised when comparing the STOP vs Stop text). So that's no longer required, I guess.

Blaisorblade:
OK, I suspected as much. Could you please drop that commit then?

Blaisorblade:
Re double decimals: for the average that makes sense, but for the intermediate lines they're redundant—all decimal digits are equal to the last before the 0. In your first line, for instance, the number is 22/45*100 = 22/9 * 20 = 48.8888888... Since all IV percentages are some natural number * 20 / 9, that's always the case.

app/src/main/res/values/strings.xml
```diff
@@ -178,15 +178,15 @@
...
+    <string name="ivtext_iv">Min: %1$.2f%% Average: %2$.2f%% Max: %3$.2f%%\n</string>
     <string name="ivtext_evolve">Can be evolved into %1$s:\n</string>
     <string name="ivtext_evolve_further">Can be further evolved into %1$s:</string>
     <string name="ivtext_many_possibilities">There are too many possibilities for this pokemon. Try powering it up!</string>
     <string name="ivtext_max_lvl_cost">Cost to reach lvl %1$s:</string>
     <string name="ivtext_max_lvl_cost2">Candy: %1$s Stardust: %2$s</string>
     <string name="main_battery_saver">Battery Saver (Screenshot to Scan)</string>
     <string name="ivtext_cp_lvl">CP at lvl %1$s: %2$s - %3$s</string>
-    <string name="ivtext_stats">Atk: %1$02d Def: %2$02d Sta: %3$02d — %4$3d%%</string>
+    <string name="ivtext_stats">Atk: %1$02d Def: %2$02d Sta: %3$02d — %4$.2f%%</string>
```
Blaisorblade: Also, however many decimals we use here, it's also good to keep specifying a width to ensure alignment.

oskomonzon:
> Re double decimals: for the average that makes sense, but for the intermediate lines they're redundant—all decimal digits are equal to the last before the 0. In your first line, for instance, the number is 22/45*100 = 22/9 * 20 = 48.8888888... Since all IV percentages are some natural number * 20 / 9, that's always the case.

Do you think we should show just one decimal in that case then?

Blaisorblade:
> Do you think we should show just one decimal in that case then?

No decimals would be fine then—all the info is in the last IV digit. To make it simpler: the minimal change is more than 2% (1/45 * 100 = 2.22....%), so you don't need decimals there.

oskomonzon:
But we do have some cases where that's not true... Like the first and last prediction from the screenshot attached where:
Atk: 10, Def: 0, Sta: 12 = 48.89%
Atk: 06, Def: 07, Sta: 12 = 55.56%

Also, I find odd to show round numbers on the prediction and then show double digits on the Min/Max values.
Wouldn't be better to keep it for the sake of clarity?

Blaisorblade:
I know, but 48.89 is just 48.888888 rounded correctly to two digits.
Rounding it to 0 gives 49. Again, all accounted for in the math above,
though I didn't want to explain everything.
-- 
Paolo Giarrusso
From smartphone, sorry for typos or excessive brevity

oskomonzon:
You are right, @Blaisorblade. I'll revert to rounded percents on the list and just leave the Average as double. I hope everyone agrees :)

farkam135:
For the redesign can we just get all percentages rounded, thanks!

Blaisorblade:
> For the redesign can we just get all percentages rounded, thanks!

Why? There were some users requesting this, and for averages it can make sense no?

farkam135:
Only a handful a people care about that precision and it's such a minor detail. The inconsistency between the percentages is a greater hit to the experience than knowing the exact percentage of the Average, especially since people prioritize some values over others.

Blaisorblade:
So in the end the precision change was merged? 😕  See #151.

PR #140, opened 2016-08-23
Cache OCR results for parts of the bitmap that don't change often

sarav:
This pretty much excludes caching CP OCR since the background changes a
lot and is probably taking the most amount of time to OCR. But I think
this should still help improve the performance because I think the
hashBitmap() function should still take less time than all the replace
colors and tesseract code.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -11,6 +11,7 @@ import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.SharedPreferences;
 import android.graphics.Color;
 import android.graphics.PixelFormat;
 import android.net.Uri;
@@ -40,6 +41,7 @@ import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -68,6 +70,7 @@ public class Pokefly extends Service {
     private static final String KEY_SEND_INFO_LEVEL = "key_send_info_level";
     private static final String KEY_SEND_SCREENSHOT_DIR = "key_send_screenshot_dir";
 
+    private static final String PREF_USER_CORRECTIONS = "user_corrections";
 
     private final int MAX_POSSIBILITIES = 8;
 
@@ -81,6 +84,7 @@ public class Pokefly extends Service {
     private WindowManager windowManager;
     private DisplayMetrics displayMetrics;
     ClipboardManager clipboard;
+    private SharedPreferences sharedPref;
 
     private boolean infoShown = false;
     private boolean infoShownReceived = false;
@@ -193,7 +197,9 @@ public class Pokefly extends Service {
                 getResources().getIntArray(R.array.defense) ,
                 getResources().getIntArray(R.array.stamina),
                 getResources().getIntArray(R.array.DevolutionNumber));
+        sharedPref = getSharedPreferences(PREF_USER_CORRECTIONS, Context.MODE_PRIVATE);
         userCorrections = new HashMap<>(pokeCalculator.pokedex.size());
+        userCorrections.putAll((Map<String, String>) sharedPref.getAll());
         cachedCorrections = new LruCache<>(pokeCalculator.pokedex.size() * 2);
     }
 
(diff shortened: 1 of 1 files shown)


sarav:
@farkam135 Don't merge this yet.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -116,6 +118,7 @@
     private int pokemonHP;
     private boolean pokeFlyRunning = false;
     private int trainerLevel;
+    private LruCache<String, String> OCRCache = new LruCache<>(200);
```
Blaisorblade: 1. `OCRCache` is a class name—field names should be lowercase by conventions, so `ocrCache`.
2. Mapping from String is questionable.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -638,38 +651,65 @@ private void scanPokemon(Bitmap pokemonImage, String filePath) {
...
-            }else{
-                pokemonName = getResources().getString(R.string.pokemon032);
+        Bitmap name = Bitmap.createBitmap(pokemonImage, displayMetrics.widthPixels / 4,
+                                            (int) Math.round(displayMetrics.heightPixels / 2.22608696),
+                                            (int) Math.round(displayMetrics.widthPixels / 2.057),
+                                            (int) Math.round(displayMetrics.heightPixels / 18.2857143));
+        String hash = "name" + hashBitmap(name);
+        pokemonName = OCRCache.get(hash);
+        if (pokemonName == null) {
+            name = replaceColors(name, 68, 105, 108, Color.WHITE, 200);
```
Blaisorblade: Having `replaceColors` here might mean less cache matches, in general.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -625,6 +628,16 @@ private double getPokemonLevel(Bitmap pokemonImage){
...
         return 1;
     }
+
+    /**
+     * Get the hashcode for a bitmap
+     */
+    String hashBitmap(Bitmap bmp) {
+        int[] allpixels = new int[bmp.getHeight() * bmp.getWidth()];
+        bmp.getPixels(allpixels, 0, bmp.getWidth(), 0, 0, bmp.getWidth(), bmp.getHeight());
+        return Integer.toHexString(Arrays.hashCode(allpixels));
```
Blaisorblade: Why `toHexString`? That seems just a way to waste memory. `LruCache<int, String>` might not work, but `LruCache<Integer, String>` will.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: This is not a final patch, just to get opinion. It was initially called "pokenameOCRCache" and then I decided to combine it into one image cache instead of one per image type.

I used to use "Integer" before, but switched to string because I had to add the "name", "candy", "hp" prefix to the key to avoid the theoretical possibility of a hash collision between a name image and a hp image.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: Maybe, but it's also a lot less work -- replaceColors does a lot of floating point arithmetic. Also, at best you save a few bytes by getting two color different images to match. Not worth it IMHO. Thoughts?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: No good reason. Could have gone with .toString(). HexString was just easier to read on the logs (deleted that line before I uploaded). Otherwise I'd get "name: -245423".

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: > I had to add "name", "candy", "hp"

Ah I see. In a proper impl, it'd be cleaner and a bit more efficient to use a pair of the image hash and an enum for the key type, something like (pseudocode)
```
enum KeyType {NAME_KEY, CANDY_KEY, HP_KEY};
class Key { KeyType k; int imgHash; };
```
plus equals/hashcode methods, generated with an IDE. (I realize that's lots of code, which is why I prefer other JVM languages instead of Java).

sarav: Well, I won't want to hash them all together. If the same HP occurs in another pokemon, it might still get a cache hit.

But agree that using strings isn't the best way to add a "tag" to a hash.

Blaisorblade: Re generated equals/hashCode, since we're using Lombok already we can avoid it: http://jnb.ociweb.com/jnb/jnbJan2010.html#equals

> Well, I won't want to hash them all together. If the same HP occurs in another pokemon, it might still get a cache hit.

I'm proposing to use `HashMap<Key, String>`; that way, looking up an HP hash can't find a Candy result, exactly like in your code using strings.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: It's much more expensive than I had guessed. #158 avoids FP without crazy tricks.

Blaisorblade: > I'd still not bother doing this for OCR caching though because this is still a lot of work to save a few bytes of OCR cache.

Fine with that, with all those multiplications this is likely more expensive than the hashing.

sarav:
Ah, looks like I don't know how to use github fully yet. I'll close this one to avoid picking up the wrong commit.

PR #142, opened 2016-08-23
fix loop which is not a loop, ignore okio errors

jehy:
Replaced loop which is not a loop, also removed warnings from okio (see this: http://stackoverflow.com/questions/36010942/getting-warnings-with-proguard-with-external-libraries).


nahojjjen:
But when gen 2 comes out, if it scans Oddish, the current implementation should find gloom, vileplume and bellossom - this implementation would miss bellossom, wouldn't it?

jehy:
Devided this into two pull requests.

PR #143, opened 2016-08-23
Add instructions for lombok

jehy:
My Android Studio did not understand notations until I installed the plugin.


Blaisorblade:
Thanks for figuring this out, I also had the same problem!

PR #144, opened 2016-08-23
fix loop which is not a loop

jehy:
Replaced loop which is not a loop.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -653,14 +653,14 @@ public class Pokefly extends Service {
                 returnVal += "<br>"; //breakline
                 //for following stage evolution (example, dratini - dragonair - dragonite)
                 //if the current evolution has another evolution calculate its range and break
-                for (Pokemon nextEvo : evolution.evolutions) {
+                if(evolution.evolutions.size()>0) {
+                    Pokemon nextEvo=evolution.evolutions.get(0);
                     pokemonName = nextEvo.name;
                     returnVal += "\n" + String.format(getString(R.string.ivtext_evolve_further), pokemonName);
                     returnVal += "<br>"; //breakline
                     CPRange range2 = pokeCalculator.getCpRangeAtLevel(nextEvo, ivScanResult.lowAttack, ivScanResult.lowDefense, ivScanResult.lowStamina, ivScanResult.highAttack, ivScanResult.highDefense, ivScanResult.highStamina, Math.min(trainerLevel + 1.5, 40.0));
                     returnVal += "\n" + String.format(getString(R.string.ivtext_cp_lvl), range2.level, range2.low, range2.high);
                     returnVal += "<br>"; //breakline
-                    break;
                 }
             }
 


hlplulustg:
The loop is needed for multiple further evolutions (gen 2), remove break for that. If you are going to remove the loop we need some notes containing needed adjustments for gen 2.

jehy:
If it is needed for future purposes, at least ToDo should be added.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -653,14 +653,14 @@ private String getIVText() {
                 returnVal += "<br>"; //breakline
                 //for following stage evolution (example, dratini - dragonair - dragonite)
                 //if the current evolution has another evolution calculate its range and break
-                for (Pokemon nextEvo : evolution.evolutions) {
+                if(evolution.evolutions.size()>0) {
```
Blaisorblade: Nitpick: Can we keep to coding style? `if (`, not `if(` (note the space).

jehy: Coding style is kept by Android Studio formatter. But in this case I didn't want to auto format because it would reformat the whole file and cause merging conflicts. So it can be done later and also it isn't even worth mentioning.

Blaisorblade:
I agree that having a loop and break is silly. The original idea was different.

https://github.com/farkam135/GoIV/commit/5b86abc0a4bb3e56c3dbe8f9a3f0a57e3965e33d#commitcomment-18744543

FWIW, the least fragile way to write this uses a loop without that `break`—I agree with @hlplulustg.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: You can still write new code with proper style. Please fix it.

Also, does the coding style need space on either side of ">" ?

sarav:
Sorry about the bug.

sarav:
Actually, what pokemon did this break?

Blaisorblade:
> Actually, what pokemon did this break?

Not sure that happens right now—the current code is the result of so many commits I don't know. Ignoring generation 2 pokemon (where there would be a bug as @hlplulustg pointed out), a for with just one iteration is just harder to read than it could be.

It's not alone of course, but I'm just trying to do code review on the new code so that code quality improves.

And if I dig into history, it's not to accuse—it's just to understand whether the existing code should be like it is that because of some reason I'm missing.

sarav:
Yeah, no worries about digging into history. I'm trying to understand the code too. Didn't understand what the break was trying to do. In any case, I shouldn't have swallowed it in my new loop.

I think we need to pause the project and do a lot of clean up and then move again. But that's @farkam135 call.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
jehy: Sorry but I can't waste time on manual code formatting. If this pull request is declined - so be it. If not - a separate formatting commit is always welcome.

Blaisorblade:
I posted #160 removing break and adding comments, as alternative to this PR.

Issue #150, opened 2016-08-23
Galaxy Note 3 level arc misaligned


fredkilbourn:
Hey I have a Galaxy Note 3 and the level arc is a little bit misaligned, enough to cause false readings at my current level (27).

This is what is detected (lvl 28 mon):
![screenshot_2016-08-23-07-49-26](https://cloud.githubusercontent.com/assets/658757/17892500/34a265ca-6907-11e6-9550-fc024b384412.png)

This is what is accurate (lvl 28.5 mon):
![screenshot_2016-08-23-07-49-30](https://cloud.githubusercontent.com/assets/658757/17892502/38048ef0-6907-11e6-9f11-58de093e80fe.png)

Notice that both dots aren't aligned properly but the correct one is further misaligned.  Not sure what needs to change in the code but the whole overlay needs to be shifted vertically slightly I think.

Remiscan:
I think I know why this happens!

At level 1, the white dot is centered where the level arc starts. But at max level, it is not centered where the level arc ends! (but the red dot from GoIV is). Take a look a these pictures to understand what I mean:

http://imgur.com/a/BRNUP

This could be the cause of most misalignment issues, don't you think?

max1c:
Galaxy S7 Edge had this problem for past 2-3 versions. It's extremely annoying to deal with because I never know which reading is the right one. It's always either bellow or above the dot.

Blaisorblade:
I'm marking this as duplicate of #175 to track all "arc issues" together.

Issue #96, opened 2016-08-20
Arc misalignment


sarav:
I'm interested in fixing arc misalignment issues, but the code that calculates arc math has some magic values. Where do they come from?

Given the max possible pokemon level, is the arc not evenly split across all possible levels (and +.5 ones)? If it is, getting the arc to align properly shouldn't be an issue at all. If it's not evenly split, is there some documentation someone can point me to on how the math is done between the angles to levels?

farkam135:
it's not evenly split it's based on the CpM. Right now the issue seems to be coming from an inconsistency in trainer levels so from level 20-29 you subtract the degree by 0.5, from 30-? you add 1 degree, and from the screenshots i have seen if looks like 1 degree is also added to trainers level 10-19? This is what i'm currently looking into but I need more info on trainers and whether or not their arc is properly aligned, I was actually going to start a google forms for this. Since this theory is completely busted if a level 15 trainer comes out saying the arc is perfect while another says it's misaligned.

sarav:
I feel like there's some radius or arc center error too. Maybe floating point rounding errors? The red dot moves farther out from the center if I take it to the higher pokemon levels and aligns well with the arc when I take it to the lowest level.

farkam135:
With images I received from users I was able to check the formula for arc center and radius with the following resolutions:
1. 1440x2392
2. 1440x2560
3. 1080x1776
4. 1080x1920
5. 540x960
6. 720x1184
7. 720x1280
8. Just in the last commit 480x800
If there is a resolution not listed above it has not been tested but is probably a rounding issue and just needs to be adjusted by 1.

sarav:
Mine is 720x1280. Trainer level is 25. Problem was also there for level 24. Here are example screenshots.

![minlevel](https://cloud.githubusercontent.com/assets/247061/17829906/348ee47e-6672-11e6-9c53-8eff5ccf19af.png)
![maxlevel](https://cloud.githubusercontent.com/assets/247061/17829905/348f00f8-6672-11e6-84fa-a1954591ff22.png)

farkam135:
That looks fine, how does it look aligned over the pointer?

sarav:
Look at the circle on the 2nd picture. See how it's off center from the arc line? Compare it to the first pic. The center of the red dot lines up with the arc line nicely. It's not a huge issue, but it lacking the polish and it should be fixable. I think it's floating point arithmetic error. If you can explain some of the constants in the code, I might be able to take a look.

[...]

PR #152, opened 2016-08-23
Fix alignment issues with the level arc

Remiscan:
Based on these pictures (http://imgur.com/a/BRNUP), I was able to calculate that the max angle for the position of the white dot on the level arc is approximately 178.4 degrees, not 180. I tried to fix the code to account for that fact - you can see these magnificent calculations here: 
image

I replaced 202.04 by 200.00 in the code, deleted these weird -0.5 / +0.5 that were necessary before, and tested it on my phone (Nexus 5, Android 6). It fixed all the alignment problems I had! It probably needs to be tested on more devices, but I'm quite enthusiastic about this change.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -385,11 +385,6 @@ public class MainActivity extends AppCompatActivity {
 
         for (double pokeLevel = 1.0; pokeLevel <= trainerLevel + 1.5; pokeLevel += 0.5) {
             double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * 200.00 / Data.CpM[trainerLevel * 2 - 2];
-            if (angleInDegrees > 1.0 && trainerLevel < 30) {
-                angleInDegrees -= 0.5;
-            } else if (trainerLevel >= 30) {
-                angleInDegrees += 0.5;
-            }
 
             double angleInRadians = (angleInDegrees + 180) * Math.PI / 180.0;
 
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -384,7 +384,7 @@ public class MainActivity extends AppCompatActivity {
         Data.arcY = new int[indices];
 
         for (double pokeLevel = 1.0; pokeLevel <= trainerLevel + 1.5; pokeLevel += 0.5) {
-            double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * 202.04 / Data.CpM[trainerLevel * 2 - 2];
+            double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * 200.00 / Data.CpM[trainerLevel * 2 - 2];
             if (angleInDegrees > 1.0 && trainerLevel < 30) {
                 angleInDegrees -= 0.5;
             } else if (trainerLevel >= 30) {
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -383,8 +383,11 @@ public class MainActivity extends AppCompatActivity {
         Data.arcX = new int[indices];
         Data.arcY = new int[indices];
 
+        double maxAngle = 178.4;
+        double levelCoeff = maxAngle * Data.CpM[trainerLevel * 2 - 2] / ( Data.CpM[(int) ( (trainerLevel + 1.5)* 2 - 2 )] - Data.CpM[0] );
+
         for (double pokeLevel = 1.0; pokeLevel <= trainerLevel + 1.5; pokeLevel += 0.5) {
-            double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * 200.00 / Data.CpM[trainerLevel * 2 - 2];
+            double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * levelCoeff / Data.CpM[trainerLevel * 2 - 2];
 
             double angleInRadians = (angleInDegrees + 180) * Math.PI / 180.0;
 


Blaisorblade:
I've looked at your reasoning and it makes a lot of sense. But instead of measuring 178.4, would it be more robust to compute that from the circle radius?
I'm not sure know what's constant across screens vs what changes, but I guess the overall arc changes size, while I'm not sure about the circle. Not that I fully get the current logic.

Remiscan:
On the imgur album I linked, the first screenshot is from my phone, and the other two are from someone else's. The margins on each side of the arc are a different size on each phone, and yet the 178.4 value, computed from the last two screenshot, works perfectly on phone too. That's what made me think this value would be constant for all phones, but yes, it should be verified.

Remiscan:
Wait, I just tried to do the same calculation for a trainer level 7, to confirm that I would find the same value, and I found something completely different.

The value of x to reach a 178.4 degrees angle at level 7 is 212.20. And with a value of 202.04, as it was before, I only reach a max angle of 170 degrees... So then I launched GoIV 2.2.0 (the original one, the one without my change), set it to level 7, and the red dot at max level doesn't reach the end of the level arc, as predicted.

This is weird. Has it always been like that? Can someone else try it, set the app at level 7 (or any other level) and check if the red dot reaches the end of the arc at max level? Because it looks as if the 200.00 value has to be different for every level, and that would mean that even without my change, le position of the red dot at max level was very wrong before, particularly at levels far from 27.

Maybe I'm making a stupid mistake, because it's 2 a.m. here, but still, I wanted to post this because I find it very weird.

It would mean that, for each trainer level, you'd have to calculate x this way:

`x = 178.4 * Data.CpM[trainerLevel * 2 - 2] / ( Data.CpM[(int) ( (trainerLevel + 1.5)* 2 - 2 )] - Data.CpM[0] );`

and then put it into the angleInDegrees calculation:

`double angleInDegrees = (Data.CpM[(int) (pokeLevel * 2 - 2)] - Data.CpM[0]) * x / Data.CpM[trainerLevel * 2 - 2];`

Blaisorblade:
I can reproduce this at level 7. At level 11 the discrepancy is smaller. At level 1, the red dot goes beyond 180 degrees...

Remiscan:
That's what I saw as well. Now my question is... why did no one notice that before? Maybe because GoIV users are close enough to level 27, which makes the discrepancy really small?

Calculating x for every trainer level as I suggested in my last comment seems to solve these problems. I tried the modified app at level 1, level 7, level 26 and level 27, and the positions of the red dot look correct.

But I'm not very familiar with github. Should I close this pull request and create another one with the correct code?

Blaisorblade:
FWIW, I'm level 21.

Here's something much simpler.

Since pokeLevel <= trainerLevel + 1.5, and since they're 1-based but we want them to be zero-based, if the steps are to be uniform one should have something like:

```
angleInDegrees = (pokeLevel - 1)/(trainerLevel - 1 + 1.5)*180
```

Under the hypothesis I've taken, that's the correct formula—if you have something else it must be equivalent. Or I screwed up some hypothesis—which one?

Because of the issue you described, you'd want instead:
```
angleInDegrees = (pokeLevel - 1)/(trainerLevel - 1 + 1.5)*178.4
```

I would have thought CpM is some sort of lookup table to optimize this computation, but since CpM is used for computing the IV, I'm confused it makes sense here as well.
(Of course it might, but let's keep things simple for a second).

farkam135:
@Remiscan This was noticed except I was solving it by manually adjusting the degree based on trainer level, your solution is a lot more elegant :). Go ahead and update your pull request and i'll merge it!

Remiscan:
@Blaisorblade the steps are not uniform, they get smaller the closer you get to the max, and their size depends on CpM.

I just updated the pull request. I don't know anything about Java or Android development or Github so I'm not sure if I'm doing anything correctly or not. But my code worked on my phone :p

sarav:
I'm not fully sure what the 202.4 or 200 is representing. I'll read it more when I get home. But @Remiscan please combine those 3 commits and give a meaningful commit text explaining what the changing is doing. "Update MainActivity.java" isn't good. If you don't know how to use git, I can help.

sarav:
Thanks for figuring this out though. This is the most important issue in GoIV. I'll test it later today.

farkam135:
Are we sure that the max poss angle is 178.4?

Remiscan:
@farkam135 That's the value I calculated based on the screenshots from #150 (Galaxy Note 3), and that works on my phone (Nexus 5). I tried it on a Nexus 4 and it seems to be the correct value for that phone as well. But I don't know if it's the correct value for every phone.

farkam135:
@Remiscan Can you go ahead and post a screenshot of the dot lined up using your changes. I tested it out and it seems like it's like .5 degrees more than the actual point, but it could be other factors as well.

Remiscan:
@farkam135 Okay so here are two sets of screenshots:

http://imgur.com/a/xb1bw this first set shows that the red dot is correctly positioned at max level (I superimposed all the screenshots and the red dot was always exactly at the same place, at trainer levels 1, 7, 21 and 27).

http://imgur.com/a/XvWgD this second set shows that the level is correctly detected by GoIV on various Pokémon, and that the red dot is correctly positioned in each case (although it seems the red dot is a bit too "far away" along the radius of the arc).

farkam135:
@Remiscan interesting, I think trainer level still might have something to do with it and rounding issues. I'm level 22 and you can see i'm getting like .5 degrees more, here are some screenshots of my Grimer and a max level Dragonite:  
https://dl2.pushbulletusercontent.com/GBsIbUKIY05TDk87cWShZf9K0S4zbUqF/Screenshot_2016-08-23-22-39-39.png 
https://dl2.pushbulletusercontent.com/IoBSthAq9TsMrqHn3VvQZlLIWLNQRkUh/Screenshot_2016-08-23-22-39-49.png

Not a huge deal and i'll still merge this since it's a better alternative than the old code, but something we should look into.

sarav:
@Remiscan I wish for all those screenshots, you also clicked check IV and posted the screenshot to show that the level was actually maxed out.

Btw, I too have been saying that the dot seems to be drawn with a larger radius when it's at max. But is fine when it's at the lowest. 

@farkam135 I wonder if the arcCenterY (or whatever it's called) should be stored as a float. For example, the arcCenterY for a 720p device is computed to be 360. But I wonder if it should be 360.5 since that's the true center of the screen. But I could be wrong.

farkam135:
@sarav For the arcCenterY and Radius I manually went into each of the resolutions screenshots and got the values, making sure the formula comes out to the appropriate values. There is no point in storing the Y as a float because there is no float pixel location, you can't find the coordinate (0,360.5) on the screen.

sarav:
I know there isn't a float pixel location :) But when you compute a point in an arc/circle, it could result in rounding +/- 1 pixel. But I think the red dot if off by more than just 1 pixel. So, that's probably not it. This is very weird though. I wonder how it can be wrong along the radius.

farkam135:
@sarav Ah right, rounding is actually already considered since the ratio being used now is off by one point for certain resolutions and that is why it is manually adjusted for those. You can see that right after arcInitialY is created: https://github.com/farkam135/GoIV/blob/37336a661e400d2c0a47b4823f6ed6347954d9ee/app/src/main/java/com/kamron/pogoiv/MainActivity.java#L192-L198

sarav:
@farkam135  I said arcCenterY, but I was meaning arcX this whole time.

But looks like storing the Y in float would make sure you don't have to fix it per screen.

Remiscan:
@farkam135 If you give me a screenshot of your Dragonite without the red dot, I will calculate the precise max possible angle for your phone. I'd need screenshots from different phones, at different resolutions, of non-fully powered Pokémon (so that I can clearly see the end of the arc to measure its exact radius). I'll ask on reddit and to my friends and see if I get interesting results.

sarav:
@Remiscan With you patches I still have a misalignment. Could you check for my screenshots in #96 too please?

Remiscan:
@sarav I checked them, it looks as if the 178.4 max angle should be correct for you. Could you take a screenshot of the misalignment, and if you have a fully-powered Pokémon, a screenshot of that one as well (this last one without GoIV)?

Remiscan:
From the screenshots I gathered on reddit, the max possible angle should be between 174.39 and 174.45 degrees for every phone I could test (the average is 174.42, I tested 174.3 and 174.45 on my phone and the difference is really small). So the value of the max angle should be ok.

Here's the data I collected (and used to calculate max possible angle using math instead of measuring it manually): http://i.imgur.com/tRRjjsr.png

Now, these max possible angles should be correct if:
- the level arc is a half-circle (it seems to be, or it's very close),
- the position of the white dot on a level 1 Pokémon is at angle 0 (it is on my phone, and I guess everyone can check with GoIV, since GoIV places the red dot at angle 0 for a level 1 Pokémon -- since I saw most people complain about GoIV being aligned on low-level Pokémon and misaligned on high-level Pokémon, I guess this assumption is correct),
- the bottom of the white dot at max level is exactly at the same height as the center of the white dot at level 1 (once again, I checked it on my phone and it's correct, I would need screenshots of level 1 AND max level Pokémon from the same phones to verify it for other phones).

If these 3 assumptions are correct, I don't see why you would observe misalignment as you see it on your phone @farkam135. If anyone can think of another potential cause of misalignment, I'd like to know :p

sarav:
@Remiscan I'm pretty sure I saw your patches in my git log when I compiled and tried it earlier today. But maybe the adb push failed and I didn't notice it or something? In any case, the dot looks pretty well aligned now. There's a few pixels of misalignment but, it's almost nothing. So, I'm happy.

Thanks for fixing it! I'll respond to your other questions in the new issue that @Blaisorblade created.

PR #157, opened 2016-08-23
Fix failing build

imapokemon4:


Blaisorblade:
Can you compare with #145? @jehy Do we need both or is one enough? (Just noticed possible overlap, no clue what's best).

farkam135:
We can just not warn since they aren't used.

imapokemon4:
Yup.  
  
This fixed it for my debug build but to generate a signed apk I only had to put -dontwarn okio.** in pro-guardrules.pro.

imapokemon4:
Actually I'm not even sure I needed it for the debug build. I first tried to generate the apk right off, saw the error and added the dependency and tried to debug and it worked so I thought I needed it.  
  
But yeah I think the dontwarn is enough.

jehy:
Yup, exactly :)

PR #158, opened 2016-08-23
Avoid floating-point ops in replaceColors

Blaisorblade:
@sarav pointed out `replaceColors` does lots of floating-point computation: https://github.com/farkam135/GoIV/pull/140#discussion_r75915929

This PR avoids FP computation while giving the same results (unless I screwed up).

Untested—don't merge before review.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -910,27 +910,28 @@ public class MainActivity extends AppCompatActivity {
 
     /**
      * replaceColors
-     * Replaces colors in a bitmap that are not the same as a specific color.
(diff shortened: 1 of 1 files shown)


Blaisorblade:
@sarav Do you think this makes sense?

jehy:
Yup, that's definitely better.

sarav:
Looks good. I'd still not bother doing this for OCR caching though because this is still a lot of work to save a few bytes of OCR cache.

Blaisorblade:
> I'd still not bother doing this for OCR caching though because this is still a lot of work to save a few bytes of OCR cache.

Agreed.

So I think we're all happy with this! Ready for merge I think @farkam135 :-)

Blaisorblade:
Merging. I don't like merging my own PRs in general, but this appears uncontroversial. (Plus, Git has revert).

PR #159, opened 2016-08-23
Narrows down spinner options to candy evolution line

nahojjjen:


Blaisorblade:
We've been discussing alternatives in #130. It'd be probably good to add still add remaining pokemons in sorted order, even when detection worked, as proposed by @sarav here:
https://github.com/farkam135/GoIV/issues/130#issuecomment-241621066

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -449,6 +447,21 @@ public void backToIvForm() {
...
 
     /**
+     * changes the spinner content to the scanned pokemon candy name evolution line
+     *
+     * if the candy evolution line is 0 (complete read error) the whole pokedex will instead be shown
+     */
+    private void populateSpinnerWithEvolutionLine(){
+        ArrayList<Pokemon> evolutionLine = pokeCalculator.getEvolutionLine(pokeCalculator.get(candyName));
+        if (evolutionLine.size() >0){
+            pokeAdapter.updatePokemonList(pokeCalculator.getEvolutionLine(pokeCalculator.get(candyName)));
```
Blaisorblade: I'd rather reuse `evolutionLine` instead of recomputing it.

Blaisorblade:
@nahojjjen Thanks for reacting, but the idea we discussed was to add all other pokemons as well.

Issue #162, opened 2016-08-23
Temporary Bloat Clean


farkam135:
GoIV is getting pretty fragmented and bloated so let's clean some stuff up before the redesign. Below is a list of everything that should be high priority (visual changes). If anyone objects to any of these changes let me know, but please realize this is only temporary changes until the redesign and it's going to be hard to change my mind on these:  

- [ ] Shorten instructions to allow more room for 480p devices
- [x] Round ALL percentages (min/avg/max)
- [x] Possibilities, show only Min and Max combinations and then the number of total possibilities below. 3 Lines total! (if only 1 more possibility show it as well)
- [x] For evolution CP show it at current level not max level.
- [ ] Make CP/HP move up when keyboard opened (works for some devices)  
I'll be adding to this list as new things need to be adjusted.

sarav:
1. What redesign?
2. So, this checklist the stuff that's considered "bloat"? As in UI bloat? Or these are the things that need to be done before the redesign?

In any case, I think the code needs some refactor and clean up to make it more friendly to code. Here are some suggestions:

0. Run some autoindenter on the whole time and make one single commit. That's make the rest of the steps more readable and manageable.
1. Move the pokedex and pokemap from PokeInfoCalculator. Those two have nothing to do with just calculations any more. If anything, we should move them into static variables in Pokemon.java or create a Pokedex.java
2. getPossiblePokemon should be part of Pokemon.java
3. All the OCR analysis code should move to something like OCR.java or PokeOCR.java. And maybe even make it a singleton class. The constructor could do the tesseract init, the arcpoints calculations, etc
4. Pokefly wouldn't need to send intents back and forth between MainActivity and Pokefly to process screenshot then.
5. Probably move any other remaining Intent receivers into Pokefly
6. Delete all the useless comments of code. That's actually bad style when using git. Bring back old code is what git revert is for.

I can work on some of these, but I wouldn't work on it without some consensus on what the people working on the refactor are going to do. Otherwise, we'll have a ton of wasted effort from merge conflicts.

jehy:
Reformatting commit is a good one. Better to do make it when @farkam135  is online to avoid merge conflicts.

Blaisorblade:
Agreed on the reformatting—we just need to minimize conflicts, probably by deciding on the open PRs. For the rest I see lots of sensible ideas but I haven't looked in detail.

[...]

PR #163, opened 2016-08-23
Solves 3 of the "bloat issue" things

nahojjjen:
Average % rounded
Only min/max IV shown
Current level evolution CP estimate instead of trainer max


Blaisorblade:
For the record, this is about #162.

PR #164, opened 2016-08-24
Fix IV button disappearing and GoIV crashing on stop

### LINKED ISSUE #135: Gastly crashes GoIV
1. Start service
2. Try to scan Gastly, 4 out of 5 times the IV button just disappears.
3. Go back to GoIV screen, click STOP button, the application closes with "Unfortunately, GoIV has stopped.".

Samsung Note 5, 6.0.1, GoIV 2.1.0. It didn't happen with the last version from the Play Store, it would just fail to recognize the CP when the smoke covered it.

Blaisorblade:
This PR fixes #135. The various bugs are described in commit messages. There are a few refactorings at the beginning which I did while reviewing possibly relevant code.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -215,5 +215,5 @@
     <string name="auto_update_setting_summary">Automatically check for updates when the app starts up</string>
     <string name="check_for_update_setting_title">Check for update</string>
     <string name="check_for_update_setting_summary">See if an update is available</string>
-    <string name="scan_pokemon_failed">Scanning Pokémon failed, scroll to the top and try again!</string>
+    <string name="scan_pokemon_failed">Scanning failed, please scroll up and try again!</string>
 </resources>
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -180,7 +180,7 @@
     <string name="ivtext_iv">Min: %1$s%% Average: %2$s%% Max: %3$s%%\n</string>
     <string name="ivtext_evolve">Can be evolved into %1$s:\n</string>
     <string name="ivtext_evolve_further">Can be further evolved into %1$s:</string>
-    <string name="ivtext_many_possibilities">There are too many possibilities for this pokemon. Try powering it up!</string>
+    <string name="ivtext_many_possibilities">There are too many possibilities for this Pokémon. Try powering it up!</string>
     <string name="ivtext_max_lvl_cost">Cost to reach lvl %1$s:</string>
     <string name="ivtext_max_lvl_cost2">Candy: %1$s Stardust: %2$s</string>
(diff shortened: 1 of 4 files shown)


Blaisorblade:
Tested this on top of 33d9f66584248377d4ba5ab2fb4e49e15466ef96, rebased on new master, should still work well. Retesting.
I would also appreciate some reviewing work—this should prevent #135 but I might easily have missed some trickier corner case.

Blaisorblade:
Caveat: if when one presses the IV button we fail to scan a pokemon, now we just show again the IV button. That's what the code tried to do, only now it works, but we should probably tell the user. I'll try informing the user with a Toast...

To reproduce this in testing, I follow @rpgrca's tip:

> By the way, if you are in a Pokemon screen, scroll up the page (so that the CP goes out of scope) and press the IV button, it disappears and when you stop the service it crashes as it did with the Gastly problem.

jehy:
Looks good on first glance.

sarav:
Does this handle battery saver mode? You shouldn't show the IV button in battery saver mode.

Blaisorblade:
> Does this handle battery saver mode? You shouldn't show the IV button in battery saver mode.

Haven't tested that, but after reading the code I think the button shouldn't be shown there and it's hard to mess that. Of course that should still be tested.

Blaisorblade:
@DJCrapsody linked to this report, haven't tested this PR with it yet (will do later). Looking at the code, I think this bug is orthogonal to the changes here.

https://www.reddit.com/r/GoIV/comments/4zarig/bug_if_you_click_the_iv_button_while_the_pokemon/d6urlno

app/src/main/res/values/strings.xml
```diff
@@ -215,5 +215,5 @@
     <string name="auto_update_setting_summary">Automatically check for updates when the app starts up</string>
     <string name="check_for_update_setting_title">Check for update</string>
     <string name="check_for_update_setting_summary">See if an update is available</string>
-    <string name="scan_pokemon_failed">Scanning Pokémon failed, scroll to the top and try again!</string>
+    <string name="scan_pokemon_failed">Scanning failed, please scroll up and try again!</string>
```
Blaisorblade: @nahojjjen That's OK with you? I can put what you prefer :-)

nahojjjen: Looks better

Blaisorblade:
Just tested this also in battery saver mode, it seems fine and doesn't show the IV button. The only slight annoyance is that taking a screenshot after scrolling shows no error; however, after that, taking a correct screenshot worked again.

Since otherwise there's consensus, I think we can finally merge this!

sarav:
Btw, a lot of this wouldn't have been needed if we didn't use as many unnecessary Intents I refer to in #162

PR #165, opened 2016-08-24
Preserve user corrections of pokemon name OCR across app launches

sarav:
So, this finally saves the user corrections. So, once the user corrects the Beedrill issue or any of their renamed pokemon, the app should remember them across restarts.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -11,6 +11,7 @@ import android.content.ClipboardManager;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
+import android.content.SharedPreferences;
 import android.graphics.Color;
 import android.graphics.PixelFormat;
 import android.net.Uri;
@@ -38,6 +39,7 @@ import java.text.NumberFormat;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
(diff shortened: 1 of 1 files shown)


nahojjjen:
I tried using this, and while in normal case it seems to work well, there is the scenario  that if you're rescanning something that you've renamed to an iv range (for example I rename all my 10-40% to "20" when I'm lazy) and then rescan another pokemon which has the same rename, it will think that the name is connected to the pokemon.

For example, my rattatta "20", I tell GoIV that it's a rattatta, and then go onto my golbat "20". It now thinks the golbat is a rattatta. 

I'm just pointing out a potential issue, but I think the issue is rare enough that this pull can still be accepted, and potentially fixed afterwards.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -68,6 +70,7 @@
     private static final String KEY_SEND_INFO_LEVEL = "key_send_info_level";
     private static final String KEY_SEND_SCREENSHOT_DIR = "key_send_screenshot_dir";
 
+    private static final String PREF_USER_CORRECTIONS = "user_corrections";
```
Blaisorblade: Based on docs in https://developer.android.com/training/basics/data-storage/shared-preferences.html#GetSharedPreferences, we should use here `com.kamron.pogoiv.USER_CORRECTIONS` as name, IIUC to avoid conflicts. Can you fix that?

Otherwise the code LGTM! 👍

Blaisorblade:
Regarding the stored preferences: maybe we should also limit the file size so it doesn't grow too much. I guess we'll notice during testing, if we actually get to play ;-).

sarav:
Will address the comment. We can fix the size if people open issues. I doubt a few names are going to take up more than a few KB.

Blaisorblade:
Thanks for addressing this... LGTM (= Looks Good To Me). 👍 
@nahojjjen, if you don't have further comments please merge!

PR #170, opened 2016-08-24
Updated README to 2.2.0

DJCrapsody:


jehy:
Change log should show corresponding version.

DJCrapsody:
Possibly redundant as version is stated under Download. If there is one more objection with my reasoning, I will put the version back in the header.

jehy:
Change log should be kept for all versions, not only for last one.

Blaisorblade:
I agree the changelog should include older changes, but maybe not inside the README? We could have a ChangeLog.md file in the repo and link to it, and put there the latest release on top.

jehy:
Yup, separate change log is fine too.

Blaisorblade:
I think we can merge this too since the README is still out of date; I'll accept PRs to improve this further as discussed.

Issue #141, opened 2016-08-23
Cache Bitmap OCR results


sarav:
I gave it a shot in #140 . The logic works.

If the Bitmap is fixed, I don't see any reason the OCR results should change each time. I'd also think that the OCR code is a lot more complicated and floating point intensive than hashcode generation. Unfortunately, Bitmap.hashcode() doesn't generate a hashcode based on the bitmap picture. So, I had to extract the pixels to do it. I'm unable to officially verify it though. I guess I could do a timestamp check. Didn't think of that -- but I got to log off for today.

Can others (@Blaisorblade @Eccenux @mattaereal @nahojjjen @farkam135) give it a shot and let us know what you think? Does it feel a bit faster when you check the IV for the same type of Pokemon (assuming you didn't rename it) or when you check the same Pokemon again? Is the LRU size of 200 too small? I'd like to hear some thoughts.

farkam135:
It might just be me but I usually don't rescan unless there was an obstruction. It sort of ruins the opportunity for me to rescan it at a better time if the results are cached.

sarav:
I think you misunderstood the issue/code (not sure you looked at it). If the issue is an obstruction, the hashes won't match. I'm using the cached value only if the "pictures" are identical. If the pictures are identical, I'd expect tesseract to give the same result.

I rescan because I go back and forth between a few pokemon that I'm trying to decide on transferring and I keep forgetting their IV range.

farkam135:
Ah yea, just looked at the code my bad, how much size does each cached scan take up?

agrajaghh:
Do you think you will ever get two identical bitmaps? Since there is the background animation with the bubbles and the pokemon is moving..

Blaisorblade:
> I'm using the cached value only if the "pictures" are identical.

Well... if there are no collisions. Avoiding this problem (with a `HashMap<HashableImage, String>` from a new `HashableImage` class with a suitable `hashCode` method), OTOH, would mean storing the image.

Blaisorblade:
> Do you think you will ever get two identical bitmaps?

The patch hashes the areas containing the pokemon and candy names, so that shouldn't matter.

agrajaghh:
> The patch hashes the areas containing the pokemon and candy names, so that shouldn't matter.

ah okay, so hashing would just be useless for CP...

Blaisorblade:
Re size: right now we use little because we don't store the images.

[...]

PR #178, opened 2016-08-25
Cache OCR results for parts of the bitmap that don't change often

sarav:
Most of the regions of the image that need OCR don't change often across
different times the same pokemon is analyzed. Things like name and candy
name are even common across different pokemon. So, we can safely cache
the results of OCRing these images.

To save memory, we cache the results using a hash of the image and not
the entire image.

The CP image changes a lot due to changing background. So, there's not
much of a point to cache it. So, we don't cache OCR results for the CP.

Profiling of this code showed that the caching operation takes about
1-3 milliseconds. Whereas OCR takes from 50-260 milliseconds depending
on the image being OCRed. So, this change definitely bring in a lot of
performance and power saving benefits.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -39,6 +39,7 @@ import android.support.v7.app.AlertDialog;
 import android.support.v7.app.AppCompatActivity;
 import android.text.method.LinkMovementMethod;
 import android.util.DisplayMetrics;
+import android.util.LruCache;
 import android.view.Display;
 import android.view.Menu;
 import android.view.MenuInflater;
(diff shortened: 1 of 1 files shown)


agrajaghh:
what do you think about using md5 or sha, since they are not much slower? This would offer the possibility to store the cache between app starts which is not possible with the current hashCode implementation (see https://github.com/farkam135/GoIV/issues/141#issuecomment-241832909)

sarav:
I did think of that. And I'll probably do that, but I need to figure out the limits to set and iron out a few thoughts. So, I'll do that later as a separate patch. Thanks for measuring it. I'm kinda surprised that md5 is that fast. Also, if I cache a lot, the collisions could increase. But no reason for a single run of the app to not benefit from this now.

I'd also like to see if we can speed up OCR (see other issue I opened) and maybe caching won't be needed.

Blaisorblade:
LGTM and @nahojjjen, merging.

PR #181, opened 2016-08-25
Made adjustments for offline release

### LINKED ISSUE #172: Add offline version to releases
Downloading offline version from Google drive seems very clumsy. I think that offline version should be also released on github - at least to avoid fakes.
However, this implementation requires minor changes to updater - it should not replace online version with offline one.

rhari991:
When making a new version, you need to update all three versionName variables in the build.gradle file. Also,the string "(Offline)" should be present in two places :

1) The versionName variable in the offline closure in the build.gradle file and
2) The tag name in the GitHub release of the offline version

Fixes #172

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,10 +21,8 @@ android {
     compileSdkVersion 24
     buildToolsVersion "24.0.1"
 
-    //When making a new version, you need to update all three versionName variables below. Also,
-    //the string "(Offline)" should be present in two places :
-    //1) The versionName variable in the offline closure in this file and
-    //2) The tag name in the GitHub release of the offline version
+    //When making a new version, please ensure that the apk files for both online and offline flavors
+    //are in the same GitHub release and the offline apk filename has "Offline" in it.
 
     defaultConfig {
         resValue "string", "app_name", "GoIV"
@@ -42,7 +40,6 @@ android {
             buildConfigField "boolean", "isInternetAvailable", "false"
             resValue "string", "app_name", "GoIV Offline"
             ext.enableCrashlytics = false
-            versionName "2.2.0 (Offline)"
         }
 
         online {
@@ -50,7 +47,6 @@ android {
             buildConfigField "boolean", "isInternetAvailable", "true"
             resValue "string", "app_name", "GoIV"
             ext.enableCrashlytics = true
-            versionName "2.2.0"
         }
     }
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,6 +21,11 @@ android {
     compileSdkVersion 24
     buildToolsVersion "24.0.1"
 
+    //When making a new version, you need to update all three versionName variables below. Also,
+    //the string "(Offline)" should be present in two places :
+    //1) The versionName variable in the offline closure in this file and
+    //2) The tag name in the GitHub release of the offline version
+
     defaultConfig {
         resValue "string", "app_name", "GoIV"
         applicationId "com.kamron.pogoiv"
@@ -37,6 +42,7 @@ android {
             buildConfigField "boolean", "isInternetAvailable", "false"
             resValue "string", "app_name", "GoIV Offline"
             ext.enableCrashlytics = false
+            versionName "2.2.0 (Offline)"
         }
 
         online {
@@ -44,6 +50,7 @@ android {
             buildConfigField "boolean", "isInternetAvailable", "true"
             resValue "string", "app_name", "GoIV"
             ext.enableCrashlytics = true
+            versionName "2.2.0"
         }
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kamron/pogoiv/updater/AppUpdateLoader.java
```diff
@@ -27,10 +26,19 @@ public void run() {
...
             JSONArray releaseInfo = new JSONArray(response.body().string());
-            JSONObject latestRelease = releaseInfo.getJSONObject(0);
-            JSONObject releaseAssets = latestRelease.getJSONArray("assets").getJSONObject(0);
 
-            AppUpdate update = new AppUpdate(releaseAssets.getString("browser_download_url"), latestRelease.getString("tag_name"), latestRelease.getString("body"));
+            //Loop through the releases till we get to the first online release
+            JSONObject latestOnlineRelease = releaseInfo.getJSONObject(0);
+            for(int i = 0;i < releaseInfo.length();i++){
+                JSONObject latestRelease = releaseInfo.getJSONObject(i);
+                if(!latestRelease.getString("tag_name").contains("(Offline)")) {
```
Blaisorblade: This is a Git tag_name—and the tag name for online and offline releases would usually be the same (since they're based off the same sources).
Looking at https://api.github.com/repos/farkam135/GoIV/releases, it'd be easier to look at the assect name and check whether it contains `-Online` (like `GoIV-Online.2.2.0.apk`) or not.

app/src/main/java/com/kamron/pogoiv/updater/AppUpdateLoader.java
```diff
@@ -27,10 +26,19 @@ public void run() {
         try {
```
Blaisorblade: As API endpoint we could replace `https://api.github.com/repos/farkam135/GoIV/releases` with an endpoint for the latest version, `https://api.github.com/repos/farkam135/GoIV/releases/latest`.
https://developer.github.com/v3/repos/releases/#get-the-latest-release

app/build.gradle
```diff
@@ -37,13 +42,15 @@ android {
...
             ext.enableCrashlytics = false
+            versionName "2.2.0 (Offline)"
         }
 
         online {
             applicationId "com.kamron.pogoiv"
             buildConfigField "boolean", "isInternetAvailable", "true"
             resValue "string", "app_name", "GoIV"
             ext.enableCrashlytics = true
+            versionName "2.2.0"
```
Blaisorblade: Is the versionName actually used? Still, I'd prefer to use the filename, which is more free. The current one, `GoIV-Online.2.2.0.apk` isn't `${app_name}.${version}.apk`.

app/src/main/java/com/kamron/pogoiv/updater/AppUpdateLoader.java
rhari991: Is it possible to release both apk files under a single release ? I'm asking because I haven't used GitHub releases yet.

app/build.gradle
rhari991: I'll rewrite this according to filename.

app/src/main/java/com/kamron/pogoiv/updater/AppUpdateLoader.java
Blaisorblade: I *guess* so, because `assets` is an array and I hope the the API is sensible enough. I hope there are docs somewhere...
The "nice" thing is, even I can't test this on this project without triggering autoupdates. `/latest` also skips prereleases which could be good. (Or maybe we should do the check ourselves and allow users to run on beta, but all of that requires more testing to get right).

app/src/main/java/com/kamron/pogoiv/updater/AppUpdateLoader.java
```diff
@@ -19,18 +17,19 @@
...
                 .build();
 
         try {
             Response response = httpClient.newCall(request).execute();
-            JSONArray releaseInfo = new JSONArray(response.body().string());
-            JSONObject latestRelease = releaseInfo.getJSONObject(0);
-            JSONObject releaseAssets = latestRelease.getJSONArray("assets").getJSONObject(0);
+            JSONObject releaseInfo = new JSONObject(response.body().string());
+            JSONObject releaseAssets = releaseInfo.getJSONArray("assets").getJSONObject(0);
+            if(releaseAssets.getString("name").contains("Offline"))
```
Blaisorblade: BTW, can you please reformat the code using Android Studio's formatter? I've reformatted stuff, we should keep it that way from now on.

rhari991: You mean use Code -> Reformat code ?

Blaisorblade: Yep.

Blaisorblade:
Thanks for the patience, and sorry for not giving all feedback in one go.

PR #186, opened 2016-08-25
Detach CHANGELOG from README

DJCrapsody:
As previously discussed, Added all changelog since 1.3.0. Need to limit on how many changelogs we are keeping, not feasible to keep all changelogs. Maybe maximum of 5?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,66 @@
+# Changelog
+
+## 2.2.0
+
+### New features
+* In-app update checker! The application will check the github repo for any newer version, and ask if you want to download an update, if a new update comes. (This feature is only for the internet version of the app)
+* Settings menu! Toggle if pokemon go launches automatically on start, if you want the app to skip the input field and show results directly, delete screenshots in battery saver mode & if the app should copy info to clipboard.
+* Now on tap support - If you're running in battery saver mode, you can use now on tap instead of taking screenshots. Press the share button, and share with GoIV.
+
+### Improvements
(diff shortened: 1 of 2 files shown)


Blaisorblade:
> Need to limit on how many changelogs we are keeping, not feasible to keep all changelogs. Maybe maximum of 5?

I see no reason to delete old changelogs, and I haven't seen people do it. I won't require digging them, but I'd accept any PRs adding them. But the PR as-is looks good!
@jehy what do you think?

jehy:
Let's take ACRA as example:
https://github.com/ACRA
On the main page it only publishes last version and has a link to changelog which resides in wiki:
https://github.com/ACRA/acra/wiki/ChangeLog
I thinks that it's a comprehensive method which helps to keep crap it of the repository.

However, that becomes useful only when there are already many version and large volumes of change logs. For now we can be fine with a separate changelog.md file. Bit it should contain all changelogs. And the main readme.md can have a link to this file.

DJCrapsody:
We can follow ACRA, having the version number on the main page and linking to the changelog for others to view the changelogs.

However, they keep changelogs till the first version of a major version. It's 4.9.0 so they keep changelogs till 4.0.0 instead of 3.0.0 or older.

jehy:
Yeah, hovewer older acra's changes can be found in wiki history. I think that that's a good approach.

Blaisorblade:
Again, I don't see reasons to delete old changelogs—they're just at the end. And they're not what makes the repo big.

Anyway, merging this PR for now; we can discuss removing old entries when there's an actual need, if ever.

PR #192, opened 2016-08-25
Add French translation

pace6666:
I'm not sure with the apostrophes used in french and the code. I used \ before


Blaisorblade:
Hi, thanks for the PR! Have you had a chance to try it out? Using `'` should work IIUC (looking at the other translations), but it'd be good to try it out anyway.

PR #195, opened 2016-08-26
Move OCR code to separate class

sarav:
Initial set of commits I'd like to merge. I'll proceed to clean this up further.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -52,7 +52,6 @@ import android.widget.EditText;
 import android.widget.TextView;
 import android.widget.Toast;
 
-import com.googlecode.tesseract.android.TessBaseAPI;
 import com.kamron.pogoiv.updater.AppUpdateEvent;
 import com.kamron.pogoiv.updater.AppUpdateLoader;
 import com.kamron.pogoiv.updater.AppUpdateUtil;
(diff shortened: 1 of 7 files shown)


sarav:
I tested this on a few pokemon. Works fine.

app/src/main/res/values/strings.xml
```diff
@@ -4,7 +4,6 @@
 
     <string name="pokemon029">Nidoran♀</string>
     <string name="pokemon032">Nidoran♂</string>
```
Blaisorblade: Can we also remove `pokemon029` and `pokemon032` like `pokemon133 `, since they're also redundant?

Blaisorblade:
LGTM!

Blaisorblade:
Thanks @nahojjjen for merging!

BTW, as a general rule, as I wrote on Discord: I think we should try to merge refactorings fast, in general, so they don't have to be updated too much.
Otherwise doing refactorings will get too painful.

In case of conflicts with long running branches... we'll see in each case.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -253,6 +250,7 @@ public void onClick(View v) {
 
 
         displayMetrics = this.getResources().getDisplayMetrics();
+        initOCR();
```
Blaisorblade: This call uses `candyOrder` before it's initialized correctly. Fixed in c9521d0.

Issue #171, opened 2016-08-24
GoIV crashes when deleting CP/HP


Blaisorblade:
Reported by https://www.reddit.com/r/GoIV/comments/4zarig/bug_if_you_click_the_iv_button_while_the_pokemon/d6urlno.

To just fix the crash, we need a try..catch here, and show, say, a Toast to report the error:

https://github.com/farkam135/GoIV/blob/37336a661e400d2c0a47b4823f6ed6347954d9ee/app/src/main/java/com/kamron/pogoiv/Pokefly.java#L421-L422

But it'd be good to have some nicer validation UX.

Blaisorblade:
Fixed by #216.

PR #216, opened 2016-08-27
Multiple small fixes

oskomonzon:
Quick fix addressing #171


oskomonzon:
Another quick fix aiming to improve user experience by not showing the result box if there are no IV combinations for the data provided.
Instead we display a toast informing the user and stay in the input dialog.

This was discussed in #217, pending review.-

oskomonzon:
One last quick fix addressing point number 5 discussed in #217.
Limiting user input to 4 digits should be enough. We are still allowing parameters over 9000!

farkam135:
I like this. Let's see what @nahojjjen and @Blaisorblade think as well.

PR #219, opened 2016-08-27
Optimize ocr

sarav:
Fixed the improper cutting of candy bitmap and made change that skips a ton of math. replaceColors() should be much faster now.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OCRHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OCRHelper.java
@@ -195,7 +195,7 @@ public class OCRHelper {
      * @return the candy name, or "" if nothing was found
      */
     private String getCandyNameFromImg(Bitmap pokemonImage) {
-        Bitmap candy = Bitmap.createBitmap(pokemonImage, widthPixels / 2, (int) Math.round(heightPixels / 1.3724285), (int) Math.round(widthPixels / 2.057), (int) Math.round(heightPixels / 38.4));
+        Bitmap candy = Bitmap.createBitmap(pokemonImage, widthPixels / 2, (int) Math.round(heightPixels / 1.3724285), (int) Math.round(widthPixels / 2.1), (int) Math.round(heightPixels / 38.4));
         String hash = "candy" + hashBitmap(candy);
         String candyName = ocrCache.get(hash);
 
--- a/app/src/main/java/com/kamron/pogoiv/OCRHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OCRHelper.java
@@ -69,14 +69,26 @@ public class OCRHelper {
      * @param keepCb       The blue color to keep
      * @param replaceColor The color to replace mismatched colors with
      * @param distance     The distance threshold.
+     * @param simpleBG     Whether the bitmap has a simple background
      * @return Bitmap with replaced colors
      */
-    private Bitmap replaceColors(Bitmap myBitmap, int keepCr, int keepCg, int keepCb, int replaceColor, int distance) {
+    private Bitmap replaceColors(Bitmap myBitmap, int keepCr, int keepCg, int keepCb, int replaceColor, int distance, boolean simpleBG) {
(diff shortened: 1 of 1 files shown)


sarav:
Paging profile master @agrajaghh Would be nice if you could time replaceColors() on the same device you profiled before.

agrajaghh:
@sarav here you go:

|       | Optimized replaceColors | original replaceColors |
|-------|-------------------------|------------------------|
| **name**  | 1,92                    | 2,58                   |
| **candy** | 1,03                    | 1,40                   |
| **hp**    | 0,73                    | 0,99                   |
| **cp**    | 1,04                    | 1,37                   |

its faster, but why is cp faster as well? I run replace colors 500 times, and the results are the average from 3 pokemons.

branches:
https://github.com/agrajaghh/GoIV/tree/sarav_replaceColors_timing_1
https://github.com/agrajaghh/GoIV/tree/sarav_replaceColors_timing_2

sarav:
What's the unit here? I thought last time you reported numbers, replace colors was about 11ms. The numbers here seem too large for ms and too high for microseconds.

The only time CP could run faster is when the background has a lot of the replacement color (black in the case of CP). Did these Pokemon have noticeable amount of black around CP?

sarav:
Also, thanks a lot for the profiling work.

agrajaghh:
sorry, forgot to say it's **ms**.  The difference between the last timing is maybe that I now run replaceColors in a loop 500 times, which should be more accuarte than just running it once I guess.

agrajaghh:
> The only time CP could run faster is when the background has a lot of the replacement color (black in the case of CP). Did these Pokemon have noticeable amount of black around CP?

maybe a dark background, but its not 100% black.
It was Hypno, Vaporeon and Magmar

app/src/main/java/com/kamron/pogoiv/OCRHelper.java
```diff
@@ -195,7 +195,7 @@ private String getPokemonNameFromImg(Bitmap pokemonImage) {
      * @return the candy name, or "" if nothing was found
      */
     private String getCandyNameFromImg(Bitmap pokemonImage) {
-        Bitmap candy = Bitmap.createBitmap(pokemonImage, widthPixels / 2, (int) Math.round(heightPixels / 1.3724285), (int) Math.round(widthPixels / 2.057), (int) Math.round(heightPixels / 38.4));
+        Bitmap candy = Bitmap.createBitmap(pokemonImage, widthPixels / 2, (int) Math.round(heightPixels / 1.3724285), (int) Math.round(widthPixels / 2.1), (int) Math.round(heightPixels / 38.4));
```
Blaisorblade: LOL—I just tried a random number on my Nexus 5. If this actually works well... I'm scared.

Blaisorblade: To clarify what I'm scared (a bit) of: we have magic numbers with 7 significant digits, yet they can be fixed this easily?

sarav:
@agrajaghh 
I guess you are using "," as a decimal point in your language?
I have no clue why CP is faster. For those 3 pokemon, I doubt it even helped for a single pixel. It's probably one of those complicated system interactions. It could be the cache hit rate improvement because we aren't running all that additional code for the non-CP OCR, or the CPU and DDR might be scaling differently due to these changes, etc.

Btw, the whole profiling based on time is actually not a good metric (good enough for us for now) in systems where the CPU and other hardware blocks can scale based on load. If you put enough load, you might push the CPU to a higher performance level that'll finish your work faster -- but it's bad because you are finishing faster due to more load (wasting power) and not less load.

But I don't think that's happening here.

agrajaghh:
> I guess you are using "," as a decimal point in your language?

yeah, sorry, forgot to change it :smirk: 


> It could be the cache hit rate improvement because we aren't running all that additional code for the non-CP OCR, or the CPU and DDR might be scaling differently due to these changes, etc.

I cleared the app data before I made the timings to get real results

sarav:
I meant the L1 and L2 cache. Not the ocr cache in the app :-)

PR #224, opened 2016-08-27
Fix Advanced Info and Triangle

DJCrapsody:
Capitalise 'I' in Information for consistency
Replace BLACK RIGHT POINTING POINTER to BLACK RIGHT POINTING TRIANGLE so both triangles are the same

### DIFF
--- a/app/src/main/res/layout/dialog_results.xml
+++ b/app/src/main/res/layout/dialog_results.xml
@@ -249,10 +249,11 @@
                 android:textColor="@color/colorPrimary"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:text="▼ "
+                android:text="▼"
                 android:layout_gravity="center"
                 android:textSize="18sp"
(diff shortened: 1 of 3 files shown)


app/src/main/res/layout/dialog_results.xml
```diff
@@ -249,10 +249,11 @@
                 android:textColor="@color/colorPrimary"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
-                android:text="▼ "
+                android:text="▼"
                 android:layout_gravity="center"
                 android:textSize="18sp"
-                android:clickable="true"/>
+                android:clickable="true"
+                android:paddingLeft="4dp" />
```
Blaisorblade: Ah, the padding belong before and not after! Nice catch. And I agree that using layout instead of spaces is probably cleaner.

Blaisorblade:
LGTM. This should get in before next release, no?

PR #225, opened 2016-08-27
Quickfix for crash

Blaisorblade:
This fixes the crash noticed on Crashlytics for 3.0. I have more cleanups on this code, but I'd leave figuring out the root cause for next release.


nahojjjen:
Looks good to me

PR #226, opened 2016-08-27
Refactor Unused Strings

DJCrapsody:
Removes unnecessary strings from pre-UI overhaul
Note: Check before doing a refactor, some strings are still being used (Nidoran, screen capture and scan failed strings are used in MainActivity)


Blaisorblade:
LGTM!

Blaisorblade:
We should bring at least `ivtext_possibilities` back. That's unused because we've gone from using translatable resources back to non-translateable strings.

PR #240, opened 2016-08-28
stop keeping a static reference to context in hte crashlytics wrapper

harkin:
Static references to context often cause memory leaks. This one was fine but it's better to just not have them (people copy code patterns they see, the next time it mightn't be the application context that's kept around)

### DIFF
--- a/app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
+++ b/app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
@@ -11,12 +11,8 @@ import timber.log.Timber;
 
 public class CrashlyticsWrapper {
 
-    private static Context mContext;
-
     public static void init(Context context) {
-
-        mContext = context;
-        if (BuildConfig.isInternetAvailable && GoIVSettings.getSettings(mContext).getSendCrashReports()) {
+        if (BuildConfig.isInternetAvailable && GoIVSettings.getSettings(context).getSendCrashReports()) {
             // Set up Crashlytics, disabled for debug builds
             Crashlytics crashlyticsKit = new Crashlytics.Builder()
                     .core(new CrashlyticsCore.Builder()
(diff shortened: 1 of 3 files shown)


app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
```diff
@@ -30,10 +26,16 @@ public static void init(Context context) {
...
+
+        public CrashReportingTree(Context context) {
+            this.context = context;
+        }
+
         @Override
         protected void log(int priority, String tag, String message, Throwable t) {
 
-            if (BuildConfig.isInternetAvailable && GoIVSettings.getSettings(mContext).getSendCrashReports()) {
+            if (BuildConfig.isInternetAvailable && GoIVSettings.getSettings(context).getSendCrashReports()) {
```
Blaisorblade: Should we keep building the settings each time? Right now this code won't work correctly anyway if `getSendCrashReports` changes—or do we call again `CrashlyticsWrapper.init` then?

harkin: Nah building settings each time seems unnecessary. It looks like restarting the application is required to pick up that change in settings. There's maybe a thornier problem in that there's no obvious way to turn crash reporting off after Fabric has been initialised.

That's a problem for another PR though

Blaisorblade: So, I'd say, let's cache the flag. Can you change the PR to do that, while you're at it? Or do you want to make things different?

Blaisorblade:
Semi-OT: Actually, couldn't `getSettings` just cache the settings object and listen on some event to invalidate it whenever settings are actually changed?

app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
harkin: I'd all about single focused PRs 😄 I'd prefer to merge this and tackle settings as a separate issue

harkin:
> couldn't getSettings just cache the settings object and listen on some event to invalidate it whenever settings are actually changed?

It could, though I don't relish the idea of caching more static state somewhere. Looking a little closer at it I think there's scope to simplify settings a lot. I think the class could be a simple wrapper around shared preferences that just exposes some static method to check each setting. 

That would make firing an event/taking action when a specific setting is toggled more difficult. Kicking off an app update checker would be more difficult, likewise enabling/disabling crash reporting (aside from any issues involved in actually enabling/disabling these). If waiting until the next time the app is started for those changes to take effect is palatable I can make those changes. Perhaps unrelated, there's already some [weirdness](https://github.com/farkam135/GoIV/blob/master/app/src/main/java/com/kamron/pogoiv/SettingsActivity.java#L72:L79) happening when certain settings are toggled (why would turning it off kick off an update?)

app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
Blaisorblade: OK, let's merge this in.

Issue #235, opened 2016-08-27
3.0.1 refuses to process a 13CP/10HP pokemon


Blaisorblade:
We had changed that to refuse processing if *both* HP/CP are 10, this regressed (I assume already in 3.0.0). Moreover, it says there are *no* possible combination (for a Lvl 1 Eevee).

DJCrapsody:
It shows no possibility instead of too many possibilities whenever CP or HP is 10.

Currently I have:
CP10/HP10 Magikarp showing no possibility.
CP10/HP10 Caterpies showing no possibility. 
CP11/HP10 Pikachu showing no possibility.

Blaisorblade:
I tried having as message "Too many possible IV combinations", but it doesn't fit the layout :-(.

Blaisorblade:
This regressed in cf8ab33a5b7a69693be631137a2c58b5c4d7f9b7, merged in #202 as part of the new GUI. That commit:
- changes the code back to stop if either CP or HP is 10; @farkam135 had suggested otherwise;
- changes away from using `null` to signal too many possibilities (which is probably good);
- forgot to change the detection for "no possibilities".

PR #243, opened 2016-08-28
Allow processing pokemon without HP=CP=10 (#235)

Blaisorblade:
The code flipped back and forth. @farkam135 suggested using `||`, I changed it, and @nahojjjen changed it back in cf8ab33a5b7a69693be631137a2c58b5c4d7f9b7. GoIV can give useful info here, especially with narrowing—and even without there's a delay but I find it acceptable.

What do you think?


nahojjjen:
I honestly don't remember why I changed that

Issue #217, opened 2016-08-27
Feedback regarding latest build (correct IV guessed, no possible IV, colors)


rpgrca:
Just tried the latest code to check if it fixed the dot misalign (it does in Note 5 at least), quite nice! Leaving here some general feedback about it.

![2016-08-27 02 02 54_](https://cloud.githubusercontent.com/assets/15602473/18024490/fca4d1a2-6be1-11e6-8ac6-502ddc44b49b.jpg)

1) If the correct IV has been guessed, I believe it would be much more useful to show the IVs instead of showing the three boxes with the percentage.

![2016-08-27 02 03 04_](https://cloud.githubusercontent.com/assets/15602473/18024491/fca8418e-6be1-11e6-897e-010decdfbd3c.jpg)

I'm assuming the majority of people would end up expanding the combinations to check how the IV end up looking, so by showing the result in the main screen you prevent people from having to do an extra click.

~~2) If there's 0 possible combinations (like, clicked the wrong appraisal or had the wrong level, etc), maybe the "Show All" link should be disabled / not appear at all, since it shows an empty screen only. Also noticed that if you have too many results possible, it won't show any (like a L1 bulbasaur that has 4096 possible combinations): the link should also be disabled there.~~

~~3) Blue text with black background is somewhat hard to see in the combo box with the pokemon names.~~

oskomonzon:
Hey, thanks for the feedback. We are currently doing lots of quick iterations over the new changes (lots of them). So we really appreciate users input.

1 - I agree, maybe we can incorporate something different when there is a single match. Something cheerful to make the user feel awesome :P
2 - About this one, I think we need to do a few more passes and consider the best approach. Maybe we shouldn't even display a result screen if there's nothing to show.
3 - It's already addressed :+1: 

Again, thanks for the feedback and keep up the good work.

rpgrca:
Geez, nice to see I'm not the only Ingresser who's playing Pokemon ;-) Roberto here ;-)

Couple of crashes:

4) Start GoIV (with "Launch Pokemon Go" disabled to make it faster), kill it. Restart it, kill it again and it will crash.

5) Textboxes for CP / HP don't have input limit, inputting a large number crashes the application when checking IVs.

oskomonzon:
Made a quick fix for 2: If no results inform the user and keep the inputs dialog. (Pending review)
I'll take a look at 5) since that seems to be an easy fix (falling asleep right now).

rpgrca:
[...]

PR #247, opened 2016-08-28
Single match layout behaviour

oskomonzon:
Added new UI elements to address the single match scenario. This was partially discussed on Issue #217.
The rationale behind it is that if there's a single match there is no need for further refinement, and its just a hassle for the user to tap "See all" to see the one result.

Result values are color coded with the same ratios as the percentages from the IV calculations.

Here's an example of how it looks like:
screenshot_2016-08-28-04-22-51


nahojjjen:
Looks good

PR #248, opened 2016-08-28
Clean up screen handling code and intent flow

### LINKED ISSUE #203: Simplify intent flow to prevent bugs
Sometimes IV button still won't show on pokemons. Reason: `Pokefly.infoShownSent` sometimes gets stuck to `true`, because the `MainActivity` still doesn't answer with the pokemon screenshot.

BTW, we're also getting the screenshot twice for no reason it seems (with copy-pasted code), you'd think once would be enough for both detecting "this is a pokemon" and getting its stats?

I've just gotten it again when I accidentally suspended the apps (by pressing the home button); I suspect we can't assume intent delivery is guaranteed. In that case, the current flow needs rearchitecting even more.

I think this will be fixed by @sarav's refactorings of the intent flow.

### LINKED ISSUE #234: [3.0.1] Killing the application makes it unusable thereafter
Coming from #217 4):

Start GoIV (with "Launch Pokemon Go" disabled to make it faster), kill it. Restart it, kill it again and it will crash.

It seems the crash itself is fixed. However, it would look as if GoIV stops working if you kill it just once: the IV button won't do anything and battery saver mode won't trigger detection. Only way to get it working again (from what I tried) was reinstalling.

sarav:
PR is not yet complete. Making a request so people can see the progress and make comments. There are tiny hacks (accessing public variables, making unnecessary dependencies, etc) to keep each commit small enough to review. When the PR is done, most of those hacks should go away.

- [x] Move the screen capture code out of MainActivity
- [x] Move the periodic scan of the screen into the service
- [x] Remove unnecessary Intent passing to take a screenshot and process it.

I'm stuck at the last one because I can't (yet) find the code that makes the (IV) button disappear.

Did some quick tests and it works so far.

This should address #203 -- you can already see how the timer stopping isn't times to some public variable. So, we should at least not get any timer races.
EDIT (by @Blaisorblade): fixes #203 and fixes #234.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -12,10 +12,12 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.IntentFilter;
 import android.content.SharedPreferences;
+import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.PixelFormat;
 import android.net.Uri;
 import android.os.Build;
+import android.os.Handler;
 import android.os.IBinder;
 import android.provider.MediaStore;
 import android.support.v4.content.LocalBroadcastManager;
@@ -39,6 +41,8 @@ import android.widget.Toast;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Timer;
+import java.util.TimerTask;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -81,6 +85,14 @@ public class Pokefly extends Service {
     private DisplayMetrics displayMetrics;
     ClipboardManager clipboard;
     private SharedPreferences sharedPref;
+    private ScreenGrabber screen;
+
+    private Timer timer;
+    private int areaX1;
+    private int areaY1;
+    private int areaX2;
+    private int areaY2;
+
 
     private boolean infoShownSent = false;
     private boolean infoShownReceived = false;
@@ -308,8 +320,56 @@ public class Pokefly extends Service {
             createIVButton();
             createArcPointer();
             createArcAdjuster();
+            /* Assumes MainActivity initialized ScreenGrabber before starting this service. */
+            if (!batterySaver) {
+                screen = ScreenGrabber.init(null, null, null);
+                startPeriodicScreenScan();
+            }
         }
+
         return START_STICKY;
+    }
+
+    private void startPeriodicScreenScan() {
+        areaX1 = Math.round(displayMetrics.widthPixels / 24);  // these values used to get "white" left of "power up"
+        areaY1 = (int) Math.round(displayMetrics.heightPixels / 1.24271845);
+        areaX2 = (int) Math.round(displayMetrics.widthPixels / 1.15942029);  // these values used to get greenish color in transfer button
+        areaY2 = (int) Math.round(displayMetrics.heightPixels / 1.11062907);
+        final Handler handler = new Handler();
+        timer = new Timer();
+        TimerTask doAsynchronousTask = new TimerTask() {
+            @Override
+            public void run() {
+                handler.post(new Runnable() {
+                    public void run() {
+                        scanPokemonScreen();
+                    }
+                });
+            }
+        };
+        timer.schedule(doAsynchronousTask, 0, 750);
+    }
+
+    /**
+     * scanPokemonScreen
+     * Scans the device screen to check area1 for the white and area2 for the transfer button.
+     * If both exist then the user is on the pokemon screen.
+     */
+    private void scanPokemonScreen() {
+        Bitmap bmp = screen.grabScreen();
+        if (bmp == null) {
+            return;
+        }
+
+        if (bmp.getHeight() > bmp.getWidth()) {
+            boolean shouldShow = bmp.getPixel(areaX1, areaY1) == Color.rgb(250, 250, 250) && bmp.getPixel(areaX2, areaY2) == Color.rgb(28, 135, 150);
+            Intent showIVButtonIntent = Pokefly.createIVButtonIntent(shouldShow);
+            /* We still need an Intent here because this is executing in the timer Thread context.
+             * Not sure we can create UI from timer threads */
+            LocalBroadcastManager.getInstance(Pokefly.this).sendBroadcast(showIVButtonIntent);
+            //SaveImage(bmp,"everything");
+        }
+        bmp.recycle();
     }
 
     private boolean infoLayoutArcPointerVisible = false;
@@ -332,6 +392,9 @@ public class Pokefly extends Service {
 
     @Override
     public void onDestroy() {
+        if (!batterySaver) {
+            timer.cancel();
+        }
         super.onDestroy();
         if (IVButton != null && IVButtonShown) windowManager.removeView(IVButton);
         hideInfoLayoutArcPointer();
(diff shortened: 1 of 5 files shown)


sarav:
@Blaisorblade FYI
@nahojjjen  -- can you point me to where the IV button is hidden once we press it? The OnClick doesn't seem to have any code to hide the button,

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public void exit() {
+        if (mProjection != null) {
+            mImageReader = null;
+            mProjection.stop();
+            mProjection = null;
+            rawDisplayMetrics = null;
+            displayMetrics = null;
+            instance = null;
```
Blaisorblade: Nitpick: I am a bit concerned about an instance method clearing static state—in part because I've been bitten by bugs in similar scenarios (that don't apply here), in part simply because it's asymmetric. BTW, `OCRHelper.exit()` doesn't do that... which might actually be part of the crashes we've seen O_O.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -303,6 +361,7 @@ private void hideInfoLayoutArcPointer() {
 
     @Override
     public void onDestroy() {
+        timer.cancel();
         super.onDestroy();
```
Blaisorblade: This reminds me: `super.onDestroy` should probably be last, it's often a safer default (no matter what Android docs tell)—I've dug some discussions about it, can find it again.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -279,10 +291,56 @@ public int onStartCommand(Intent intent, int flags, int startId) {
             createIVButton();
             createArcPointer();
             createArcAdjuster();
+            /* Assumes MainActivity initialized ScreenGrabber before starting this service. */
```
Blaisorblade: That sounds... potentially fishy.  Shouldn't we call some `getInstance()` and have it fail if no initialization had been done?

sarav: This is one of the temporary hacks I was referring to in the PR text. It'll go away.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Agreed. But separate commit to fix that. I added timer.cancel() first because I'm trying to avoid timer trying to grab the screen while we exit() the ScreenGrabber.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
sarav: I look at it the other way: If you don't have access to the instance, you don't get to call exit(). If exit() was a static function, then anyone could call it. I obviously have to clear out "instance" from there though since if I don't, then I can't call .init() again.

sarav:
I think it's in a good state to test and merge. I tested it for normal mode and it works fine for me after trying a few pokemon. Battery mode works fine through Google Now on tap. I can't check the screenshot method because it never worked for me. But if Google Now works, that should work for everyone it worked before.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Actually, no, this isn't one of those hacks. This is intentional. I couldn't move it to Pokefly because getting the MediaProjection mProjection object needs some intent handling and also showing the "Grant permissions" stuff in the UI.

Pokefly has no UI. So, I couldn't move it there. I'm not sure if mProjection can be passed through an intent. If it can, then we can remove this assumption.

sarav:
I haven't yet fixed the battery saver mode that watches for screenshots because it also needs to be refactored out of MainActivity. It's also broken on a lot of devices (like mine) and needs to be rewritten (I'll open an issue for that maybe with how to fix it). Since I couldn't test it and it needs to be rewritten, I didn't clean that up yet to avoid using Intents. That's for later I guess.

If there are no issues with this PR, I'd like to merge it.

farkam135:
LGTM  
@Blaisorblade @nahojjjen Please check, whoever is the last to check please merge.

sarav:
Not sure if it's placebo, but it feels much quicker between pressing the IV button to the dialog showing up.

nahojjjen:
I havent tested it, but reading through the code it seems to make sense.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: I'm fine with this behavior and the assumption, I'd just prefer the source to read different with `getInstance()` to show it more clearly. But not so important.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -363,11 +353,7 @@ private void scanPokemonScreen() {
 
         if (bmp.getHeight() > bmp.getWidth()) {
             boolean shouldShow = bmp.getPixel(areaX1, areaY1) == Color.rgb(250, 250, 250) && bmp.getPixel(areaX2, areaY2) == Color.rgb(28, 135, 150);
-            Intent showIVButtonIntent = Pokefly.createIVButtonIntent(shouldShow);
-            /* We still need an Intent here because this is executing in the timer Thread context.
-             * Not sure we can create UI from timer threads */
-            LocalBroadcastManager.getInstance(Pokefly.this).sendBroadcast(showIVButtonIntent);
-            //SaveImage(bmp,"everything");
+            setIVButtonDisplay(shouldShow);
```
Blaisorblade: 👍 ❤️  (for whole commit).

Blaisorblade:
I'm doing the last review round.

Blaisorblade:
Merged. Congrats!

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -1102,6 +1172,74 @@ private void showInfoLayout() {
...
+     * Performs OCR on an image of a pokemon and sends the pulled info to PokeFly to display.
+     *
+     * @param pokemonImage The image of the pokemon
+     * @param filePath     The screenshot path if it is a file, used to delete once checked
+     */
+    private void scanPokemon(Bitmap pokemonImage, String filePath) {
+        //WARNING: this method *must* always send an intent at the end, no matter what, to avoid the application hanging.
+        Intent info = Pokefly.createNoInfoIntent();
+        if (ocr == null) {
+            Toast.makeText(Pokefly.this, "Screen analysis module not initialized", Toast.LENGTH_LONG).show();
```
Blaisorblade: Either the above warning is no more needed (because of this PR), or this path crashes the app. Not that you introduced it, but maybe you fixed it?

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -1102,6 +1172,74 @@ private void showInfoLayout() {
...
+
+    /**
+     * scanPokemon
+     * Performs OCR on an image of a pokemon and sends the pulled info to PokeFly to display.
+     *
+     * @param pokemonImage The image of the pokemon
+     * @param filePath     The screenshot path if it is a file, used to delete once checked
+     */
+    private void scanPokemon(Bitmap pokemonImage, String filePath) {
+        //WARNING: this method *must* always send an intent at the end, no matter what, to avoid the application hanging.
```
Blaisorblade: I mean *this* warning.

sarav:
Phew! Thanks guys. I didn't want to rebase or redo it because I have a busy week ahead. Glad it got merged.

Blaisorblade:
Given the importance of this PR, I tried to fast-track it, so to say.

PR #250, opened 2016-08-28
Tweak and refactor UI

Blaisorblade:


app/src/main/java/com/kamron/pogoiv/DevMethods.java
```diff
@@ -0,0 +1,67 @@
...
+                    poke.candyEvolutionCost = 50;
+                }
+                if (numberInEvoLine == 1){
+                    poke.candyEvolutionCost = -1;
+                }
+            }
+            if (evoLine.size()==1){
+                poke.candyEvolutionCost = -1;
+            }
+            */
```
Blaisorblade: 1. Why is this code commented out?
2. This PR adds it as part of the code. I'm okay with having it separate, I'm not OK with shipping it.

app/src/main/java/com/kamron/pogoiv/IVScanResult.java
```diff
@@ -138,19 +138,19 @@ public void addIVCombination(int attackIV, int defenseIV, int staminaIV) {
 
     /**
      * Checks if the previous scanned pokemon can be the same pokemon as the one scanned 2 scans ago
-     * checks if newer scan has same or higher level, and same or better evolution. (because pokemon cant de-level or devolve)
+     * checks if newer scan has higher level, and same or next evolution. (because pokemon cant de-level or devolve)
```
Blaisorblade: I'd say you can refine a pokemon if you *either* powered it up *or* evolved it. Unless you're sure that evolving it can never lead to refinements.

app/src/main/java/com/kamron/pogoiv/DevMethods.java
Blaisorblade: Re 2: @nahojjjen points out ProGuard would drop this, so that's OK.
Re 1: he gives reasons, they should be documented.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -556,6 +562,7 @@ public void incrementLevel() {
 
     @OnClick(R.id.btnIncrementLevelExpanded)
     public void incrementLevelExpanded() {
+        Log.d("btn increment pushed", "in pokefly");
```
Blaisorblade: Should this stay in?

app/src/main/java/com/kamron/pogoiv/DevMethods.java
```diff
@@ -0,0 +1,67 @@
...
+        //to create the evolutions
+        for (Pokemon poke : pokeCalculator.pokedex){
+            ArrayList<Pokemon> evoLine = pokeCalculator.getEvolutionLine(poke);
+            int numberInEvoLine =1;
+            for (int i = 0; i<evoLine.size(); i++){
+                System.out.println("poke:" + poke.name + " evoLine size: " + evoLine.size());
+                if (poke.name.equals(evoLine.get(i).name)){
+                    numberInEvoLine = i;
+                }
+            }
```
Blaisorblade: Please explain right here (before the commented out code) *why* this is commented out. You need something like:

> This is commented out because the evolution cost variable is final in the running build, but the method rewrites the cost, so it gives a compile warning if it's not commented out.

No less information will do. Rationale: ensure people who clean up dead code know what to do here.

app/src/main/java/com/kamron/pogoiv/IVScanResult.java
```diff
@@ -150,11 +150,9 @@ public boolean canThisScanBePoweredUpPreviousScan() {
...
             Pokemon p2 = scanContainer.twoScanAgo.pokemon;
 
-            if (p1scan.estimatedPokemonLevel > p2scan.estimatedPokemonLevel) { //later scan must have higher or same level
-                if (p1.number == p2.number || p1.isInNextEvolution(p2)) { // either same species, or 2 scans ago is an evolution of previous scan
-                    return true;
-                }
-            }
+            boolean pokemonHasLeveledUp =p1scan.estimatedPokemonLevel > p2scan.estimatedPokemonLevel;
+            boolean isEvolved = p1.isInNextEvolution(p2);
+            return (pokemonHasLeveledUp || isEvolved) ;
```
Blaisorblade: No: A pokemon needs to be compatible on both fronts, and improved on at least one. You really need something equivalent to
`(pokemonHasLeveledUp || isEvolved) && (p1scan.estimatedPokemonLevel >= p2scan.estimatedPokemonLevel) && (p1.number == p2.number || p1.isInNextEvolution(p2))`.

app/src/main/java/com/kamron/pogoiv/IVScanResult.java
```diff
@@ -152,7 +152,11 @@ public boolean canThisScanBePoweredUpPreviousScan() {
 
             boolean pokemonHasLeveledUp =p1scan.estimatedPokemonLevel > p2scan.estimatedPokemonLevel;
             boolean isEvolved = p1.isInNextEvolution(p2);
-            return (pokemonHasLeveledUp || isEvolved) ;
+
+            boolean somethingImproved = (pokemonHasLeveledUp || isEvolved);
+            boolean isSameOrHigherLevel = (p1scan.estimatedPokemonLevel >= p2scan.estimatedPokemonLevel);
+            boolean isSameOrHigherEvolution = (p1.number == p2.number || p1.isInNextEvolution(p2));
```
Blaisorblade: Good. We could reuse `isEvolved`, but merging anyway.

PR #258, opened 2016-08-28
Markdown LICENSE.md

DJCrapsody:
Nitpick: LICENSE instead of License (sorry OCD mode on)
Update with markdown, courtesy of @jnrbsn, so it's prettier (like I said, OCD)


nahojjjen:
Are you only renaming the file, or did you actually delete something? Github says row additions/deletions in this commit is +596-674 which ends up with 78 missing lines.

DJCrapsody:
I changed the entire file, formatted with markdown instead of using the plain template provided by GitHub. You can see from the rich diff intead of source diff. There are some formatting changes done so some lines were added, some lines were removed, causing 78 missing lines (mostly due to *bad* formatting)

Blaisorblade:
Ah, you took it from https://gist.github.com/jnrbsn/708961... which already explains that we're allowed to reformat the GPL (https://gist.github.com/jnrbsn/708961#gistcomment-881408).

DJCrapsody:
Yes we are allowed to reformat as long no words were changed :)

Blaisorblade:
Looks reasonable to me. @nahojjjen you can merge if you find it OK.

PR #262, opened 2016-08-28
Nahojjjen various tweaks

nahojjjen:
Small crashfixes and various tweaks, the individual commits should be pretty clear.


app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
```diff
@@ -29,7 +29,10 @@ public PokeInfoCalculator(String[] namesArray, int[] attackArray, int[] defenceA
     }
 
     public Pokemon get(int number) {
-        return pokedex.get(number);
+        if (number >= 0){
+            return pokedex.get(number);
+        }
+        return null;
```
harkin: why would this ever be called with a negative number? Having a possibility of `null` being returned means the return value should always be null checked before being used which seems like a lot of code clutter to me. 

the exception thrown seems fine to me, callers should be validating that they're doing something sane IMO

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
```diff
@@ -206,6 +209,48 @@ public CPRange getCpRangeAtLevel(Pokemon pokemon, int lowAttack, int lowDefense,
     }
 
     /**
+     * Get the combined cost for evolving all steps between two pokemon, for example the cost from caterpie -> metapod is 12,
+     * caterpie -> butterfly is 12+25 = 27
```
harkin: 12+50

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -659,6 +664,8 @@ private void populateResultsBox(IVScanResult ivScanResult) {
         }
         setResultScreenPercentageRange(ivScanResult); //color codes the result
         adjustSeekbarForPokemon(ivScanResult);
+
+        Log.d("Pokefly","populateAdvInfo called from populateResultBox");
```
Blaisorblade: Should this stay in?

app/src/main/res/layout/dialog_results.xml
```diff
@@ -343,6 +343,7 @@
 
         <include layout="@layout/expanded_results_box"
             android:id="@+id/expandedResultsBox"
+            android:visibility="gone"
```
Blaisorblade: But this seems insufficient—shouldn't we also change the arrow in `resultsMoreInformationArrow`? Amazingly, that seems the only needed change, the code doesn't need further updates.

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
```diff
@@ -206,6 +209,48 @@ public CPRange getCpRangeAtLevel(Pokemon pokemon, int lowAttack, int lowDefense,
...
+     * @param start which pokemon to start from
+     * @param end the end evolution
+     * @return the combined candy cost for all required evolutions
+     */
+    public int getCandyCostForEvolution(Pokemon start, Pokemon end){
+        Pokemon devolution = get(end.devoNumber);
+        Pokemon dedevolution = null;
+        if (devolution != null){ //devolution must exist for there to be a devolution of the devolution
+            dedevolution =  get(devolution.devoNumber);
```
Blaisorblade: @harkin this is the potential call of `get(-1)`, but I agree I prefer the error-checking for `-1` here.

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
nahojjjen: When getting a devolution, if a pokemon doesnt have one, it will call get(-1)

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
nahojjjen: Yea, thats true. Worth fixing?

app/src/main/res/layout/dialog_results.xml
nahojjjen: Changed

nahojjjen:
Its a butterfly, i doubt its even a pokemon. Nintendo has to convince me

app/src/main/res/layout/dialog_results.xml
DJCrapsody: Current implementation of arrow will display an emoji in textview... can we swap them out to 25B4 and 25B8 instead? Smaller triangle but none of them will turn into emojis.

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
harkin: in that case the caller should be doing the bounds checking IMO

DJCrapsody:
http://bulbapedia.bulbagarden.net/wiki/Butterfree_(Pokémon)

This butterfly can shoot solarbeam. It's definitely a pokemon alright

app/src/main/res/layout/dialog_results.xml
Blaisorblade: Weren't we using our own font? That should ensure consistent results... and the arrows work correctly for me. I'm guessing they're using some system font that changes across Android versions.

DJCrapsody: Unicode will turn into their emoji counterparts on most devices. And it isn't pleasant looking at one Unicode symbol and one emoji.

Blaisorblade:
Wait until #264 is re-merged before fixing merge conflicts again ;-).

app/src/main/res/layout/dialog_results.xml
Blaisorblade: > Unicode will turn into their emoji counterparts on most devices. And it isn't pleasant looking at one Unicode symbol and one emoji.

That's not a very convincing explanation—Unicode doesn't do that.

Blaisorblade: @DJCrapsody So apparently what we're using is U+25B6 and U+25BC. Those are both in the same block as the ones you want. Can you file a separate issue so we can figure out what's going on?

app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
harkin: or if you want to keep this it'd be nice to annotate the method as `@Nullable` so callers can see it can return `null` in their IDE

Blaisorblade:
We agreed to specify `get` as @nahojjjen proposed. The issues with arrows are a separate matter. Hence, merging.

PR #267, opened 2016-08-28
Add Travis-based continuous integration

### LINKED ISSUE #254: Add Travis integration
I started testing it in my fork: https://travis-ci.org/Blaisorblade/GoIV
This only ensures the code builds, for both pushes and PRs. We could try adding tests but that's not easy.

Blaisorblade:
This PR proposes to add a first form of continuous integration—this ensures that each PR compiles before merging (though doesn't enforce that), both alone and after merging with `master`.

In the future we might also try running some automated tests, but we'd have to add some.

This can only be enabled in Travis by @farkam135 (unless he gives somebody else higher privileges). I've tried out this code in my repository, results are visible in https://travis-ci.org/Blaisorblade/GoIV/builds.

Also, this makes sense if we agree to actually keep the build green.

Fixes #254.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,6 @@ jdk:
   - oraclejdk8
 
 script:
-   - ./gradlew assembleDebug
+   - ./gradlew assembleDebug testOnlineDebugUnitTest testOfflineDebugUnitTest
      # build  # Would try need the online version, that seems to need fabric IDs.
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Merging this on general agreement/silence—so we just need @farkam135 to enable the integration.
Instructions:
https://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI%3A

PR #269, opened 2016-08-28
Correct upgrade cost after lvl 26.

clvv:
Cost information taken from http://pokemongo.gamepress.gg/power-up-costs

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
+++ b/app/src/main/java/com/kamron/pogoiv/PokeInfoCalculator.java
@@ -116,11 +116,26 @@ public class PokeInfoCalculator {
             } else if (estimatedPokemonLevel > 10.5 && estimatedPokemonLevel <= 20.5) {
                 neededCandy += 2;
                 neededStarDust += 1000 + (rank * 300);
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Oh, thanks for taking a look! This looks like it matches better that table (I've checked a few entries). I'd like a second review on this though. (I might check better myself later).

hlplulustg:
The upgrade cost was based on [http://www.serebii.net/pokemongo/combatpoints.shtml](http://www.serebii.net/pokemongo/combatpoints.shtml) which relied on [https://www.reddit.com/r/TheSilphRoad/comments/4sa4p5/stardust_costs_increase_every_4_power_ups/](https://www.reddit.com/r/TheSilphRoad/comments/4sa4p5/stardust_costs_increase_every_4_power_ups/). Calculators used them, [https://buimichael.github.io/pokemongo-Stardust-Calculator/](https://buimichael.github.io/pokemongo-Stardust-Calculator/) still does. The table mentioned by @clvv uses data dump information.

Blaisorblade:
I see. This is certainly worth a changelog entry @DJCrapsody !

PR #238, opened 2016-08-27
Modified the autoupdater

rhari991:
Is a work in progress
1) Removed the EventBus library as it was not being used anywhere else in the project. Replaced it with a BroadcastReceiver.
2) The download should be carried out in a service now.

The apk file is not being started after the download has been completed. Going to bed, will finish this up once I am awake.

EDIT: Fix #222.

PR #270, opened 2016-08-28
Rewrote the auto updater

### LINKED ISSUE #222: Updater won't run in background
Is the updater working correctly? It won't continue downloading unless I keep the application on the foreground, switching it to the background will make the download stop, trying to continue it will just get a new copy downloaded, ended up with 8 or so update.apk files, several corrupted. And didn't run the apk installer after it finished downloading.

rhari991:
Fixes #222 
Update of PR #238 

The mentioned issues should all be fixed. The checking for update, cancelling and going back to MainActivity to get another dialog was not intentional, but it should work that way if auto-update is enabled, right ?


Blaisorblade:
> The checking for update, cancelling and going back to MainActivity to get another dialog was not intentional, but it should work that way if auto-update is enabled, right ?

Regarding the old behavior I saw (*without* having yet tested the new one), no: on auto-update the check should start when I start the main activity, not just when I check manually (which is what I observed). Then I can dismiss the dialog or update. If I don't update, then go to settings, then check for update, I should again get one dialog, and not an extra one when I go back to main.

app/src/main/res/values/strings.xml
```diff
@@ -68,4 +68,8 @@
     <string name="stamina_hp">HP</string>
     <string name="missing_inputs">Missing data. Please fill CP and HP inputs.</string>
     <string name="ivtext_no_possibilities">There are no IV possible combinations for this input. Please double check the information.</string>
+    <string name="update_check_failed">An error occured while checking for an update. Please check your internet connection</string>
```
Blaisorblade: `s/internet/Internet/`

rhari991:
I fixed that behaviour, it should now show a dialog only once when manually checking for an update.

The second dialog that you see when coming to the MainActivity is not triggered by the update check in the settings page - It is triggered by a separate auto-update check when you reenter the MainActivity. It will be triggered even if you have not pressed anything in the settings page. I think this is better than the current system of checking for updates only when starting up, as it will notify the user faster when a new update is available.

Blaisorblade:
Fixing conflicts with master and testing.

Blaisorblade:
Closing and reopening with conflicts fixed.

PR #279, opened 2016-08-29
Implement drawable for expand/collapse

### LINKED ISSUE #276: Triangle for Evolution displays as emoji
Alt Text

Will try out another set of Unicode symbols that wouldn't get converted to emoji in certain devices.

DJCrapsody:
Looks prettier than having a unicode in a textview.

EDIT: fixes #276.


DJCrapsody:
fixes https://github.com/farkam135/GoIV/issues/276

Blaisorblade:
Thanks! I'd like a review by @oskomonzon here.

oskomonzon:
@Blaisorblade I apporve of this change. Let's have @nahojjjen giving it a look.

The alingment seems a little off on my device, I'm not sure if that's intended (and a matter of personal taste) or it's just something that happens on my screen density (Not home at the moment, I can't test it on more devices).

:+1:

Blaisorblade:
@oskomonzon IIUC, that's because there's embedded padding in all the `png`s? I guess all the padding should be done in the XML instead (and below there should be none).

DJCrapsody:
I trimmed all paddings on Android Asset Studio. Will check on the arrows again. I took the 18dp reference from the previous text based icon, maybe I will try 24dp instead.

Blaisorblade:
Re paddings, I just looked at the PNGs in GitHub so it might well be nonsense.

PR #281, opened 2016-08-29
Send non-Crashlytics logs to logcat at least in releases

Blaisorblade:
This could be helpful to users of offline versions that have problems and try to use logcat on it. @harkin thought this was worth of discussion (and I misunderstood the logic when originally committing), so here's a PR to discuss.

### DIFF
--- a/app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
+++ b/app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
@@ -18,9 +18,8 @@ public class CrashlyticsWrapper {
 
         @Override
         protected void log(int priority, String tag, String message, Throwable t) {
-            // there is no logging of crash reports in offline builds
-            // XXX we use error level but should look at the priority.
-            Log.e(tag, message, t);
+            // there is no online logging of crash reports in offline builds
+            CommonLogger.log(priority, tag, message, t);
         }
     }
 }
--- a/app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
+++ b/app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
@@ -1,6 +1,7 @@
 package com.kamron.pogoiv;
 
 import android.content.Context;
+import android.util.Log;
 
 import timber.log.Timber;
 
@@ -18,6 +19,8 @@ public class CrashlyticsWrapper {
         @Override
         protected void log(int priority, String tag, String message, Throwable t) {
             // there is no logging of crash reports in offline builds
+            // XXX we use error level but should look at the priority.
+            Log.e(tag, message, t);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
```diff
@@ -18,6 +19,8 @@ public CrashReportingTree(Context context) {
         @Override
         protected void log(int priority, String tag, String message, Throwable t) {
             // there is no logging of crash reports in offline builds
+            // XXX we use error level but should look at the priority.
+            Log.e(tag, message, t);
```
harkin: This will log everything as an error regardless of the actual log level.

You probably want to do `Log.println(priority, tag, message);` and concatenate the stacktrace in yourself. 

Or maybe something like: 
```
        if (priority == Log.ASSERT) {
          Log.wtf(tag, message);
        } else {
          Log.println(priority, tag, message);
        }
```

There is a character limit for logs too though I don't know much about it

app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java
```diff
@@ -42,6 +43,9 @@ protected void log(int priority, String tag, String message, Throwable t) {
                 } else if (!TextUtils.isEmpty(message)) {
                     Crashlytics.log(message);
                 }
+            } else {
+                // XXX we use error level but should look at the priority.
+                Log.e(tag, message, t);
```
harkin: same comment

harkin:
I don't have a good view of what actually is logged in the app. I think it's worth limiting what gets put to logcat (and perhaps crashlytics) to things in the warn/error/assert bracket

app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
Blaisorblade: What I didn't know: does Timber uses Android log priorities or other constants?

Blaisorblade:
> I don't have a good view of what actually is logged in the app. I think it's worth limiting what gets put to logcat (and perhaps crashlytics) to things in the warn/error/assert bracket

Just grep for `Timber.`, you'll see it's only `Timber.e` being used for pretty bad exceptions, which is why I hacked things to use `Log.e` (that was still very lazy on my part, sorry).

app/src/offline/java/com/kamron/pogoiv/CrashlyticsWrapper.java
harkin: It uses the android priorities

Blaisorblade: OK, I'll revise and deduplicate this :-).

Blaisorblade:
Here's a user who looked at his logcat:
https://github.com/farkam135/GoIV/issues/298#issuecomment-243430045

Blaisorblade:
And BTW, we probably want this also for users of online release versions.

Blaisorblade:
> And BTW, we probably want this also for users of online release versions.

The user in #298 wants this too.

PR #282, opened 2016-08-29
added a basic unit test

harkin:
This can be run with `./gradlew testOfflineDebugUnitTest` or `./gradlew testOnlineDebugUnitTest`

`./gradlew test` cannot be used because of the crashlyitcs id needed for online release builds


Blaisorblade:
BTW that test can also be run in Android Studio, which actually displays individual results. Adding this to my Travis PR.

harkin:
oh yeah, you can run them all through AS too, or get more specific about what you want to run from the command line too

PR #285, opened 2016-08-29
don't obfuscate release builds

harkin:


Blaisorblade:
Obfuscation removes line numbers in stack traces, which is annoying. @harkin explained this only saves 1-2 KBs (which is irrelevant), and I'm happy to trust this without numbers.
Also, even the Android system libraries have line numbers!

Merging.

harkin:
I was guesstimating, the bigger saving is that obfuscation will rename classes and packages so `com.kamron.goiv.MainActivity` might become `a.b.c.d` and you get all that saving on characters. Which isn't much and not worth worrying about

PR #286, opened 2016-08-29
Allow crashlytics to receive unobfuscated reports

rhari991:
From 
https://docs.fabric.io/android/crashlytics/dex-and-proguard.html


rhari991:
Closing as obfuscation has been removed

PR #290, opened 2016-08-29
Remember cp instead of recalculate

nahojjjen:
IVScanResult holds the scanned CP, so advanced results does not need to recalculate it later.


Blaisorblade:
Thanks! But wait, didn't you fix this already? O_O

nahojjjen:
Github said i hadnt pushed it as a commit, so I assumed I hadnt. I can go and check though

PR #291, opened 2016-08-29
Move game logic into separate package

Blaisorblade:
Goal: separate core logic from GUI logic. That's just a step to clarify which is which.

I also hope this helps writing unit tests (though @harkin already contributed a way to mock Android, not sure how well that works).

Thoughts?
(Let's be careful with merging this and conflicts on other PRs).

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/DevMethods.java
+++ b/app/src/main/java/com/kamron/pogoiv/DevMethods.java
@@ -1,5 +1,8 @@
 package com.kamron.pogoiv;
 
+import com.kamron.pogoiv.logic.PokeInfoCalculator;
+import com.kamron.pogoiv.logic.Pokemon;
+
 import java.util.ArrayList;
 
 /**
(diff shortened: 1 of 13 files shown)


sarav:
I think this is too soon to move. We still have a bit of refactoring to finish. I think we should wait for things to settle down a bit before moving files.

Blaisorblade:
Is `git blame --follow` not enough? There are no open overlapping PRs, and in fact Git merges renames quite well. (I've rebased this commit N times over tons of changes with 0 manual intervention over `rebase master`).

The architecture I want is basically modularization: specifically `logic` can't depend on anything in the GUI package (not that I can enforce that), on widgets, etc. Declaring stuff to be a module by listing files seems harder to follow and more fragile.

sarav:
Mmm... Ok. Reasonable answers. No strong opinion.

harkin:
> though @harkin already contributed a way to mock Android 

Currently you can mock individual objects with `Mockito` for tests. I didn't set up RoboElectric which shadows all the Android classes

Blaisorblade:
That is: mocking classes takes some work, so using nothing Android-specific in that package simplifies testing?

I think it's simpler to write tests where you don't need to write your own mockins, when possible, and that should work for this package.

PR #292, opened 2016-08-29
Add default IDEA coding style as a project one

Blaisorblade:
Goal: enforce a consistent coding style across contributors.

Because the main thing that matters is consistency, I picked the Android Studio default, which is what I've been already using since ages. Consistency matters, among other things, to have readable diffs.

The only point of contention is whether to break long lines. @sarav is strongly in favor, and @nahojjjen finds it decreases readability. The default coding style does not break long lines. I dislike this default but not strongly.

Questions:
- does this allow a consistent workflow for all? Can we agree to reformat all PRs automatically that way?

Non-questions:
- which coding style is best? People have strong opinions, but the particular coding style matters much less than consistency (source: Code Complete).

WARNING: I reserve the right to delete anything debating coding style in this PR. DON'T. Open separate issues if you deeply care.

### DIFF
--- a/.idea/codeStyleSettings.xml
+++ b/.idea/codeStyleSettings.xml
@@ -1,0 +1,231 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectCodeStyleSettingsManager">
+    <option name="PER_PROJECT_SETTINGS">
+      <value>
+        <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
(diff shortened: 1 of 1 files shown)


nahojjjen:
I support this idea

Blaisorblade:
Merging with @nahojjjen's approval. Let's try out if this works.

Blaisorblade:
For the record: I'd consider having a commit hook that automates this, but I'm not sure that's easy enough to setup:
- can you get those settings on the command line?
- can you ensure all contributors have the needed tools installed?
etc. etc. I'd be happy to accept PRs if they solve all these problems.

PR #299, opened 2016-08-30
Reworked All Results UI

### LINKED ISSUE #280: New result layout is slower on many possibilities
Now we add widgets for each entry in the table. That's too slow with many possibilities, and slower than before. But note: computing them is totally fine, it's just displaying them that's the problem. Should we truncate that display after 100 possibilities or so?

oskomonzon:
Made a series of changes that fix #280 
All results layout now displays data in a RecyclerView. This is way more efficient and should reduce the choppyness of the UI in cases where there is a results overload.
Most of the work is done in a new custom Adapter, so I got rid of some of the stuff from Pokefly that shouldn't be needed anymore.
Headers now support localization (used the same strings as in other layouts, added just one new for the percentages column).
That's from the top of my head. I tweaked things here and there but nothing too crazy.

So there you have it. Pending review. :+1:

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/IVResultsAdapter.java
+++ b/app/src/main/java/com/kamron/pogoiv/IVResultsAdapter.java
@@ -1,0 +1,82 @@
+package com.kamron.pogoiv;
+
+import android.content.Context;
+import android.graphics.Color;
+import android.support.v7.widget.RecyclerView;
+import android.view.LayoutInflater;
+import android.view.View;
+import android.view.ViewGroup;
+import android.widget.LinearLayout;
+import android.widget.TextView;
+
+/**
+ * Created by OskO on 29/08/16.
+ */
+class IVResultsAdapter extends RecyclerView.Adapter<IVResultsAdapter.ResultsViewHolder> {
+    LayoutInflater layoutInflater;
+    IVScanResult mDataSet;
+
+    public IVResultsAdapter(IVScanResult ivScanResult, Context context) {
+        layoutInflater = LayoutInflater.from(context);
+        mDataSet = ivScanResult;
+    }
+
+    @Override
+    public ResultsViewHolder onCreateViewHolder(ViewGroup parent, int viewType) {
+        View view = LayoutInflater.from(parent.getContext()).inflate(R.layout.iv_results_item, parent, false);
+        return new ResultsViewHolder(view);
+    }
+
+    @Override
+    public void onBindViewHolder(ResultsViewHolder holder, int position) {
+        IVCombination currentSet = mDataSet.iVCombinations.get(position);
+        holder.resultAttack.setText(String.valueOf(currentSet.att));
+        holder.resultDefense.setText(String.valueOf(currentSet.def));
+        holder.resultHP.setText(String.valueOf(currentSet.sta));
+        holder.resultPercentage.setText(String.valueOf(currentSet.percentPerfect));
+
+        setTextColorbyPercentage(holder.resultAttack, (int) Math.round(currentSet.att * 100.0 / 15));
+        setTextColorbyPercentage(holder.resultDefense, (int) Math.round(currentSet.def * 100.0 / 15));
+        setTextColorbyPercentage(holder.resultHP, (int) Math.round(currentSet.sta * 100.0 / 15));
+        setTextColorbyPercentage(holder.resultPercentage, currentSet.percentPerfect);
+
+        if (position % 2 != 0) holder.llRvResult.setBackgroundColor(Color.parseColor("#EFEFEF"));
+    }
+
+    @Override
+    public int getItemCount() {
+        return mDataSet.iVCombinations.size();
+    }
+
+    class ResultsViewHolder extends RecyclerView.ViewHolder {
+        TextView resultAttack, resultDefense, resultHP, resultPercentage;
+        LinearLayout llRvResult;
+
+        public ResultsViewHolder(View itemView) {
+            super(itemView);
+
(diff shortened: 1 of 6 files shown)


oskomonzon:
Adding a screenshot from my trusty Weedle so you can get a feel on how it looks.
screenshot_2016-08-30-00-43-07

app/src/main/java/com/kamron/pogoiv/IVResultsAdapter.java
```diff
@@ -0,0 +1,85 @@
...
+        if(position%2!=0) holder.llRvResult.setBackgroundColor(Color.parseColor("#EFEFEF"));
+    }
+
+    @Override
+    public int getItemCount() {
+        return mDataSet.iVCombinations.size();
+    }
+
+    class ResultsViewHolder extends RecyclerView.ViewHolder{
+        /*@BindView(R.id.resultAttack) TextView resultAttack;
```
Blaisorblade: Let's drop this and auto-reformat code!

.idea/codeStyleSettings.xml
```diff
@@ -63,9 +63,7 @@
           </extensions>
         </Objective-C-extensions>
         <XML>
-          <option name="XML_KEEP_LINE_BREAKS" value="false" />
-          <option name="XML_ALIGN_ATTRIBUTES" value="false" />
-          <option name="XML_SPACE_INSIDE_EMPTY_TAG" value="true" />
+          <option name="XML_LEGACY_SETTINGS_IMPORTED" value="true" />
```
Blaisorblade: Certainly we don't want to keep editing the coding style, it should stay consistent.

Blaisorblade:
The code makes generally lots of sense—I have added a few comments but about minor issues.
I suspect most of the commits are better squashed together (except the merge) but it's not so important, especially if you're not a Git expert.

I think we can merge when those comments are addressed.

oskomonzon:
Great, thanks for the review, @Blaisorblade. I'll try to address them today if I'm able to.

.idea/codeStyleSettings.xml
oskomonzon: I... don't remember to be messing with that file. Not sure what happened there.

Blaisorblade:
BTW haven't tested, but this looks relatively robust and doesn't change much logic, so hopefully we don't need 50 people testing it to get it right (or maybe I'm naive and it breaks on different Android variants?)

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -975,14 +944,6 @@ private void setTextColorbyPercentage(TextView text, int value) {
         }
```
oskomonzon: Yes we want it in one place. And I'm betting we wanted out of Pokefly. But I decided to wait for the approach we are going to take for Pokefly rework.
Maybe the best way to deal with this case in particular is to extract it into a helper class?

oskomonzon:
It should be pretty solid, but I can easily gather some people from my local community to give it a test-run from my debug build.

.idea/codeStyleSettings.xml
Blaisorblade: I'm sure changing it is Android Studio's fault. Committing it is another matter ;-).

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: > Maybe the best way to deal with this case in particular is to extract it into a helper class?

> EDIT: I just realized that all the color code stuff is related to the IV Results IVCombinations class, so maybe we can drop it there.

Since you sent a PR with copy-pasted code (knowingly), I'm afraid I'll have to repeat the following after millions have said it.

We should NOT copy-paste code, not even temporarily, unless there's a big reason, a big warning in both places to keep them in sync, and probably even an issue to fix it. And Copy-paste = bugs and all that. Yes we already have this problem, let's not make that worse.

I think one copy is just never used and can be dropped. If that didn't work (in general), this method can be reused as is because it can be made static (1e71958567220c5a8fdd0b6bdeb77c95bc6c99f7). If even that failed, an helper class could be used (and maybe is a good idea anyway, but can be deferred).

Blaisorblade:
The travis build failed (https://travis-ci.org/farkam135/GoIV/builds/156211197) with
```
* What went wrong:
A problem occurred configuring project ':app'.
> failed to find target with hash string 'android-24' in: /usr/local/android-sdk
```

but I've looked at the logs and it seems a spurious failure O_O. @harkin have you seen those?

Anyway, triggering a rebuild. I'm afraid that's common with Travis, but hey it's free.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
oskomonzon: I'll take care of it as we discussed previously as soon as I can. As I told you earlier, I didn't want to start messing with other classes without discussing the best approach beforehand.

nahojjjen:
Works except the zebra striping - it resets to gray if its recycled offscreen, so you get only gray lines if you scroll it outside view. Im merging this and posting that as an issue though

harkin:
@Blaisorblade I haven't, but I use a different CI provider generally. Looks like it hadn't finished installing the sdk components, weird

Blaisorblade:
Network connection on Travis machines tends to be flaky — so network operations should be bracketed by `travis_retry` to retry on failure. But the install is done by travis code, so I don't think we can do much.

We'll have to just restart builds when that happens—it happens in almost every project I use Travis with.

PR #307, opened 2016-08-30
Prevent scanPokemon from hanging the app (#284)

### LINKED ISSUE #284: `scanPokemon` might hang the application
This is a regression introduced (I think) in https://github.com/farkam135/GoIV/commit/9ac1355ba784d1b85700d110edfd775c8421c377 from #262.

@sarav's #248 changed the code, but we have still Pokefly sending an intent to itself, and flags not getting changed if that doesn't happen.

I'm not 100% sure there's a bug, but if there is no bug it's by luck, not by design. And luck isn't reliable. We should refactor this.

- To fix this bug, we could send again an empty intent when the OCR isn't initialized.

   But the new logic uses default values (which I find worse), so an empty intent is not something we're handling.

  Moreover, those default values end up in the dialog instead of having something *actually* missing there.

- Alternatively, now that @sarav's fixed the root cause for the OCR being not initialized, we could drop that code for ocr == null. Recent releases have shown that handling problematic conditions badly is more painful than crashing—crashing resets the app to a working state, hiding the problem makes it harder to reset the app, with users claiming they need to reboot their phone (though I still don't believe that's necessary).

Blaisorblade:
Revert "Warning toast in case of error case or ocr slow
init" (9ac1355ba784d1b85700d110edfd775c8421c377). As discussed in #284,
this null check should be unnecessary; if it is necessary, crashing
might be better than mitigating the bug and hang.

Fix #284.
Don't merge yet since we're apparently still debugging it. Ping @sarav.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1158,20 +1158,15 @@ public class Pokefly extends Service {
     private void scanPokemon(Bitmap pokemonImage, String filePath) {
         //WARNING: this method *must* always send an intent at the end, no matter what, to avoid the application hanging.
         Intent info = Pokefly.createNoInfoIntent();
-        if (ocr == null) {
-            Toast.makeText(Pokefly.this, "Screen analysis module not initialized", Toast.LENGTH_LONG).show();
-        } else {
-            try {
-                ocr.scanPokemon(pokemonImage, trainerLevel);
-                if (ocr.candyName.equals("") && ocr.pokemonHP == 10 && ocr.pokemonCP == 10) { //the default values for a failed scan, if all three fail, then probably scrolled down.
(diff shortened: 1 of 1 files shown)


nahojjjen:
Shouldn't the try-catch also be removed, if we want it to crash? As it is now, if it fails, the button when pressed will just do nothing, without crashing?

Blaisorblade:
> Shouldn't the try-catch also be removed, if we want it to crash? As it is now, if it fails, the button when pressed will just do nothing, without crashing?

Relevant question, but it's a try...finally, so exceptions are propagated. Maybe GoIV should take extra effort to ensure it dies on an exception there—I haven't checked what the callers do.

nahojjjen:
Seems like the calling methods arent using try catch, so the exception should cause a crash

sarav:
Agree that the try catch is pointless. It'll crash anyway, no point in sending the intent if it's going to crash. But this is a good start.

PR #308, opened 2016-08-30
Pokemon dropdown selector / autocomplete textview on inputscreen

nahojjjen:
The start of the solution for #130 . 

This allows the user to pick pokemon by either the dropdown, which is now only 1 evolution chain instead of the entire pokedex, or by searching for the pokemon name in an autocompletetextview.

I know the button is not the nicest looking thing, but I'd like people to test the functionality.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -651,6 +651,7 @@ public class Pokefly extends Service {
     public void toggleSpinnerVsInput() {
         if (autoCompleteTextView1.getVisibility() == View.GONE) {
             autoCompleteTextView1.setVisibility(View.VISIBLE);
+            autoCompleteTextView1.requestFocus();
             inputScreenPokemonSpinner.setVisibility(View.GONE);
         } else {
             resetToSpinner();
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -38,6 +38,7 @@ import android.view.WindowManager;
 import android.widget.AdapterView;
 import android.widget.ArrayAdapter;
 import android.widget.AutoCompleteTextView;
+import android.widget.Button;
 import android.widget.CheckBox;
 import android.widget.EditText;
 import android.widget.ImageView;
@@ -221,6 +222,10 @@ public class Pokefly extends Service {
     //results pokemon picker auto complete
     @BindView(R.id.autoCompleteTextView1)
     AutoCompleteTextView autoCompleteTextView1;
+
+    @BindView(R.id.pokePickerToggleSpinnerVsInput)
+    Button pokePickerToggleSpinnerVsInput;
+
 
 
     private String pokemonName;
@@ -586,6 +591,8 @@ public class Pokefly extends Service {
         pokeAdapter = new PokemonSpinnerAdapter(this, R.layout.spinner_pokemon, pokeCalculator.pokedex);
         pokemonList.setAdapter(pokeAdapter);
 
+        initializePokemonAutoCompleteTextView();
+
         pokeEvolutionAdapter = new PokemonSpinnerAdapter(this, R.layout.spinner_evolution, new ArrayList<Pokemon>());
         extendedEvolutionSpinner.setAdapter(pokeEvolutionAdapter);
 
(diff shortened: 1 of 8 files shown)


oskomonzon:
Looks good, great job :+1:

nahojjjen:
Don't merge, I just noticed there's a crashbug if you search a pokemon, then close, then try to open a pokemon with less evolution index than the previous typed.

I know where the problem is, and I can fix it by switching from checking the boolean in the if-else that represents if the spinner or the textbox is used, to looking at the visibility of them. Im not at pc, so cant commit fix atm, will do tomorrow.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -688,8 +718,24 @@ public void checkIv() {
...
+            pokemon = pokeCalculator.get(userInput);
+            if (pokemon == null) { //no such pokemon was found, show error toast and abort showing results
+                Toast.makeText(this, userInput + getString(R.string.wrongPokemonNameInput), Toast.LENGTH_SHORT).show();
+                return;
+            }
+
+        }
+
+
+        Log.d("AutoCompletePokemon", "translated to: " + pokemon.name);
```
Blaisorblade: Logging still in here (also elsewhere).

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -688,8 +718,24 @@ public void checkIv() {
...
+        //below picks a pokemon from either the pokemon spinner or the user text input
+        Pokemon pokemon;
+        if (usingSpinnerPickerForPokemon) { //user picked pokemon from spinner
+            int selectedPokemon = pokemonList.getSelectedItemPosition();
+            pokemon = pokeCalculator.get(selectedPokemon);
+        } else { //user typed manually
+            String userInput = autoCompleteTextView1.getText().toString();
+            pokemon = pokeCalculator.get(userInput);
+            if (pokemon == null) { //no such pokemon was found, show error toast and abort showing results
+                Toast.makeText(this, userInput + getString(R.string.wrongPokemonNameInput), Toast.LENGTH_SHORT).show();
```
Blaisorblade: This sort of thing is bad for translator—translation requires using format strings since the pokemon name might not be the first word in all languages.

sarav:
As much as I love this feature, I think this PR should wait till the tech debt issue is handled. Otherwise, things are moving way too fast to finish that. Just my 2 cents (meaning, I won't be annoyed if you ignore this).

Blaisorblade:
> As much as I love this feature, I think this PR should wait till the tech debt issue is handled. 

I'm not sure, but the fact this is tricky enough to cause a crash suggests caution.

Blaisorblade:
Clearing the milestone for now while we decide (I haven't yet).

Blaisorblade:
After the first testing session, the feature itself mostly works, but:

- if we press the pencil button to go to "pencil mode", the next field is selected;
- when typing in the field, the input dialog is covered by the keyboard (what we had forever for CP and HP and then fixed);
- after pressing cancel, when pressing IV again "pencil mode" is still selected, with the text I wrote previously.

oskomonzon:
I've been testing it too, besides what @Blaisorblade mentioned... At least on my end it's always picking the last evolution of everything I scan.
I love the feature but I think it will require a few more passes before it's ready. I hope we manage to include it before feature lock, but if it's not on point it can wait.

Blaisorblade:
> At least on my end it's always picking the last evolution of everything I scan.

Not here. But I'm reading the code that updates the "Pokemon name:" input spinner—it's two lines of code, but I'm reading the docs and I still can't guarantee they'll work correctly (though they do the right thing there).

nahojjjen:
It doesnt select the last evolution anymore

Blaisorblade:
After discussion we agreed to postpone this to be the first PR merged in 3.2.0, before the refactorings. During 3.2 we'll have time to iron out any remaining bugs and refactor the code to be more confident it works.

kaneki1985:
Check passt

DJCrapsody:
What did you rename it to?

imapokemon4:
Numbers.

Blaisorblade:
> Maybe we could just put all 6 of the nidoran family in the dropdown?

We can detect gender rather reliably from the color—we just should use this when candy says Nidoran. See #381.

PR #304, opened 2016-08-30
Fix input dialog and expanded results box

DJCrapsody:
As this changes were based on @nahojjjen's fixInfoDialogHeight branch, this PR should supersede https://github.com/farkam135/GoIV/pull/294.

Additional things done:
* Rename "Appraisal narrowing" to "Appraisal Refining"
* Remove "(optional)" as refining is closed by default now
* Fix layout of Appraisal Refining
* Add rotating animation to appraisal refining box arrow
* Change rotating animation to happen before dialog shows/hides
* Fix alignment of Evolution spinner using negative margins
* Fix devolution CP text by checking if result CP is negative

Fixes https://github.com/farkam135/GoIV/issues/272 and fixes https://github.com/farkam135/GoIV/issues/293.

PR #294, opened 2016-08-29
Makes input smaller so hp is visible

nahojjjen:
Solves that the HP is covered for some screen densities, which makes it hard to confirm that the hp scan is correct.

Known not to work in this pull request:
The arrow for expanding the appraisal info does not change on expand, I haven't looked into how to animate drawables.

EDIT: Fixes #239.

PR #311, opened 2016-08-30
Fix input dialog and expanded results box, V3

### LINKED ISSUE #272: Typo fix when displaying CP of "involution"
Splitting from #217 

> If you choose a "devolution" from the "Evolution" combo in the Advanced information, the CP might try to show negative numbers (for a 538 Pidgeotto, choosing "Pidgey" and lowering the level will show CP as 413 (+-125).
It should be displayed as "413 (-125)" instead.

### LINKED ISSUE #293: Evolution and powerup info arrow is not animated
This is something that DJCrapsody will hopefully fix, I'm making an issue so it's not forgotten.

Blaisorblade:
I took #304, merged conflicts, and did a couple refactorings on the new code to reduce copy-pasting. Credits to @nahojjjen and @DJCrapsody for most of the work.

----
Quoting from #304:
As this changes were based on @nahojjjen's fixInfoDialogHeight branch, this PR should supersede https://github.com/farkam135/GoIV/pull/294.

Additional things done:
* Rename "Appraisal narrowing" to "Appraisal Refining"
* Remove "(optional)" as refining is closed by default now
* Fix layout of Appraisal Refining
* Add rotating animation to appraisal refining box arrow
* Change rotating animation to happen before dialog shows/hides
* <s>Fix alignment of Evolution spinner using negative margins</s> (I removed that)
* Fix devolution CP text by checking if result CP is negative

Fixes #272 and fixes #293.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -629,41 +629,31 @@ public class Pokefly extends Service {
 
     }
 
-
-    @OnClick({R.id.resultsMoreInformationText})
-    public void toggleMoreResultsBox() {
+    private void toggleVisibility(TextView expanderText, LinearLayout expandedBox) {
         int boxVisibility;
         Drawable arrowDrawable;
-        if (expandedResultsBox.getVisibility() == View.VISIBLE) {
+        if (expandedBox.getVisibility() == View.VISIBLE) {
             boxVisibility = View.GONE;
             arrowDrawable = getResources().getDrawable(R.drawable.arrow_collapse);
         } else {
             boxVisibility = View.VISIBLE;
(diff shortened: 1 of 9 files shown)


Eccenux:
Why do you think negative margins are a problem? Seems like a valid practice currently. See: http://stackoverflow.com/questions/10673503/is-it-a-bad-practice-to-use-negative-margins-in-android

Blaisorblade:
The negative margin simply didn't work for me, see screenshots in #312, and I wanted to merge the rest. I mentioned that in the original PR but forgot to write it here too.

I'm uneasy with development by trial and error, even for layouts, unless we have some evidence the end result is going to be robust. Turns out there was a padding we could set to 0 (see #316).

If a negative margin of the same amount is guaranteed to be safe, *and* we can avoid duplicating a magic number in distant places or make it maintainable, no problem.

Issue #295, opened 2016-08-30
Don't associate pokemon name with custom name on cancel


rpgrca:
1. Rename a Rattata as Raticar.
2. Open GoIV, pokemon name is recognized as "Raticate" in red.
3. Cancel.
4. Reopen GoIV, pokemon name recognized as "Raticate" in green.

Pressing Cancel should prevent GoIV from associating Raticar with Raticate. One should assume the pokemon name has been correctly guessed only if the user presses "Check IV" instead.

sarav:
I know why this happens and where the code changes should move to. I just don't know where that code actually is on the files.

The code where I store the usercorrections needs to be moved into the "on Click" event of the spinner or whatever event fires when the user actually clicks it. I don't know how to use Butterknife or even find the spinner in some layout to add an onWhatever event to it.

But it shouldn't be happening for the cancel path though. I don't think the code to store it would have been executed!

Blaisorblade:
@sarav looking at the code, I think this is due to `cachedCorrections` rather than `userCorrections` (which you're IMHO hinting at). I tried to address the TODO in #313.

Indeed, if you restart GoIV the recognized text is red again.

When I reproduce this, Step 4 means "press the IV button again".

sarav:
I think I might have hacked up the userCorrections to work around the point I made when I made the commit to save it. I still think it needs to go into some event Listener method.

Blaisorblade:
> I think I might have hacked up the userCorrections to work around the point I made when I made the commit to save it. I still think it needs to go into some event Listener method.

Failed to parse, but I hope #313 addresses that.

sarav:
Actually, I knew that the match distance goes to 0 (confident match) if we cache it. The reason I let that happen is that "if the user sees the guess and doesn't correct it, then that's an implicit okay from the user. So the app is more confident the next time that it got the guess right"

So, Raticar and Raticate-80/90 would both be guesses as Raticate by the app. If the user "Check IV" for Raticate-80/90, then the app will be confident Raticate-80/90 is Raticate the next time. But if the user opens "Raticar" and doesn't "CheckIV" it should be just as "not confident" as before.

The issue reported is that we are doing that even if the user chose not to "Check IV". We can fix that my moving this caching into where user corrections are cached too. But that'll mean the distance calculation would be done each time the user opens a Pokemon, but hits cancel -- but that's hopefully not too often.

I don't have a strong preference between what I said above and what @Blaisorblade has done. So, either is fine by me.

[...]

PR #313, opened 2016-08-31
Save corrections on spinner item selection

Blaisorblade:
Address @sarav's TODO mentioned in #295. Tested, but might be a bad idea. *Don't merge yet.*
It also conflicts with other ongoing PR #308 IIUC.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -690,8 +690,8 @@ public class Pokefly extends Service {
     }
 
     @OnItemSelected(R.id.spnPokemonName)
-    public void storeUserCorrection() {
-        Pokemon pokemon = pokeCalculator.get(pokemonList.getSelectedItemPosition());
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Actually: should the correction be saved as correct if the user doesn't change it but clicks OK? Or when he clicks OK and valid combinations are found?

Blaisorblade:
This PR is about explicit user corrections should be just explicit user corrections. I'm not sure we should save those if you correct and then click cancel, though—something that doesn't happen now and happens after this PR. So I doubt this is a good idea.

Blaisorblade:
While we debate this, closing for now. We can continue discussion here.

sarav:
I think you should squash the 2 commits. Final history don't need to see mistakes we made while coding.

I see your point about this saving user corrections before they click any button. But I don't think it's very likely that the user will change the spinner but not click "Check IV". So, I don't think it's a big deal. However, we can still address your concern -- set a flag when user selects something on the spinner. And we'll save it only if the flag is set AND we meet the existing conditions in that "if".

PR #314, opened 2016-08-31
Remember text distance in pokemon name correction cache

### LINKED ISSUE #295: Don't associate pokemon name with custom name on cancel

Blaisorblade:
Fixes #295. <s>Under testing</s> Appears to work. Review by @sarav.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1091,8 +1091,6 @@ public class Pokefly extends Service {
      * @return the likely pokemon number against the char sequence as well as the similarity
      */
     private int[] getPossiblePokemon(String poketext, String candytext) {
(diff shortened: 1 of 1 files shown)


Blaisorblade:
BTW we could use Pair instead of array for the return value.

Blaisorblade:
@sarav ping? This is orthogonal from #313, but this is the actual fix and makes more sense.

Blaisorblade:
Merged after this comment here:
https://github.com/farkam135/GoIV/issues/295#issuecomment-243823230

Issue #273, opened 2016-08-28
Small modifications on start screen


rpgrca:
Splitting from #217:

> Limit level input to 2 digits (currently any number of them can be added).

> Change instructions when selecting battery saver mode (instructions continue to say "Press IV button" when it should be "take a screenshot", might confuse people who forgot they have turned the option on and then don't see the button anywhere).

oskomonzon:
What do you guys think about a number picker instead? It may be the right UI element for this scenario.
Most users will set it up once and then only change it in increments of one at a time.

About the instructions we can add the missing string for that as a quick fix. But, depending on the ammount of features we end up adding to the app (there are a lot of cool ideas around), the best approach would be to get rid of those instructions and add a fully fledged tutorial/welcome screen/whatever. 

It would be a clean way to introduce and explain to the user new features after every update.

DJCrapsody:
Number picker from 1 to 40 seems good.

Blaisorblade:
> the best approach would be to get rid of those instructions and add a fully fledged tutorial/welcome screen/whatever.

That might be good. Development-wise, I guess that's probably better as soon as the core feature set starts stabilizing more—assuming that such a tutorial takes more effort and care to update than instructions here and there.

rpgrca:
Just checked the latest code, somehow the number picker begins in level 1 with 1 highlighted (which gives the impression it can be edited. Also, the control makes (in my case) the screen larger, so you can't get to see the links nor the version of the program, you need to scroll the page (which isn't obvious since there's no scrollbar).

![2016-08-30 02 44 31](https://cloud.githubusercontent.com/assets/15602473/18074444/0e2f3b66-6e43-11e6-88a5-21ab5e06c2e1.jpg)

Blaisorblade:
> which gives the impression it can be edited

I think that's correct—just tap on the number. That's also the standard behavior of this NumberPicker control. (https://developer.android.com/reference/android/widget/NumberPicker.html).

> Also, the control makes (in my case) the screen large

It's even worse here :-|, I hadn't realized a number picker works like that. Maybe we should reuse horizontal seekbars with +/- like elsewhere, but we'd also need a text field.

But the version could go elsewhere next to GoIV, and the links could move elsewhere.

DJCrapsody:
We can use dropdown for worse case scenario.

oskomonzon:
Or we can adjust/customize the control.

Blaisorblade:
[...]

PR #316, opened 2016-08-31
Quick fix for the Pokedex spinner and MainActivity numberpicker display issues

### LINKED ISSUE #312: Adjust Evolution spinner in expanded info box
Previous attempt was unable to fix the alignment therefore reverted.

oskomonzon:
Fixes #312.
NINJA EDIT: This PR also reduces main activity numberpicker height as discussed in #273.

I'm summoning @DJCrapsody and @nahojjjen to the discussion.

I gave a look to the code of our spinners and it turns out that we are inflating a single textbox in the adapter. Currently it's the same xml resource that we use to show the selected Pokémon, so they share their settings.

What I do here as a quick fix is to leave the TextView without margins and padding (so it aligns nicely to the right on the advanced information panel) and provide the inflated TextView instance that we use to display as each item in the dropdown with some new properties (padding, text align, gravity), so it shows its contents at the center.

Now, this works, but I'd like to make a sencond pass later and maybe rework this small layouts and adapter. To maybe display two columns of data as a means to make it easier to the eyes to quickly scan the information. That would be a column for the 'mon Pokédex number, another for the name (we can discuss how to align the texts) and maybe use a striped pattern like the results screen.

I'd like to hear what you guys think. 
Also, feel free to merge this change in the meantime because I'm a little constrained with my times lately and I would hate to not make it in time for the next release, so I'll be at peace if at least this small fix is working.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GUIUtil.java
+++ b/app/src/main/java/com/kamron/pogoiv/GUIUtil.java
@@ -1,6 +1,8 @@
 package com.kamron.pogoiv;
 
+import android.content.Context;
 import android.graphics.Color;
+import android.util.TypedValue;
 import android.widget.TextView;
 
 /**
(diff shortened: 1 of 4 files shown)


Blaisorblade:
I'm in a hurry, but I added this to the 3.1 milestone so it should get in.

Blaisorblade:
Logic-wise the code LGTM, so the main question is if the GUI guys like the look. @DJCrapsody wrote in chat this looks good to him, right?

Blaisorblade:
I edited the desc replacing "Addressing issue #312" with "Fixes #312". Only then will merging the PR close the issue.

PR #322, opened 2016-08-31
Make more strings translatable and update French translation

Remiscan:
The ATK, DEF, STA and MIN, AVG, MAX strings couldn't be translated, so I tried to change that. Here's how it looks on my phone now, in French: http://imgur.com/a/MNWYf

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -806,7 +806,7 @@ public class Pokefly extends Service {
      */
     private void populateResultsHeader(IVScanResult ivScanResult) {
         resultsPokemonName.setText(ivScanResult.pokemon.name);
-        resultsPokemonLevel.setText(getString(R.string.level) + ": " + ivScanResult.estimatedPokemonLevel);
+        resultsPokemonLevel.setText(getString(R.string.level) + " " + ivScanResult.estimatedPokemonLevel);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -806,7 +806,7 @@ private void populatePrevScanNarrowing(IVScanResult ivScanResult) {
      */
     private void populateResultsHeader(IVScanResult ivScanResult) {
         resultsPokemonName.setText(ivScanResult.pokemon.name);
-        resultsPokemonLevel.setText(getString(R.string.level) + ": " + ivScanResult.estimatedPokemonLevel);
+        resultsPokemonLevel.setText(getString(R.string.level) + " " + ivScanResult.estimatedPokemonLevel);
```
Blaisorblade: @oskomonzon Can you weigh in on this?

I also think this should be a format string (because that's always the right thing for translateable strings).

Blaisorblade:
This looks all good to me, but maybe other translators want to weigh in? Summoning also @Eremiell.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
oskomonzon: I agree, format string should be the way to go. We were a little sloppy there.

oskomonzon:
Other than the format string bit (not your fault, but it looks like a good moment to get it right). It looks good.

Blaisorblade:
Let's merge this meanwhile. Thanks @Remiscan!

PR #329, opened 2016-08-31
Load credentials from keystore.properties

imapokemon4:
I don't know how everyone else generates signed release builds, but I thought this might be helpful.

Please let me know if there is a better way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,4 @@ captures/
 # Keystore files
 *.jks
 **/*.jks
+keystore.properties
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -32,3 +32,6 @@ You can do so by following these *rules* (taken from this great [article](http:/
 * Use the imperative mood in the subject line
 * Wrap the body at 72 characters
 * Use the body to explain what and why vs. how
+
+## Generating Signed Release Builds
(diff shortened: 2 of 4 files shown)


Blaisorblade:
Restarted the build, the last failure was spurious (network problems in installing dependencies).

Blaisorblade:
Thanks for the PR!

It'd be good to add a `keystore.properties.sample` to show the format. And generally to have a description of what you're doing and your goals.

Personally, I looked into something like this so we could build a fake release version on travis (not your goal), and maybe to build one myself. If we store the key securely (which might be possible), Travis might be able to build actual releases (maybe we want a separate key though).

Finally, I don't know how @farkam135 does his release builds, and it'd be bad to break his workflow (but probably he just needs to fill in the config the right way).

We might also also need to link to a guide to generate the keystore—the tutorial in the Android docs (https://developer.android.com/studio/publish/app-signing.html) is decent, using the tools by hand IIRC is painful. (Not because it's CLI, but because `keytool` and its docs are awful).

imapokemon4:
`git push --force` to the rescue!

DJCrapsody:
LGTM

imapokemon4:
imapokemon4
  
I'm a contributor! :)

PR #336, opened 2016-09-01
Color IV based on appraisal thresholds

leppie:
Use text color based on Appraisal system.

image

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GUIUtil.java
+++ b/app/src/main/java/com/kamron/pogoiv/GUIUtil.java
@@ -10,19 +10,43 @@ import android.widget.TextView;
  */
 public class GUIUtil {
 
+    static final String RED = "#8A0808";
+    static final String ORANGE = "#DBA901";
+    static final String LIGHT_GREEN = "#A0A53F";
+    static final String GREEN = "#088A08";
     /**
-     * Sets the text color to red if below 80, and green if above
+     * Sets the text color based on IV to match the in-game appraisal system
      *
      * @param text  the text that changes color
-     * @param value the value that is checked if its above 80
+     * @param value the value that is checked (from 0 - 15)
      */
-    public static void setTextColorbyPercentage(TextView text, int value) {
-        if (value >= 80) {
-            text.setTextColor(Color.parseColor("#088A08")); //dark green
-        } else if (value >= 60) {
-            text.setTextColor(Color.parseColor("#DBA901")); //brownish orange
+    public static void setTextColorByIV(TextView text, int value) {
+        if (value == 15) {
+            text.setTextColor(Color.parseColor(GREEN));
+        } else if (value >= 13) {
+            text.setTextColor(Color.parseColor(LIGHT_GREEN));
+        } else if (value >= 8) {
+            text.setTextColor(Color.parseColor(ORANGE));
         } else {
-            text.setTextColor(Color.parseColor("#8A0808")); //dark red
+            text.setTextColor(Color.parseColor(RED));
+        }
+    }
+
+    /**
+     * Sets the text color based on total IV percentage to match the in-game appraisal system
+     *
+     * @param text  the text that changes color
+     * @param value the value that is checked
+     */
+    public static void setTextColorByPercentage(TextView text, int value) {
+        if (value > 81) { // between 80 and 82.2
+            text.setTextColor(Color.parseColor(GREEN));
+        } else if (value > 65) { // between 64.4 and 66.7
+            text.setTextColor(Color.parseColor(LIGHT_GREEN));
+        } else if (value > 50) { // between 48.9 and 51.1
+            text.setTextColor(Color.parseColor(ORANGE));
+        } else {
+            text.setTextColor(Color.parseColor(RED));
         }
     }
 
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1000,12 +1000,9 @@ public class Pokefly extends Service {
         resultsDefense.setText(String.valueOf(ivScanResult.iVCombinations.get(0).def));
         resultsHP.setText(String.valueOf(ivScanResult.iVCombinations.get(0).sta));
 
-        GUIUtil.setTextColorbyPercentage(resultsAttack,
-                (int) Math.round(ivScanResult.iVCombinations.get(0).att * 100.0 / 15));
-        GUIUtil.setTextColorbyPercentage(resultsDefense,
-                (int) Math.round(ivScanResult.iVCombinations.get(0).def * 100.0 / 15));
-        GUIUtil.setTextColorbyPercentage(resultsHP,
-                (int) Math.round(ivScanResult.iVCombinations.get(0).sta * 100.0 / 15));
+        GUIUtil.setTextColorByIV(resultsAttack, ivScanResult.iVCombinations.get(0).att);
+        GUIUtil.setTextColorByIV(resultsDefense, ivScanResult.iVCombinations.get(0).def);
+        GUIUtil.setTextColorByIV(resultsHP,ivScanResult.iVCombinations.get(0).sta);
 
         llSingleMatch.setVisibility(View.VISIBLE);
         llMultipleIVMatches.setVisibility(View.GONE);
@@ -1106,9 +1103,9 @@ public class Pokefly extends Service {
             ave = ivScanResult.getAveragePercent();
             high = ivScanResult.getHighestIVCombination().percentPerfect;
         }
-        GUIUtil.setTextColorbyPercentage(resultsMinPercentage, low);
-        GUIUtil.setTextColorbyPercentage(resultsAvePercentage, ave);
-        GUIUtil.setTextColorbyPercentage(resultsMaxPercentage, high);
+        GUIUtil.setTextColorByPercentage(resultsMinPercentage, low);
+        GUIUtil.setTextColorByPercentage(resultsAvePercentage, ave);
+        GUIUtil.setTextColorByPercentage(resultsMaxPercentage, high);
 
 
         if (ivScanResult.iVCombinations.size() > 0) {
--- a/app/src/main/java/com/kamron/pogoiv/widgets/IVResultsAdapter.java
+++ b/app/src/main/java/com/kamron/pogoiv/widgets/IVResultsAdapter.java
@@ -14,6 +14,9 @@ import com.kamron.pogoiv.R;
 import com.kamron.pogoiv.logic.IVCombination;
 import com.kamron.pogoiv.logic.IVScanResult;
 
+import java.text.DecimalFormat;
+import java.text.NumberFormat;
+
 /**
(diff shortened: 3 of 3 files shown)


Blaisorblade:
> Use text color based on Appraisal system.

That could be fine with me, I'd let the others decide.

> Show IV percent to 1 decimal point in possibilities.

We're talking about the individual possibilities right? That was discussed at length and rejected.

Also decimals add no information: A single IV percentage can be 84%, 87% or 89%, nothing in the middle that needs decimals; a single IV point gives a difference of 2.222%. There's no point showing those are really 84.444...% or 86.666...% or 88.888...%.

leppie:
Ok, I'll bin the decimal point. Should I remove the % sign too?

Blaisorblade:
> Should I remove the % sign too?
I guess yes. That'd leave only "Use text color based on Appraisal system".

For the rest, does @nahojjjen or @oskomonzon have a preference?
And if possible, please have one commit with descriptive title, like "Use text color based on Appraisal system".

leppie:
All done. Another idea might be to multi level check boxes for appraisal hints too.

Eg: 1 click makes check, subsequent cycles the matching colors, then clear again.

DJCrapsody:
Would be good if the javadoc comments are updated to match the current codes.

leppie:
Will do

Blaisorblade:
We're nearing a release so I'd like to minimize changes and potential regression—it's not like we have an extensive test suite.

> Another idea might be to multi level check boxes for appraisal hints too.
> 
> Eg: 1 click makes check, subsequent cycles the matching colors, then clear again.

For now probably not—but feel free to open an issue with a clearer description.

> All done.
>
> Edit: Arghhhh, number formatter did something stupid...

I'd guess just take away that entire part of the change—no reason to change `percentPerfect` to `double`.

leppie:
> I'd guess just take away that entire part of the change—no reason to change percentPerfect to double.

Leaving it as is, from the diff it looks cleaner :) And who knows, maybe it will be needed later if the game devs change stuff again.

Blaisorblade:
> Leaving it as is, from the diff it looks cleaner :)

I'm looking for a minimal diff, and I'm rejecting changes to the core logic next to a release.

> And who knows, maybe it will be needed later if the game devs change stuff again.

That's a bad argument. I find those changes a regression.

leppie:
OK, reverted back to int, also adjusted to midpoints to work correctly with ints.

Blaisorblade:
If possible, I'd like @oskomonzon's opinion on the design (basically on the color choice), since he's our master designer. But otherwise LGTM.

leppie:
The blueish was picked at random :)

nahojjjen:
I'd like to avoid blue coloring, as it does not fall into the intuitive red-orange-yellow-green spectrum that people are used to for seeing how perfect something is.

Blaisorblade:
Maybe blue -> orange would be an improvement?

nahojjjen:
I'd like clarification of what the change actually is in this PR. The ivs are already color coded, so what this pr does is change the color coding - why, and to what?

oskomonzon:
I agree with @nahojjjen on this one. I'm not against the overall idea but we need to avoid introducing changes that lead to user confusion, I feel the blue would fall in that category.
To clarify: I don't mind at all the change to 4 steps of distinction, but I'd say we stay in the green to red spectrum.

oskomonzon:
@nahojjjen as far as I understand what's propossed here... I think the objective is to match the color coding to the different phrases the team leader uses to distinct IV ranges.

Instinct, as an example, goes by: 
80% - 100% : Overall, your (Pokemon Name) looks like it can really battle with the best of them!
67% - 79% : Overall, your (Pokemon Name) is really strong!
51% - 66% : Overall, your (Pokemon Name) is pretty decent!
0% - 50% : Overall, your (Pokemon Name) has room for improvement as far as battling goes.

So he wants the color code in four steps to match that.

Blaisorblade:
See the table on [GamePress](http://pokemongo.gamepress.gg/pokemon-appraisal) for the correct ranges (there are slight differences)—the PR itself is fine there.

Blaisorblade:
@oskomonzon Are you fine with any orange or do you want to pick colors that match with some care?

leppie:
Would the color ordering be the like this? 

image

oskomonzon:
Thanks, @Blaisorblade, I updated my comment.
Maybe we can borrow colors from the Material Design color palette (https://material.google.com/style/color.html#color-color-palette, all praise Duarte).

I'm not on the computer right now, but I guess something like this can work:
#F44336 - Red
#FF9800 - Orange
#FFEB3B - Yellow
#4CAF50 - Green

oskomonzon:
@leppie I'd avoid too vibrant/washed colors to prevent hurting readability.

leppie:
@oskomonzon Yes, of course, was just interested in the ordering.

oskomonzon:
@leppie, ordering makes sense :+1: 
What do you think about those?, same for @nahojjjen. Keep in mind I'm just guessing, I will not be able to see them in action uptil later.

leppie:
I will make a build and post a screenshot

leppie:
The yellow is quite washed out... :(

I will wait for better suggestions :)

image

nahojjjen:
I find it confusing that the % and the ivs do not use the same color scale.

leppie:
A bit off-topic, but can one add sorting on that listview, so clicking on any column/column header will toggle it?

leppie:
@nahojjjen: Maybe that can be an option, so user has choice.

Blaisorblade:
Re confusion: that's true, but the scales come externally. The colors could help compare the results 
with appraisals. OTOH, we might just add extra appraisal checkboxes.

> A bit off-topic, but can one add sorting on that listview, so clicking on any column/column header will toggle it?

That is probably a good idea. We've wanted to sort by attack first and somehow nobody did a PR. What you're proposing is probably possible but will certainly have to wait a while longer (we need to close 3.1 and refactor things).

oskomonzon:
I share @nahojjjen's concern about the different scaling in the color codes.
To be a little more clear, we are creating a scenario where the color value means something different depending on the data we are displaying...
For stats (Atk, def, sta) Green means 15 (The maximum value for that stat, 100%, so to speak)
On the other hand, for IV% Greens would mean "above 82%".

Blaisorblade:
Also: yes it's conceivable, so please file an issue so we don't forget about it. Same for your other idea (that I didn't even get).

And to be sure: thanks for your proposals!

> Maybe that can be an option, so user has choice.

We're trying to avoid having options for everything—it's more confusion for the user and more testing (currently manual) to get it right.

Blaisorblade:
Since we're not sure yet, scheduled for after the refactorings.

Remiscan:
I think red-orange-yellow-green is a bit confusing, because yellow makes these results look bad/average when they're actually good (13 or 14 isn't bad at all), and orange makes them look bad when they're average to quite good. Maybe something like red-yellow-lighter green-green?

oskomonzon:
@Remiscan that makes sense. I'd use orange instead of yellow in that case.

oskomonzon:
Overall, I like the idea, but I'm conflicted by the lack of consistency as explained earlier (https://github.com/farkam135/GoIV/pull/336#issuecomment-244146439).

leppie:
Went back to original colors, and changed the blue I added to a khaki green, which does look kinda nice.

image

Remiscan:
I think it looks better, but thak khaki green doesn't make me feel as if the result is "good" enough. I think I'd go with something a bit "greener", like #99CC00 although that may be too vivid.

I also find the lack of consistency a bit confusing, and I felt like making the colors closer to one another by having two shades of green would make the difference less "shocking". But seeing the last screenshot, it still looks a bit confusing.

DJCrapsody:
IMO, too much colors. My suggestion would be the entire row will use the same color based on % IV. Then bold the highest IV(s) for the row. IV don't make a very significant difference compared to movesets.

amcameron:
I disagree, @DJCrapsody. If IVs are unimportant then uninstall GoIV and just use the built-in appraisal system. Movesets are more important than IVs, but how do you distinguish between Pokémon at different levels with the same movesets?

Bolding the highest entries in each row would be redundant because of the optional Appraisal input step before you activate the "Check IVs" button.

Personally, I like the colours best in the very first comment on this pull request, although I would switch blue and green. Bad/red, decent/yellow, good/green, excellent/blue. This way, the "strong" range still looks good, and the absolute best are special.

The lack of consistency mentioned by @oskomonzon does not bother me. To me, there is no lack of consistency: one is used based on the percentage range of the IV sum, and the other is for discrete IVs between 0–15, and in both cases the colour thresholds change with the appraisal feedback thresholds. To me, colour-coding only based on value is pointless visual flair. The colours are useful if they allow me to look up entries based on the appraisal feedback; otherwise, they simply distract and there's no point having a three- or four-element colourmap and you might as well use 46 distinct colours, one for each possible IV%.

Blaisorblade:
> The colours are useful if they allow me to look up entries based on the appraisal feedback

That again. I also refine results by hand and think we should go the full way (everybody kind of agrees). So the highlighting should IMHO be designed assuming we do that.

https://github.com/farkam135/GoIV/issues/341#issuecomment-244556217

amcameron:
Another idea, instead of using four colours or a different shade of green, is to use the same shade of green for IVs from 13–15, and for 15 to also draw a shadow with TextView.setShadowLayer(). That would avoid subjective opinions about colour while emphasising the different appraisal tiers.

Remiscan:
I tried a few colour combinations so I thought I'd show everyone: 
- Light Green / Dark Green : http://i.imgur.com/b1rgsOE.png (I like this one)
- Green / Dark Green : http://i.imgur.com/2MrQO4k.png
- Green / Turquoise : http://i.imgur.com/XSG56kJ.png (and that one too)
- Green / Blue : http://i.imgur.com/HUvqkHl.png
- Dark Green / Blue : http://i.imgur.com/qT99AJZ.png

I think shadows may look weird but I will try and see.

leppie:
Maybe the colors for IV and % should be different (one darker than other?), but follow the same 'progression' .

amcameron:
IMHO of those ones I like Green/Blue or Dark Green/Blue the best.

leppie:
I like these ones the most:
* Light Green / Dark Green 
* Dark Green / Blue

crd:
As a neutral third party I agree:

* coloring should be consistent with appraisal values
* I'm not too worried about misleading colors -- users of GoIV are sharp enough to understand what's happening
* I would optimize for visibility, not red -> green coloring. It's obvious that a 15 is better than a 7
* inconsistent coloring across appraisals versus overall percentage doesn't bother me. Take that one up with nuantic

Looking forward to this PR (or a variant) being accepted!

Blaisorblade:
In the end I've merged this PR with the existing colors, mostly to make progress and because the resulting code is also a bit cleaner.

In the end consistency among different things isn't to be expected. Maybe it'd be better to use disjoint sets of colors for the two scales to reduce ambiguity, but that requires enough colors that are different enough.

I've opened #388 to remember to track the color choice—I'll leave that to somebody else.

PR #338, opened 2016-09-01
Added checkstyle rules and configured checkstyle plugin

harkin:
Things this does that might be controversial:

120 character line limit (do you care about line limits? I like 120 but will happily change it)

No Hungarian notation. It's awful.

Some rules about javadoc formatting.

Naming of abbreviations: `theOCR` is not allowed, `theOcr` is allowed

There are tonnes of violations currently, I _think_ CI can be configured to allow a certain number so at least more won't creep in

To run: `./gradlew checkstyle`

### DIFF
--- a/checkstyle/checkstyle-rules.xml
+++ b/checkstyle/checkstyle-rules.xml
@@ -1,0 +1,189 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+    "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+    "http://www.puppycrawl.com/dtds/configuration_1_3.dtd">
+
+<!--
+    Checkstyle configuration that checks the Google coding conventions from Google Java Style
+    that can be found at https://google.github.io/styleguide/javaguide.html.
+
+    Checkstyle is very configurable. Be sure to read the documentation at
+    http://checkstyle.sf.net (or in your downloaded distribution).
+
+    To completely disable a check, just comment it out or delete it from the file.
+
+    Authors: Max Vetrenko, Ruslan Diachenko, Roman Ivanov.
+ -->
+
+<module name = "Checker">
+    <property name="charset" value="UTF-8"/>
+
+    <property name="severity" value="warning"/>
+
+    <property name="fileExtensions" value="java, properties, xml"/>
+    <!-- Checks for whitespace                               -->
+    <!-- See http://checkstyle.sf.net/config_whitespace.html -->
+    <module name="FileTabCharacter">
+        <property name="eachLine" value="true"/>
+    </module>
+
+    <module name="TreeWalker">
+        <module name="SuppressWarningsHolder"/>
+        <module name="OuterTypeFilename"/>
+        <module name="IllegalTokenText">
+            <property name="tokens" value="STRING_LITERAL, CHAR_LITERAL"/>
+            <property name="format" value="\\u00(08|09|0(a|A)|0(c|C)|0(d|D)|22|27|5(C|c))|\\(0(10|11|12|14|15|42|47)|134)"/>
+            <property name="message" value="Avoid using corresponding octal or Unicode escape."/>
+        </module>
+        <module name="AvoidEscapedUnicodeCharacters">
+            <property name="allowEscapesForControlCharacters" value="true"/>
+            <property name="allowByTailComment" value="true"/>
+            <property name="allowNonPrintableEscapes" value="true"/>
+        </module>
+        <module name="LineLength">
+            <property name="max" value="120"/>
+            <property name="ignorePattern" value="^package.*|^import.*|a href|href|http://|https://|ftp://"/>
+        </module>
+        <module name="AvoidStarImport"/>
+        <module name="RedundantImport"/>
+        <module name="UnusedImports"/>
+        <module name="OneTopLevelClass"/>
+        <module name="NoLineWrap"/>
+        <module name="EmptyBlock">
+            <property name="option" value="TEXT"/>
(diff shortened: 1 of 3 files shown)


nahojjjen:
We had a discussion about char limits in discord. I argued for line limits with 150 chars, and others argued with line limits for 80 chars. We came to the conclusion that we should use Android studio default, which I believe it somewhere in between there. I'm ok with 120 chars.

harkin:
I'm used to 120 but could be swayed for longer. Anything less is a little annoying and you bump into it too often I think

Blaisorblade:
120 is an improvement for me over what we have, @sarav will be happy too.

Blaisorblade:
How do you invoke this? Can it be done on the command line?

harkin:
@Blaisorblade yes! `./gradlew checkstyle`

sarav:
Is this integrated into Android Studio somehow? I like using vim or Android Studio. Don't like invoking scripts to format the code after I'm done coding.

Blaisorblade:
The `GoIVCodeStyle.xml` *can* instead be used by Android Studio for reformatting, `DEVELOPMENT.md` has the instructions.

checkstyle does *not* reformat code, it just complains. And too much. (You can invoke it via Android Studio through Gradle, but it's not integrated, you can't even click errors). So in practice we're not using it yet.

We could 
- maybe reduce the rules
- fix violations of the the remaining ones (we already started, I broke longer lines)
- and have Travis run checkstyle and fail builds for violations
but we'd have some fighting to do for that.

I could volunteer to fix stuff we care about.

harkin:
There's a plugin you can get for AS if you want to run the checks there, it's `Checkstyle-IDEA`

This weekend I might (hopefully) have some time to fix remaining violations. If there are rules you disagree with I'd absolutely get rid of them but I wouldn't get rid of a rule because existing code doesn't comply.

For CI I should be able to set it up to have an acceptable number of warnings/violations. Setting that to the current number and turning it on would stop more being introduced at least.

checkstyle/checkstyle-rules.xml
```diff
@@ -0,0 +1,189 @@
...
+                     value="Package name ''{0}'' must match pattern ''{1}''."/>
+        </module>
+        <module name="TypeName">
+            <message key="name.invalidPattern"
+                     value="Type name ''{0}'' must match pattern ''{1}''."/>
+        </module>
+        <module name="MemberName">
+            <property name="format" value="^([a-z][a-z0-9]|[a-ln-rt-z][_A-Z0-9])[_a-zA-Z0-9]*$"/>
+            <message key="name.invalidPattern"
+                     value="Member name ''{0}'' must match pattern ''{1}''."/>
```
Blaisorblade: This is the rule about Hungarian notation? I'm getting violations for `mContext` and similar.

harkin: yes

Blaisorblade:
Ah, plugin found once I clicked "Browse repositories" in the plugins dialog (which I found rather confusing BTW).

## Choice of rules and violations

Things we should IMHO care about, in this order:

1. working code review on diffs (because code review is effective)
2. a working workflow where menial details are automated.
3. applying useful rules. Missing braces for control statements can cause bugs... if your indentation is not consistent (but we get that far).
4. apply rules that give consistency. I care about this, but further rules than we have are IMHO less important in case of a conflict.

Workflows relying on ignoring false positives fail these rules. False positives are already a problem with bug finders.

> For CI I should be able to set it up to have an acceptable number of warnings/violations.

CheckStyle doesn't work well that way.

With the above criteria, ignoring violations in CI seems a no-go given current tooling. Specifically, I don't see a way to run CheckStyle only on actual lines you changed. If you run it on files you changed, you need to fix violations in new code, but leave alone violations in existing code. Or you'll get unreviewable commits fixing unrelated style issues together with a hidden one-line fix that in fact introduces some other regression.

That can be worse than getting rid of a rule because existing code doesn't comply. If a rule matters a lot we should fix it. If it only matters a bit... I'd rather leave it out. 

But if nobody cares enough to fix the violations, then maybe the actual community doesn't care about those rules. Matter of fact, AFAICS nobody is using this style right now.

I suspect I'd rather remove all checks triggering violations, force a "no error" policy, and then accept new rules together with patches fixing that up.

## Rule choice

1. Why not `theOCR`? We seem to use that consistently, especially for IV, HP, CP, and even in the app name. I doubt whether `Iv` is more readable than `IV`. I get this allows segmenting words in camelcase identifiers a bit more easily... but we use a few clear acronyms, so that seems not a problem.

2. I'm against any rules where autoformat introduces violations, unless this can be fixed by reconfiguring autoformat.

3. If we require braces for control statements, we should configure the reformatter for automatically fixing violations. Same for other rules where this is applicable.

harkin:
> For CI I should be able to set it up to have an acceptable number of warnings/violations.
CheckStyle doesn't work well that way.

It helps get it off the ground for now. As legacy violations are fixed the number can be lowered. 

> the actual community doesn't care  

They aren't enforced, chances are many people don't know they exist and won't until builds start failing

> Why not theOCR?

I used rules I'm familiar with from another project :) We  broadly follow the google java style guide, specifically [this](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case)

> unless this can be fixed by reconfiguring autoformat.

Autoformat can be configured to do anything but sticking close to the defaults is probably a good idea

I'm happy to actually fix all these violations but I haven't had any free time lately and I'm not sure when I will either. I'll try to get it done this evening.

Blaisorblade:
I've fixed/disabled all violations and send #386 to force compliance, let's see what the others think.

PR #342, opened 2016-09-01
Quickfix manual screenshot mode

### LINKED ISSUE #315: Battery-saver mode is only force-enabled on 4.4 when entering settings
To reproduce:
- start GoIV on Android 4.4
- press start
- observe a crash

To workaround:
- enter the settings menu (without having to touch anything), return, and start. Tis must be done only once IIUC, the issue stays fixed.

Solution approach:
- force battery-saver mode on when starting the MainActivity, not in settings.

Blaisorblade:
Fix #315. A bit hacky, but I'm not sure how to change settings from
elsewhere, and where, and we're in a hurry.

And force screenshot mode also for API level 20.

Nobody should be running GoIV on Android Wear IMHO, but there are
unexpected crashes on Android 5 with NoClassDefFoundError for the
MediaProjectionManager API, so let's rule that out more clearly.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
+++ b/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
@@ -40,7 +40,7 @@ public class GoIVSettings {
 
     public boolean isManualScreenshotModeEnabled() {
         //XXX unify with code in SettingsActivity.java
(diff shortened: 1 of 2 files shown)


nahojjjen:
I do not have a kitkat device to test with, but the changes seem to make sense

PR #351, opened 2016-09-03
Update DEVELOPMENT.md

DJCrapsody:
Corrected location path for all 3 OSes
{ANDROID_STUDIO_FOLDER} is used as different versions have different folders

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -8,12 +8,14 @@
 
 ## Project Set-up for GoIV (Required)
 * [Fork GoIV Repo] (https://help.github.com/articles/fork-a-repo/)
-* Copy GoIVCodeStyle.xml into the codestyles folder under the Android Studio config/codestyles directory *(create if it does not exist)*
+* Copy GoIVCodeStyle.xml into the codestyles folder under the Android Studio preferences folder *(create codestyles folder if it does not exist)*
  *  Windows: `\%USERPROFILE%\.{ANDROID_STUDIO_FOLDER}\config\codestyles\`
  *  OS X: `~/Library/Preferences/{ANDROID_STUDIO_FOLDER}/codestyles/`
  *  Linux: `./.{ANDROID_STUDIO_FOLDER}/config/codestyles/`
 * Open the Project in Android Studio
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -6,10 +6,10 @@
...
 * Set up [Android Development Environment] (https://spring.io/guides/gs/android/)
 * [Fork this repo] (https://help.github.com/articles/fork-a-repo/)
-* Copy GoIVCodeStyle.xml into the codestyles folder under the IntelliJ home directory (create if it does not exist)
- *  Windows: `\.\config\codestyles\`
- *  OS X: `~/Library/Preferences//codestyles/`
- *  Linux: `~/./config/codestyles/`
+* Copy GoIVCodeStyle.xml into the codestyles folder under the Android Studio config/codestyles directory *(create if it does not exist)*
+ *  Windows: `\%USERPROFILE%\.{ANDROID_STUDIO_FOLDER}\config\codestyles\`
+ *  OS X: `~/Library/Preferences/{ANDROID_STUDIO_FOLDER}/codestyles/`
+ *  Linux: `./{ANDROID_STUDIO_FOLDER}/config/codestyles/`
```
Blaisorblade: If the Windows one has a dot, I'd expect the Linux to have one too.
`{ANDROID_STUDIO_FOLDER}` should be something like `AndroidStudio2.1` right? Better say so...

DJCrapsody: Oops yeah, linux has a dot too, probably missed that out.
Like I've explained in the commit message, it could be AndroidStudioBeta, AndroidStudio1.5, AndroidStudio2.2Beta, AndroidStudio3.6, etc. As long as you update AndroidStudio, it'll be different.

Blaisorblade: I read the commit message. Just saying that docs should be clear on their own, so we should explain what `{ANDROID_STUDIO_FOLDER}` means and give at least an example.
(And right now everybody should be using 2.1, using older versions probably means looking for trouble).

DJCrapsody:
Do not merge yet. Testing Enable Annotation Processing... Somehow I'm getting errors for getter/setter methods that is only created by lombok, builds fine tho.

DJCrapsody:
Ready to merge.

PR #352, opened 2016-09-03
First bunch of refactorings

Blaisorblade:
This is me starting to read the code and fix nonsense.
- [x] TODO: test at least once.

Non-obvious goals:
- stuff that mutates state can be hard to follow, so try to *not* hide where this happens. A constructor of a data object like `IVScanResult` changing state is *bad*.
- reduce "action at a distance". If something is set in a place and used somewhere else entirely, it's hard to follow what's happening.
- if a function computes some result, it shouldn't return it by storing it in fields that are used later—if you do that, somebody can break the code by reordering calls that seem unrelated.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/DevMethods.java
+++ b/app/src/main/java/com/kamron/pogoiv/DevMethods.java
@@ -26,7 +26,7 @@ public class DevMethods {
         //to create the evolutions
 
         int evolutionCost = -99999;
-        for (Pokemon poke : pokeCalculator.pokedex) {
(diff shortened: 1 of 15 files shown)


nahojjjen:
Ive read through the changes and like them, and I've tested the changes on my phone, scanning ~10 pokemon, and it doesn't seem to break anything obvious.

I think it's mergeable.

Blaisorblade:
Thanks! The commits I added are just reformatting, and I tested it too. Merging.

PR #355, opened 2016-09-03
Change default evolution stats to current Pokemon level

### LINKED ISSUE #332: Have default IV screen show evolved stats when possible at the current level
I guess I'm probably in the minority here, but I kind of liked the way one of the older versions worked where we would see the CP if were to evolve at the current pokemon level.

I was thinking of making this an option, but trying to think of the best way this would be implemented.

For mons that do not get evolved (or are at their highest evolution already), I think it makes more sense to  default to show what the CP would be at the max level.

However, for mons that can be evolved, I would prefer to show the CP of the mon at the current level for the next step in the evolution chain.

Do other people have opinions on how to make this nicely configurable?

DJCrapsody:
Defaults to current Pokemon level (seekbar starts on the left)
<br />Added additional layer underneath seekbar to show post-trainer level
<br />Add MultiSlider library to support inverted seekbar
<br />Extra: Fix seekbar height in dialog_input.xml
<br />
<br />Fix #332 
<br />
<br />Note to devs: We will check on the requests to have the original behaviour.


oskomonzon:
I love it, great job. Hopefully we're gonna get positive feedback on this one :+1:

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -807,7 +808,12 @@ private void populateResultsBox(IVScanResult ivScanResult) {
      *
      */
     private void adjustSeekbarForPokemon() {
-        expandedLevelSeekbar.setProgress(levelToSeekbarProgress(Data.trainerLevelToMaxPokeLevel(trainerLevel)));
+        expandedLevelSeekbarBackground.getThumb(0).setThumb(getResources().getDrawable(R.drawable
+                .orange_scrubber_control_disabled_holo));
+        expandedLevelSeekbarBackground.getThumb(0).setValue(levelToSeekbarProgress(trainerLevel + 1.5f));
```
Blaisorblade: That's `Data.trainerLevelToMaxPokeLevel(trainerLevel)` not  `trainerLevel + 1.5f` (they're not always the same, the former has a few bugfixes).

Blaisorblade:
Can be merged once the feedback is addressed.

DJCrapsody:
Oops. I forgot to fix it back. Will do.

DJCrapsody:
Ready to merge

PR #368, opened 2016-09-04
Add more appraisal

Blaisorblade:
Don't merge yet. For ongoing review.


app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -200,6 +237,33 @@ public void onClick(View v) {
...
+
+        spinner.setOnItemSelectedListener(new AdapterView.OnItemSelectedListener() {
+            @Override
+            public void onItemSelected(AdapterView<?> parentView, View selectedItemView, int position, long id) {
+                settings.setPlayerTeam(position);
+            }
+
+            @Override
+            public void onNothingSelected(AdapterView<?> parentView) {
+            }
```
Blaisorblade: Please add a comment saying why this can't be selected.

app/src/main/java/com/kamron/pogoiv/logic/IVCombination.java
```diff
@@ -38,6 +38,21 @@ public boolean equals(Object o) {
...
+    public int getHighestStat(){
+        int high = 0;
+
+        if (att > high) high = att;
+        if (def > high) high = def;
+        if (sta > high) high = sta;
+
+        return high;
+    }
+    /**
```
Blaisorblade: Below in `getHighestStatSignature` we have already `Collections.max(Arrays.asList(att, def, sta))`. Let's not add another implementation of that.

Blaisorblade:
@nahojjjen when you have time can you address my comments or let me know if you have objections?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
nahojjjen: onNothingSelected is a method which is called if the spinner is removed by any action other than the user picking an item from the list. The interface for the adapter listener requires onNothingSelected to be implemented, but we don't want anything to happen if the user dismisses the spinner or if it disappears, so I left the method empty. I can add a comment in the method saying "//We dont want anything to happen when user dismiss spinner, but interface requires implementation" if you think it'd help with clarity.

app/src/main/java/com/kamron/pogoiv/logic/IVCombination.java
nahojjjen: Ill fix that

app/src/main/java/com/kamron/pogoiv/logic/IVCombination.java
```diff
@@ -39,13 +39,7 @@ public boolean equals(Object o) {
...
      */
     public int getHighestStat(){
-        int high = 0;
-
-        if (att > high) high = att;
-        if (def > high) high = def;
-        if (sta > high) high = sta;
-
-        return high;
+        return Collections.max(Arrays.asList(att, def, sta));
```
Blaisorblade: 👍

leppie:
I am upset that the team leist is not sorted alphabetically, but only because I am Instinct :D

PR #369, opened 2016-09-05
Split out PokemonNameCorrector from Pokefly

Blaisorblade:
PokemonNameCorrector is large enough for its own class, also because it has its own state. This would also help with the refactoring.

I wasn't sure what interface I should have, so I've added a few options. One even uses Lombok. What do you think?

Untested yet.


oskomonzon:
LGTM

PR #370, opened 2016-09-05
Remove okio package dependency

DJCrapsody:
Compiling okio package did not work and caused release builds to have proguard error

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -136,7 +136,6 @@ dependencies {
     compile 'com.jakewharton.timber:timber:4.3.0'
     compile 'com.rmtheis:tess-two:6.0.4'
     compile 'com.squareup.okhttp3:okhttp:3.4.1'
-    compile 'com.squareup.okio:okio:1.10.0'
     compile 'io.apptik.widget:multislider:1.2'
     provided 'org.projectlombok:lombok:1.16.10'
     apt 'com.jakewharton:butterknife-compiler:8.4.0'


Blaisorblade:
What's OKIO? Are we even using it? Do we still need the Okio Proguard directives?

harkin:
Why do you mean by 'compiling it didn't work'? A better solution might be to fix the build errors than to remove it.

@Blaisorblade 
> What's OKIO?
[Okio](https://github.com/square/okio)

> Are we even using it? 
It's a dependency of Okhttp which was introduced for use with the the app updater. Latest Okhttp depends on v1.9.0 of Okio, there've been no API changes so it's safe to build against 1.10.0 instead.

> Do we still need the Okio Proguard directives?
Yes

Blaisorblade:
> What do you mean by 'compiling it didn't work'? A better solution might be to fix the build errors rather than removing it.

@DJCrapsody tried removing `dontwarn` but that broke the release build (and Travis didn't tell us of course, as IIRC you predicted would happen).

harkin:
Ah, yeah that'd do it. 

Okio is a library that can be used in general java apps; it's not Android specific. It makes reference to some classes that aren't in the Android platform (`java.nio.*`). If those classes aren't present it doesn't use them but Proguard isn't smart enough to realise that and warns about 'missing' classes. The `-dontwarn` will be necessary for Okhttp even if this Okio dependency is removed.

My vote is to keep this in to be more explicit about dependencies and also to use the latest version of the library.

Blaisorblade:
Thanks for explaining. I think the reasoning should be in a comment, because people (me included) will try to remove those (unobvious) bits. Actually, two comments—one for the dependency and one for the `dontwarn`.

DJCrapsody:
Previous builds before changes made to proguard (removing --dontwarn) and gradle (adding okio dependency) gives the same error as when I made the change. This is just to revert the changes made overall.

Blaisorblade:
Ah, so this PR reverts this line, which was added for an invalid reason:
https://github.com/farkam135/GoIV/commit/dc9db6d917f54d312b9604eeeb54a6cee10d27bf#commitcomment-18903911

But if @harkin still prefers this in, I'm inclined to leave it.

DJCrapsody:
I'm against keeping it. I found out that okhttp3 package will also compile the dependency as it requires it and compiling it on our end makes it duplicate. https://github.com/square/okhttp/issues/2297

Proguard just hates okhttp, the only solution I've found so far is to dontwarn.

Blaisorblade:
@DJCrapsody I'm sorry but @harkin has explained the issue satisfactorily, and his argument makes sense to me. It's not that "Proguard just hates okhttp".

I understand nothing is actually duplicated in the end for us. I'm not sure how the issue you link is relevant—there somebody gets duplication after adding a JAR manually and via Gradle. [Quoting](https://github.com/square/okhttp/issues/2297#issuecomment-176809771):

> Gradle will automatically pull down the Okio dependency. This is why adding the jar manually causes duplicate classes to be present.

Closing.

harkin:
If you want to see how the dependency tree is evaluated you can run `./gradlew dependencies`. When all dependencies are grabbed through the buildscript it will by default remove duplicates and use the latest versions of each one.

> Proguard just hates okhttp, the only solution I've found so far is to dontwarn.

@DJCrapsody The solution *is* to use `-dontwarn`. The warnings Proguard throws are ignorable for this library. Many many libraries require specific Proguard rules to work properly; Okhttp is actually one of the better ones there in that the `-dontwarn` is the only thing necessary.

DJCrapsody:
It's the same as compiling it with gradle. The dependency is in okhttp3 pom.xml, which will be compiled into okhttp3, not accessible to goiv. Unless we need it, the dependency in gradle is not required.

Of course it isn't wrong to leave it in, we might use it in the future. Just not required yet. Not worth too much discussion now, maybe for the next round of cleanup. Gradle is still fresh to most of us, we're all still learning and working on it helps. Cheers!

Blaisorblade:
> Unless we need it, the dependency in gradle is not required.

That was pretty clear.

> if okhttp3 lib isn't dependent the latest okio lib, our more recently updated okio lib dependency will replace it?

@harkin has been saying that for a couple times. `./gradlew dependencies` shows nothing, but `./gradlew :app:dependencies` the following relevant lines:

```
+--- com.squareup.okhttp3:okhttp:3.4.1
|    \--- com.squareup.okio:okio:1.9.0 -> 1.10.0
+--- com.squareup.okio:okio:1.10.0
```

That's shown for offline builds too... I'm thinking of disabling the update code for those builds.

DJCrapsody:
Fair enough! More we know.

Issue #268, opened 2016-08-28
Add Notification Action Buttons


voltaek:
Adding an embedded action button within the GoIV notification to stop and start screen capture would be a nice alternative to having to switch back and forth between the PoGo and GoIV apps. This way when a user taps Stop from the notification, the notification could stay in their notification list while they are doing other things in PoGo, then when they're ready to scan things again they can just tap Start on the notification button. Alternatively, if they're all done, they could just swipe the notification away after tapping Stop (I think you'd want to continue preventing swiping it away while still running, as it do currently). To be clear, this would be just one button that they'd tap to toggle screen capture, and the text would just change from Start to Stop.

This is an alternative to #263, the always-displayed floating button suggestion, where I suggested that screen scanning only runs for a moment when the button is tapped. If that idea is not done, this seems like the next best way of easily limiting the time that your screen is being continuously scanned.

Outside of anything relating to toggling scanning, I'm sure there are other things users could use action buttons for.

Paxtez:
Would this be a duplicate of #30?

Blaisorblade:
@Paxtez yes but this issue has a better description.

Personally, I'd like we'd need keys to:
- start Pokefly
- stop Pokefly
- I'd love increase trainer level by 1 (every other change can be done via the GUI, that's the only one that is a bit annoying)
- something to close the app when Pokefly is stopped and the notification is dismissed.

Technically, notification actions require intents mostly, so this relates to #278 and the intent flow a bit. Now that other stuff was cleaned up, it should be at least possible to try.

rhari991:
I'd like to work on this if nobody is currently doing so.

Blaisorblade:
Didn't really start—assigning to @rhari991.

BTW: the one thing I did figure is that we might need to switch from Notification to [NotificationCompat](https://developer.android.com/reference/android/support/v4/app/NotificationCompat.Builder.html#addAction(android.support.v4.app.NotificationCompat.Action)) if we need APIs like [addAction(Action)](https://developer.android.com/reference/android/app/Notification.Builder.html#addAction(android.app.Notification.Action), since that requires API level 20 but we support 19.

rhari991:
Thanks, will add that in.

Eccenux:
> something to close the app when Pokefly is stopped and the notification is dismissed.

You can do that from recent apps. Closing GoIV from there is stopping the service and ends GoIV.

Blaisorblade:
[...]

PR #371, opened 2016-09-05
Notification with actions

thearaks:
Hi all!
Here I submit a Notification Actions implementation.
This adds the "Open app" and "Stop" actions and improves the overall notification style and attributes on latest Android versions.

There's also a small change to MainActivity to ensure that its GUI reflects the service state (started/stopped).

Removed unused string (notification text).

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -2,6 +2,7 @@ package com.kamron.pogoiv;
 
 import android.Manifest;
 import android.annotation.TargetApi;
+import android.app.ActivityManager;
 import android.app.DownloadManager;
 import android.content.BroadcastReceiver;
 import android.content.Context;
@@ -84,8 +85,9 @@ public class MainActivity extends AppCompatActivity {
 
     private boolean readyForNewScreenshot = true;
 
-    private boolean pokeFlyRunning = false;
     private int trainerLevel;
+
+    private Button launchButton;
 
     private final Point arcInit = new Point();
     private int arcRadius;
@@ -140,8 +142,8 @@ public class MainActivity extends AppCompatActivity {
         npTrainerLevel.setWrapSelectorWheel(false);
         npTrainerLevel.setValue(trainerLevel);
 
-        Button launch = (Button) findViewById(R.id.start);
-        launch.setOnClickListener(new View.OnClickListener() {
+        launchButton = (Button) findViewById(R.id.start);
+        launchButton.setOnClickListener(new View.OnClickListener() {
 
             @Override
             public void onClick(View v) {
@@ -181,13 +183,12 @@ public class MainActivity extends AppCompatActivity {
                         screenShotScanner.stopWatching();
                         screenShotScanner = null;
                     }
-                    pokeFlyRunning = false;
                     ((Button) v).setText(getString(R.string.main_start));
                 }
             }
         });
 
-        checkPermissions(launch);
+        checkPermissions(launchButton);
 
 
         displayMetrics = this.getResources().getDisplayMetrics();
@@ -308,7 +309,14 @@ public class MainActivity extends AppCompatActivity {
     @Override
     protected void onResume() {
         super.onResume();
+
         settings = GoIVSettings.getInstance(MainActivity.this);
+
+        if (isPokeFlyRunning()) {
+            launchButton.setText(R.string.main_stop);
+        } else {
+            launchButton.setText(R.string.main_start);
+        }
     }
 
     private int getStatusBarHeight() {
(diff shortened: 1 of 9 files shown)


DJCrapsody:
Some weird behaviours I have encountered and steps to reproduce them:
* IV button goes missing sometimes
 1. Start GoIV
 2. Pokemon Info (IV button appears)
 3. Stop GoIV from Notification
 4. Start GoIV
 5. Pokemon Info (no IV button)
 6. Stop GoIV from Notification
 7. Start GoIV
 8. Pokemon Info (IV button appears)

* Stop button does not change back to Start
 1. Open GoIV using Recent Apps or from Notification
 2. Stop GoIV from Notification (Button still stays as 'Stop' with 'Stop button onClick' event (click and it turns to Start)
 3. Leave GoIV (go to another app from Recent Apps or go to Home Screen)
 4. Return to GoIV (Button changes to 'Start' with 'Start button onClick' event (click and it starts Pokemon GO)

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -308,7 +309,14 @@ public void onClick(DialogInterface dialog, int which) {
...
     protected void onResume() {
         super.onResume();
+
         settings = GoIVSettings.getInstance(MainActivity.this);
+
+        if (isPokeFlyRunning()) {
+            launchButton.setText(R.string.main_stop);
+        } else {
+            launchButton.setText(R.string.main_start);
+        }
```
Blaisorblade: @DJCrapsody's testing suggests calling `launchButton.setText` here is not enough, it should also be called when starting/stopping the service.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -84,9 +85,10 @@
 
     private boolean readyForNewScreenshot = true;
 
-    private boolean pokeFlyRunning = false;
```
Blaisorblade: 👍

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -436,11 +443,12 @@ private void hideInfoLayoutArcPointer() {
...
     @Override
     public void onDestroy() {
-        if (!batterySaver) {
+        super.onDestroy();
+
+        if (!batterySaver && timer != null) {
             timer.cancel();
         }
 
-        super.onDestroy();
```
Blaisorblade: IIRC super.onDestroy had to be placed after `timer.cancel()`—and ideally it'd go at the end. I'd leave out unrelated changes.

app/src/main/res/values/strings.xml
```diff
@@ -16,7 +16,7 @@
     <string name="main_not_numeric">%1$s is not a number.</string>
     <string name="main_invalide_trainerlvl">Invalid Trainer Level!</string>
     <string name="notification_title">GoIV Running - Level %1$s</string>
-    <string name="notification_text">Tap to open</string>
+    <string name="notification_open_app">Open app</string>
```
Blaisorblade: Is there a behavior change that requires changing (and retranslating) the string?

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -456,21 +464,35 @@ public void onDestroy() {
      * makeNotification
      * Creates the GoIV notification
      */
-    private void makeNotification(Context context) {
-        Intent intent = new Intent(context, MainActivity.class);
+    private void makeNotification() {
+        Intent openAppIntent = new Intent(this, MainActivity.class);
+
+        PendingIntent openAppPendingIntent = PendingIntent.getActivity(this,
+                0, openAppIntent, PendingIntent.FLAG_UPDATE_CURRENT);
```
Blaisorblade: Why 0? Isn't this supposed to match `NOTIFICATION_REQ_CODE` below?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: Also, the existing logic for starting the service only works correctly if the label is correct (which seems... weird to me). @DJCrapsody in your first scenario, is it possible the label hadn't been updated? Because the logic to start the service from the button is unchanged.

Blaisorblade:
Thanks for working on this!

I have to give it a try (wrong computer); meanwhile I've taken a look at the code and added some initial comments. Feel free to correct me if I'm wrong, I am aware I'm not an Android expert.

thearaks:
Thanks @Blaisorblade and @DJCrapsody,
I'll check everything as soon as I'll have a spare hour at home!

I think I might use the LocalBroadcastManager to keep the GUI in sync as the service changes state.
Also, the start/stop controls logic should be brought out of the launch button OnClickListener. I'll study a cleaner solution.

About super.onDestroy() you're right: it should be at the end, I don't even know why I touched that line...

Thank you again!

Blaisorblade:
Thanks for your positive response! Looking forward to your new code (when you have time of course).

> I think I might use the LocalBroadcastManager to keep the GUI in sync as the service changes state.

`LocalBroadcastManager` can be a source of pain—which is why I thought we needed to refactor the app to allow for this feature. Maybe there's a way out though.

If you do that, you might want to make sure the code is robust to duplicated events (apparently broadcasts can get duplicated on some devices). Ideally, calling the handler the second time would be harmless (say because it repeats the same assignments)—technically, the handler should be idempotent. The code you added in `onResume` satisfies that condition.

FWIW: Currently we use a hack which causes trouble—an intent changes state so that a repeated intent is rejected. But those flags get out of sync and reject valid intents (causing all the "IV button disappeared" bugs). And fixing the "flags get out of sync" would involve ensuring hundreds of lines of code stop failing. As you'd expect, that's not easy 😉

Blaisorblade:
I'm closing this for now as the bug is tracked by #268 (and I'm trying to process PRs/issues), but I'm looking forward to an updated PR!

thearaks:
Hi @Blaisorblade,
I have a new version of this PR but I can't fix the Start / Stop from Notification bug found by @DJCrapsody. 
When Pokefly stops itself from the Notification Intent instead of MainActivity then it can't start up correctly anymore.
I'm investigating this strange behavior right now...

PR #375, opened 2016-09-05
Update French localization for more appraisal

Remiscan:
I translated the new appraisal-related strings, and tested it on my phone for each team. It looked good.


app/src/main/res/values-fr/strings.xml
```diff
@@ -70,8 +125,12 @@
     <string name="up_to_date">L\'appli est à jour</string>
     <string name="update_check_failed">Erreur pendant la recherche de mise à jour. Veuillez vérifier votre connexion internet</string>
     <string name="iv_perfect">% Parfait</string>
+    <string name="wrongPokemonNameInput">" n'a pas été trouvé."</string>
```
Blaisorblade: Does this mean we should be using a format string?

Blaisorblade: Merging anyway, that's something *we* should fix (and we can after this PR).

PR #377, opened 2016-09-06
Avoid depending on okio in offline builds

Blaisorblade:
This saves only ~330K. Still better than nothing.
```
$ ls -l ./app/build/outputs/apk/app-offline-debug.apk ./app/build/outputs/apk/app-online-debug.apk
-rw-r--r--+ 1 pgiarrusso staff 30057066 Sep  6 02:13 ./app/build/outputs/apk/app-offline-debug.apk
-rw-r--r--+ 1 pgiarrusso staff 30391942 Sep  6 02:14 ./app/build/outputs/apk/app-online-debug.apk
```

EDIT: don't merge until I set up a way to test this properly on Travis, otherwise this means looking for trouble.
EDIT2: offline builds are tested already. It'd be nice to test offline release builds, but that's a separate issue. So merging.


app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -106,6 +106,7 @@ public void onReceive(Context context, Intent intent) {
     private final BroadcastReceiver showUpdateDialog = new BroadcastReceiver() {
         @Override
         public void onReceive(Context context, Intent intent) {
+            assert BuildConfig.isInternetAvailable;
```
harkin: what's the value of this assertion?

I hadn't taken note of `BuildConfig.isInternetAvailable` before. In an ideal world that doesn't need to exist. Any code that does internety things would live in the `online` flavour and there would be a corresponding interface that does nothing in the `offline` flavour, like what you did with `AppUpdateUtil`

harkin:
👍

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: I wanted to ensure the *failing* stubs I added aren't called in an offline build—note they throw exceptions.

That's only true because these assertions hold. And checking that's true took some non-local reasoning and intent chasing, so I thought I'd better document it.

But indeed `MainActivity` could skip the check if `AppUpdateUtil.checkForUpdate` was a no-op in the offline version.

And it sounds like we shouldn't use this flag in `app/src/online/java/com/kamron/pogoiv/CrashlyticsWrapper.java` which is already online-only 😉 . 

I suspect I'd keep using the flag in `app/src/main/java/com/kamron/pogoiv/SettingsActivity.java` though.

Blaisorblade:
After your comments I'll rework this PR a bit.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
harkin: Do those assertions run in release builds? I've never used them but typically the jvm runs with them disabled by default, no idea at all for Android. 

Oh wow yeah that's really unnecessary for the `CrashlyticsWrapper`

It's not doing any real harm, just not necessary and maybe adding some code to the offline build that isn't ever used. There's something to be said for not having too many layers of abstraction so it might be fine in the settings activity.

Maybe it's worth a more robust internet check? This one is just "am I in the online flavour?". A better check might be to see if we have the internet permission (so it isn't relying on something in the build configuration) or maybe checking if you have internet _available_ when it's checked. The latter might require another permission, can't remember off the top of my head

Blaisorblade: > Maybe it's worth a more robust internet check?

If this matters we could have a separate feature request, but it seems relatively low-priority.

Blaisorblade: > Do those assertions run in release builds? I've never used them but typically the jvm runs with them disabled by default, no idea at all for Android.

Probably not, but they still have documentation value.

PR #379, opened 2016-09-06
Shorten/make the appraisals more consistent

imapokemon4:
I noticed that some of the appraisal phrases were shortened and others were word for word.
  
This PR makes them all <s>word for word</s> shortened. EDIT: updated by @Blaisorblade.


nahojjjen:
The reason was because when I created the appraisal input I made two spinners to pick the phrases which were on the same row, and they needed to fit the screen width. And Im on mystic, and I only tested for mystic & told other people to shorten their team phrases.

Ill test and see how these longer phrases work now

Eccenux:
Don't think it's a good idea. If you have full sentences then you have to read everything instead of just focusing on a few words. And besides even on full wide they barely fit on the screen.

Blaisorblade:
I'd like more consistency if possible, and trying to shorten them if possible sounds more functional. Using ellipsis to highlight what's shortened is also good.

imapokemon4:
I've committed my best effort at shortening everything while keeping everything as consistent as possible.

Blaisorblade:
Thanks! What do people think?

Before merging, you might want to merge or rebase with master.
I'd prefer to keep a literal ellipsis rather than `&#`, since the former is more readable, as long as it works.

imapokemon4:
@Blaisorblade Rebased with your instructions.

Blaisorblade:
Thanks, but there's still a conflict, it's not rebased on the latest master. Anyway, I think this is a good basis for getting started! What do the others think?

imapokemon4:
@Blaisorblade 
  
Did I do it correctly? Do I have to rebase again after that merge?

imapokemon4:
@Blaisorblade 
  
I tried some more rebasing and squashing. I'll leave it alone now until you tell me what I need to do next.

Blaisorblade:
I'm now inclined to merge this: long phrases are especially a problem when the spinners are closed (after picking, in case you're rechecking), beyond being too long to read.

I'll wait a bit for opinions.

nahojjjen:
My opinion is that it looks good.

PR #382, opened 2016-09-06
Makes overlay move when appraisal & color codes team picker

nahojjjen:


app/src/main/java/com/kamron/pogoiv/widgets/PlayerTeamAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.ArrayList;
+
+/**
+ * Created by Johan on 2016-09-06.
+ * An adapter class that makes the teams color coded
+ */
+public class PlayerTeamAdapter extends BaseAdapter implements SpinnerAdapter {
+
+    ArrayList<Integer> colors;
+    Context context;
```
Blaisorblade: Can we keep these private? I know we haven't cared but we've been getting better—I'd like new code to try.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -248,12 +250,9 @@ public void onClick(View v) {
      */
     private void initiateTeamPickerSpinner() {
         Spinner spinner = (Spinner) findViewById(R.id.teamPickerSpinner);
-        ArrayAdapter<CharSequence> adapter = ArrayAdapter.createFromResource(this,
-                R.array.teams, android.R.layout.simple_spinner_item);
-        adapter.setDropDownViewResource(android.R.layout.simple_spinner_dropdown_item);
+        PlayerTeamAdapter adapter = new PlayerTeamAdapter(this);
```
Blaisorblade: 👍

Blaisorblade:
I'm fine with the code (except for the small comments I left). Cool stuff. 👍

I'd like a look by @oskomonzon if possible—after that @nahojjjen can merge.

Left to do:
- [ ] hide more stuff so the appraisal menu item is actually clickable—I showed a bad prototype in chat (see https://github.com/farkam135/GoIV/tree/hide-enough-to-see-appraisals), we could have better.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -740,8 +782,37 @@ public void toggleMoreResultsBox() {
...
+    }
+
+    /**
+     * Moves the overlay down to the bottom of the screen
+     */
+    private void moveOverlayDown() {
+        WindowManager.LayoutParams newParams = (WindowManager.LayoutParams) infoLayout.getLayoutParams();
+        newParams.gravity = Gravity.BOTTOM;
+        windowManager.updateViewLayout(infoLayout, newParams);
     }
```
oskomonzon: moveOverlayDown is mostly duplicate code from moveOverlayUpOrDownToMatchAppraisalBox without the logic.

Maybe we can instead create a moveOverlayUp method and leave just the logic in moveOverlayUpOrDownToMatchAppraisalBox to avoid duplication calling the other two when needed.

Other way could be passing a parameter to "moveOverlay" true to "up" and false to "down" or something like that, since otherwise moveOverlayUp and moveOverlayDown would, once again, share most of the code.

oskomonzon: I'm thinking something like...

`private void moveOverlay(Boolean moveUp) {
	WindowManager.LayoutParams newParams = (WindowManager.LayoutParams) infoLayout.getLayoutParams();
	if(moveUp){
		newParams.gravity = Gravity.BOTTOM;
	}else{
		newParams.gravity = Gravity.BOTTOM;
	}
	windowManager.updateViewLayout(infoLayout, newParams);
}`

Can then be casted as 
`moveOverlay(appraisalBox.getVisibility() == View.VISIBLE);`

oskomonzon:
Other than that, LGTM :+1:

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: @oskomonzon I took the liberty of reformatting your code so I could read it.

Blaisorblade:
I'd really like PR comments to be addressed before merging, though I don't want the refactorings to get conflicted. (But not much time today to work on GoIV).

app/src/main/java/com/kamron/pogoiv/widgets/PlayerTeamAdapter.java
harkin: they can also be `final` if you're changing it

Blaisorblade:
I've merged the first commit (with refactorings) directly into master.

Issue #373, opened 2016-09-05
Ease seeing appraisal while inputting it


Blaisorblade:
Following up from @rpgrca in https://github.com/farkam135/GoIV/issues/327#issuecomment-244755425:

>> I'm afraid it might be a bit too much to ask for someone to remember three or more piece of information regarding appraisal.

>> Maybe it would be better if GoIV also stayed on top of the appraisal information, slightly above so that the user can read and add the information as the trainer is mentioning it. Or just two independent screens altogether, one that pops up when the appraisal information is being read so that you can complete it, with the other being the usual one with the appraisal information incorporated.

@nahojjjen adds:

> I agree with the memory problem, Ive been using the appraisal version for a day now & it was annoying. I'll see if i can find some easy way to "move up" the entire info dialog when the appraisal box is expanded so the user can click appraise without closing the dialog. Pokefly is pretty messy though so I might not be able to

oskomonzon:
IMHO this should be taken into account while we refactor Pokefly and implement the solution after that. It may be messy to achieve with the current implementation.

Blaisorblade:
@oskomonzon Can you elaborate? Do you refer to @nahojjjen's plan or to @rpgrca's?

rpgrca:
Been using the latest code with the extra appraisals. It's usable, I had to redo a few times because I had forgotten the correct appraisal. I noticed a couple of things, so I give some feedback based on those:

1. Only way to retain all the information is to remember just one word out of each appraisal. In my case, as Team Valor, I remember "like", "decent", "proud" and "amazes" for the percent phrase, and "greatness", "job", "excellent" and "WOW" for the range. So, my suggestion would be to highlight each of the words (or similar, but just one for each phrase) so that the final user can create the same association. I'm unsure whether having the whole phrase (as it's now) can be that useful, a combo with just one word (maybe even letting the user choose which word he will remember for each appraisal) could be much faster.
2. The first thing I complete are the checkboxes because they don't require extra clicks. Was thinking that maybe it should be at the top instead of the middle of both appraisals.

leppie:
I would prefer it if just remembered, then you can close GoIV, run appraisal again and fix/add missing info.

oskomonzon:
@Blaisorblade I was talking about @nahojjjen's idea of moving up the inputs layout. But this morning I took a better look at the code and it's entirely possible:

[...]

PR #384, opened 2016-09-09
Ease seeing appraisal while inputting it - move window

Eccenux:
Not sure if there is some refactoring going on, but here is my take on "Ease seeing appraisal while inputting it #373".

Tried detecting swipe gestures, but that's a mess... So I just move the info window up/down when apprise is open.

To save some space I changed first option for appraisal spinners to serve as a label.

Also shortened phrases for Valor.


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -741,7 +751,23 @@ public void toggleMoreResultsBox() {
...
+     */
+    private void moveInfoWindow(boolean toTop) {
+        if (toTop) {
+            layoutParams.gravity = Gravity.CENTER | Gravity.TOP;
+            Log.d(TAG, "moveInfoWindow: top");
+        } else {
+            layoutParams.gravity = Gravity.CENTER | Gravity.BOTTOM;
+            Log.d(TAG, "moveInfoWindow: bottom");
+        }
+        windowManager.updateViewLayout(infoLayout, layoutParams);
```
Blaisorblade: I'd expect this doesn't go all the way to the top? That's not enough—on a Nexus 5 one needs to go all the way to the top *and* to hide other widgets for this to work.

Blaisorblade:
Hi, thanks for your PR! But there's already #382 with a similar approach, maybe you can compare? In particular see:
https://github.com/farkam135/GoIV/pull/382/commits/a40d4d0352fc4a732abc9f3eed25bcbc5a965250

Also another PR, #379 by @imapokemon4, makes the phrases more complete, not shorter. I have no strong opinion on this.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Eccenux: There is a lot of space once labels are gone. Below is on Xperia Z5 Compact (smaller then Nexus 5).
goiv_appraisal_

Blaisorblade: > There is a lot of space once labels are gone.

This... actually looks good. Is there a need to shorten phrases? I'll try merging the rest.

Eccenux: > This... actually looks good. Is there a need to shorten phrases? I'll try merging the rest.

Short phrases, no, not really. But you would need to preserve the first option in each group. This need to be something meaningful (as there are no labels).

Blaisorblade: What do you think of 7f3fdcad402cbc9ab92787a70ed7a24baf8a6f52? That might need further refinement. @DJCrapsody @nahojjjen and @imapokemon4 what do you think? (I'd probably at least go for capitals).

> But you would need to preserve the first option in each group. This need to be something meaningful (as there are no labels).

I've kept your first label there. Having something meaningful could suggest longer phrases.

Blaisorblade:
Closing this—let's continue the discussion on labels in #379.

PR #386, opened 2016-09-09
Prevent checkstyle violations through Travis

Blaisorblade:
Now that I fixed all violations, enable checkstyle on Travis.

* Turn Checkstyle violations into errors.
* Run Checkstyle on Travis.
* Update the uploaded style for Android Studio.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -8,7 +8,7 @@
 
 ## Project Set-up for GoIV (Required)
 * [Fork GoIV Repo] (https://help.github.com/articles/fork-a-repo/)
-* Copy GoIVCodeStyle.xml into the codestyles folder under the Android Studio preferences folder *(create codestyles folder if it does not exist)*
+* Copy `GoIVCodeStyle.xml` into the codestyles folder under the Android Studio preferences folder *(create codestyles folder if it does not exist)*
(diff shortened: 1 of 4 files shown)


Blaisorblade:
Since this would affect all future PRs, I'm interested in opinions of all current contributors (I'm not proposing this requires a unanimity).
@nahojjjen @sarav @DJCrapsody @oskomonzon

sarav:
My only nitpick is generally line size. But we agreed to disagree there. So, as long as this enforces style as we edit and doesn't break android studio, I'm all for it.

DEVELOPMENT.md
```diff
@@ -35,6 +42,9 @@ Set-up Git on Android Studio to automate Git commands. For developers who are *l
 
 ## Contributing with Pull Requests
 
+### Ensure Commits Follow Coding Style
+Use Android Studio autoformatting and check with CheckStyle that your PR does not add coding style violations.
```
Blaisorblade: Should this be more tentative? "Try to ensure that..."?

Blaisorblade:
I've added instructions for configuring Android Studio. Note I had to change the coding style to make this work.

Also note: reformat will *not* fix all errors—the style file has ideas on names and Javadocs.
I still think this workflow is worth a try.

PR #389, opened 2016-09-10
Use Optional for scan failures in HP/CP, avoid defaulting to 10

### LINKED ISSUE #374: Stop using 10 when failing to read a number
Rather use `Optional<Integer>` to represent OCRed numbers.

EDIT: reasons:
- failing to OCR the CP is no reason to write 10 in an input box so the user has to delete it; we could later even handle this differently in the GUI (by highlighting the missing input, etc.)
- we have code that compares the OCRed numbers to 10 to check if scanning failed, and I expect that might misbehave in edge cases.

Blaisorblade:
Fix #374.

I'm not done on failures for candy recognition though: it seems that code can't fail, so I dropped the current catch block for it as dead code. There's other code depending on that.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -132,6 +132,7 @@ dependencies {
     compile 'com.jakewharton:butterknife:8.4.0'
     compile 'com.jakewharton.timber:timber:4.3.0'
     compile 'com.rmtheis:tess-two:6.0.4'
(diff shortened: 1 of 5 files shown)


Blaisorblade:
The Travis failure in https://travis-ci.org/farkam135/GoIV/builds/158892753 was spurious—a JVM crash!

PR #390, opened 2016-09-10
Screen grabbing performance: cache only one image

Blaisorblade:
Hopefully this improves performance noticeably, and I assume it can't be worse. Docs make it clear that this saves memory; hopefully this will smart enough to not actually copy the image except when we ask it to.

I want to test this again, but this seems to be working well.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
@@ -32,7 +32,7 @@ public class ScreenGrabber {
         displayMetrics = display;
         projection = mediaProjection;
         imageReader = ImageReader.newInstance(rawDisplayMetrics.widthPixels, rawDisplayMetrics.heightPixels,
-                PixelFormat.RGBA_8888, 2);
+                PixelFormat.RGBA_8888, 1);
         projection.createVirtualDisplay("screen-mirror", rawDisplayMetrics.widthPixels, rawDisplayMetrics.heightPixels,
                 rawDisplayMetrics.densityDpi, DisplayManager.VIRTUAL_DISPLAY_FLAG_PUBLIC, imageReader.getSurface(),
                 null, null);
@@ -82,9 +82,9 @@ public class ScreenGrabber {
         try {
             //Note: imageReader shouldn't be null, but apparently sometimes is.
             //Let's allow this to still happen.
-            image = imageReader.acquireLatestImage();
+            image = imageReader.acquireNextImage();
         } catch (Exception exception) {
-            Timber.e("Error thrown in grabScreen() - acquireLatestImage()");
+            Timber.e("Error thrown in grabScreen() - acquireNextImage()");
             Timber.e(exception);
         }
 


Blaisorblade:
Hm. This gives once per check the following error, though everything works:
```
09-10 17:43:03.643 24864-24864/com.kamron.pogoiv W/ImageReader_JNI: Unable to acquire a lockedBuffer, very likely client tries to lock more than maxImages buffers
```
but that doesn't make much sense to me—I'll be debugging.

Blaisorblade:
Ah, just read docs on `acquireLatestImage` vs `acquireNextImage`. The docs strongly suggest `acquireLatestImage`, but I think for our use case `acquireNextImage` with a size of 1 makes more sense (and avoids the warning, and appears to work).

Blaisorblade:
@sarav opinions?

sarav:
@Blaisorblade I already read this and thought of making the change a week or two ago. But based on the docs, I think we can get a stale image. Saw 10 frames are generated in a second. The first frame might not be the one with the pokemon screen. If we acquireNextImage (with a buffer of 1) we could just have the first frame buffered and the other 10 dropped. acquireLatestImage as I understand, drops all buffered images and gets the latest one

I don't think it'll work reliably 😢

Blaisorblade:
OK, I'll have to see more.

> If we acquireNextImage (with a buffer of 1) we could just have the first frame buffered and the other 10 dropped. acquireLatestImage as I understand, drops all buffered images and gets the latest one

The docs are confusing. They specifically say there's a queue of fixed size (2 before this patch) and that when this queue is full the producer gets stuck. But `acquireLatestImage` will discard at most one image with our settings. If images in the queue are not overwritten by screen capture, why would the current code work?

sarav:
>  and that when this queue is full the producer gets stuck.

I'm pretty sure the producer (the game or the display code) continues to produce frames on screen. I think what this line means is that the producer gets stuck in terms of being able to send us more frames.

> But acquireLatestImage will discard at most one image with our settings.

I guess I meant it'll drop all images in our case? It's been a while since I read the docs and looked at the code. But acquireLatestImage drops 1 image because it makes room for the producer to put in the latest image. So, that's why it works now.

Blaisorblade:
This code is only fully correct if the single buffer is overwritten by the producer. The code seems instead to do what you suggest, but in that scenario this code is not really worse than what we have IIUC. But I'm still rather hesitant, I should probably ask on StackOverflow.

> But acquireLatestImage drops 1 image because it makes room for the producer to put in the latest image. So, that's why it works now.

That suggests that the producer can only write the latest image *after* the check. In this case, having one or two stale images (which are stale by one check intervals) seems to make little difference.

IOW, taking your example:

> Saw 10 frames are generated in a second. The first frame might not be the one with the pokemon screen. If we acquireNextImage (with a buffer of 1) we could just have the first frame buffered and the other 10 dropped. acquireLatestImage as I understand, drops all buffered images and gets the latest one

If your reasoning, right now with a buffer of 2, 2 images are buffered and the other 8 are dropped, and `acquireLatestImage` would get the second. `acquireNextImage` with a buffer of 1 would get the 1st frame.

sarav:
No, acquireLatestImage drops one of the first 2 images and picks up the 10th. That's how I understand it.

Blaisorblade:
Well, closing for now. I might reopen if I find more evidence.

PR #393, opened 2016-09-10
Refactor and clean up ScreenShot handling code and improve device support

### LINKED ISSUE #325: Fix instructions for screenshot mode
Splitting from #273 (which came from #217), since it didn't get attention there:

> Change instructions when selecting battery saver mode (instructions continue to say "Press IV button" when it should be "take a screenshot", might confuse people who forgot they have turned the option on and then don't see the button anywhere).

@DJCrapsody Do you think you can fix this by 3.1.0?

sarav:
Moved all ScreenShot handling code into a separate class. This also fixes battery saver mode not working for a lot of devices (surprisingly, there's not issue for this. Or at least I couldn't find one).

Also fixes #325 indirectly.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
@@ -12,6 +12,9 @@ import android.os.Handler;
 import android.provider.MediaStore;
 import android.support.v4.content.LocalBroadcastManager;
 
+import java.io.File;
+import java.util.Calendar;
+
 /**
  * Created by Sarav on 9/9/2016.
  */
@@ -49,6 +52,15 @@ public class ScreenShotHelper {
 
                 final String pathChange = getRealPathFromUri(context, uri);
                 if (!pathChange.contains("Screenshot")) {
+                    return;
+                }
+
+                /* Ignore random events related to opening old screenshots by making
+                 * sure the file was created within the past 10 seconds.
+                 */
+                long now = Calendar.getInstance().getTimeInMillis();
+                long filetime = new File(pathChange).lastModified();
+                if (now - filetime > 10000) {
                     return;
                 }
 
--- a/app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
@@ -72,6 +72,12 @@ public class ScreenShotHelper {
         return instance;
     }
 
+    public void deleteScreenShot(String filePath) {
+        contentResolver.delete(MediaStore.Images.Media.EXTERNAL_CONTENT_URI, MediaStore.Files.FileColumns.DATA + "=?",
+                                new String[]{filePath});
+
+    }
+
     public void stop() {
         contentResolver.unregisterContentObserver(mediaObserver);
         contentResolver = null;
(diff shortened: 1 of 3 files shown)


sarav:
@Blaisorblade wanna review this?

The left over commits us

sarav:
*The left over commit is just deleting code from MainActivity and adding the throttling of the event handling.

sarav:
@Blaisorblade @farkam135 

One of you want to review this?

sarav:
Tested the last commit too. Works well for me.

app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
```diff
@@ -0,0 +1,47 @@
...
+public class ScreenShotHelper {
+
+    private static ScreenShotHelper instance = null;
+    private FileObserver screenShotScanner;
+
+    private ScreenShotHelper(final Context context, final String screenShotDir) {
+        screenShotScanner = new FileObserver(screenShotDir, FileObserver.CLOSE_NOWRITE | FileObserver.CLOSE_WRITE) {
+            @Override
+            public void onEvent(int event, String file) {
+                if (file != null) {
```
Blaisorblade: Memo: This commit stops using `readyForNewScreenshot` silently. Hopefully later commits clean this up.

app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
```diff
@@ -1,47 +1,81 @@
...
+                if (!pathChange.contains("Screenshot")) {
+                    return;
+                }
+
+                try {
+                    Bitmap bitmap = BitmapFactory.decodeFile(pathChange);
+                    Intent newintent = Pokefly.createProcessBitmapIntent(bitmap, pathChange);
                     LocalBroadcastManager.getInstance(context).sendBroadcast(newintent);
+                } catch (Exception e) {
+                    // TODO: Retry a few times after a wait
```
Blaisorblade: Memo: this should also be addressed somewhere.

Blaisorblade:
Review finished. Works for me too on Nexus 5—though the delay is quite noticeable. Merging.

sarav:
Phew! Finally merged. Is this noticeably slower than the file observer method? Any idea if the Nexus generates one or multiple content observer events? 

Another option is to have both observers going, but then file observer is apparently a memory hog. I guess this is what you get for battery saver mode I guess.

Blaisorblade:
> Is this noticeably slower than the file observer method?

Just now it feels noticeably slower than non-battery-saver, haven't really compared with the file observer.

> Any idea if the Nexus generates one or multiple content observer events?

Nope. I guess I'll need to set a breakpoint to catch this. We'll see.

sarav:
Yeah, battery save mode has always been slower than normal mode. I think the content observer notifications get send after adding the image to the Media data base, etc. So, might be even more slower. I'm sure even with FileObserver, it'll take more time since it needs to be written to disk and then read back (or hopefully cached) instead of directly copying from display buffer.

app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
sarav: I'm assuming you saw the later commit that just deletes that flag? We don't really need that flag in addition to the ones in Pokefly. This is part of the effort to reduce it to 1 flag. Something like "isBusyWithUser" that denotes that the IV related UI is active. When that happens, we shouldn't need to process any screens or screenshots.

But then, it's going to be really hard to trigger that anyway. Even if you take a screenshot, the pokemon scan code won't see any pokemon and bail.

PR #396, opened 2016-09-11
Speedup screen check by grabbing only needed pixels, v2

### LINKED ISSUE #391: Stop copying Image into Bitmap at each check
I noticed while debugging #390. I believe this is rather wasteful if we are to just access two pixels. Fixing this would require to index correctly the raw ByteBuffer, but I expect this is the kind of optimization that is wortwhile.

@sarav, opinions?

Blaisorblade:
Fix #391.

This reworks branch speedup-screen-check based on discussion in 4de491198c667ba3c1ccb78f8cace24e57b60d00. I ended up doing the refactoring from scratch, but the new code still works and makes sense.

This also changes pokemon-screen detection to bound-check the individual pixels rather than check if the image is in portrait format. I expect this change to be safe.

Review assigned to @sarav.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
@@ -2,6 +2,7 @@ package com.kamron.pogoiv;
 
 import android.annotation.TargetApi;
 import android.graphics.Bitmap;
+import android.graphics.Color;
 import android.graphics.PixelFormat;
 import android.graphics.Point;
 import android.graphics.Rect;
@@ -163,11 +164,11 @@ public class ScreenGrabber {
     //Inspired by http://stackoverflow.com/a/27655022/53974.
     private static @ColorInt int getPixel(ByteBuffer buffer, Point pos, int pixelStride, int rowStride) {
         int offset = pos.y * rowStride + pos.x * pixelStride;
-        int pixel = 0;
-        pixel |= (buffer.get(offset) & 0xff) << 16;     // R
-        pixel |= (buffer.get(offset + 1) & 0xff) << 8;  // G
-        pixel |= (buffer.get(offset + 2) & 0xff);       // B
-        pixel |= (buffer.get(offset + 3) & 0xff) << 24; // A
-        return pixel;
+        //This works because the image reader is configured with PixelFormat.RGBA_8888.
+        int r = buffer.get(offset) & 0xff;
+        int g = buffer.get(offset + 1) & 0xff;
+        int b = buffer.get(offset + 2) & 0xff;
+        int a = buffer.get(offset + 3) & 0xff;
+        return Color.argb(a, r, g, b);
     }
 }
--- a/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
@@ -115,6 +115,11 @@ public class ScreenGrabber {
         return bmp;
     }
 
+    /**
+     * Grab a few pixels from the current screen.
+     * @param points array of points representing coordinates to grab
+     * @return array of colors for the requested pixels, or null if any of them is out-of-bounds
+     */
     public @Nullable @ColorInt int[] grabPixels(Point[] points) {
         Image image = null;
         try {
--- a/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
@@ -4,6 +4,7 @@ import android.annotation.TargetApi;
 import android.graphics.Bitmap;
 import android.graphics.PixelFormat;
 import android.hardware.display.DisplayManager;
+import android.hardware.display.VirtualDisplay;
 import android.media.Image;
 import android.media.ImageReader;
 import android.media.projection.MediaProjection;
@@ -25,6 +26,7 @@ public class ScreenGrabber {
     private MediaProjection projection = null;
     private DisplayMetrics rawDisplayMetrics;
     private DisplayMetrics displayMetrics;
+    private VirtualDisplay virtualDisplay;
 
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     private ScreenGrabber(MediaProjection mediaProjection, DisplayMetrics raw, DisplayMetrics display) {
@@ -33,7 +35,8 @@ public class ScreenGrabber {
         projection = mediaProjection;
         imageReader = ImageReader.newInstance(rawDisplayMetrics.widthPixels, rawDisplayMetrics.heightPixels,
                 PixelFormat.RGBA_8888, 2);
-        projection.createVirtualDisplay("screen-mirror", rawDisplayMetrics.widthPixels, rawDisplayMetrics.heightPixels,
+        virtualDisplay = projection.createVirtualDisplay("screen-mirror", rawDisplayMetrics.widthPixels,
+                rawDisplayMetrics.heightPixels,
                 rawDisplayMetrics.densityDpi, DisplayManager.VIRTUAL_DISPLAY_FLAG_PUBLIC, imageReader.getSurface(),
                 null, null);
     }
@@ -58,6 +61,8 @@ public class ScreenGrabber {
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     public void exit() {
         if (projection != null) {
+            virtualDisplay.release();
+            virtualDisplay = null;
             imageReader.close();
             imageReader = null;
             projection.stop();
--- a/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
+++ b/app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
@@ -12,7 +12,6 @@ import android.media.ImageReader;
 import android.media.projection.MediaProjection;
 import android.os.Build;
 import android.support.annotation.ColorInt;
-import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 import android.util.DisplayMetrics;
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -58,6 +61,8 @@ public static ScreenGrabber getInstance() {
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     public void exit() {
         if (projection != null) {
+            virtualDisplay.release();
```
sarav: I don't think this is needed. Once we lose the reference to the all the existing private variables (by setting them to null below) it's going to get garbage collected (and released and whatever else would be needed). The only existing reference to the Virtual display before this change would be inside the "projection".

But I'm not going to push you to drop this change. No strong opinion on this.

I think this is also the case for the imageReader.close() you added below in another commit.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -58,6 +62,35 @@ public static ScreenGrabber getInstance() {
         return instance;
     }
 
+    @NonNull private static Rect getBitmapBounds(Bitmap bmp) {
+        return new Rect(0, 0, bmp.getWidth(), bmp.getHeight());
+    }
+
+    static @Nullable @ColorInt int[] getPixelsSafe(Bitmap bmp, Point[] points) {
```
sarav: Sorry, what's the point of this bounds checking again?

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -62,35 +61,6 @@ public static ScreenGrabber getInstance() {
         return instance;
     }
 
-    @NonNull private static Rect getBitmapBounds(Bitmap bmp) {
-        return new Rect(0, 0, bmp.getWidth(), bmp.getHeight());
-    }
-
-    static @Nullable @ColorInt int[] getPixelsSafe(Bitmap bmp, Point[] points) {
```
sarav: Nitpick, I don't like moving functions inside a file as part of a commit that's doing something else. Makes it hard to review because all the lines show up as "new".

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -144,4 +114,55 @@ private Bitmap getBitmap(ByteBuffer buffer, int pixelStride, int rowPadding) {
 
         return bmp;
     }
+
+    public @Nullable @ColorInt int[] grabPixels(Point[] points) {
+        Image image = null;
+        try {
+            //Note: imageReader shouldn't be null, but apparently sometimes is.
+            //Let's allow this to still happen.
```
sarav: Is this still true? We should find and fix this.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -144,4 +114,55 @@ private Bitmap getBitmap(ByteBuffer buffer, int pixelStride, int rowPadding) {
...
+
+        Image.Plane imgPlane = image.getPlanes()[0];
+        int pixelStride = imgPlane.getPixelStride();
+        int rowStride = imgPlane.getRowStride();
+        ByteBuffer buffer = imgPlane.getBuffer();
+
+        for (int i = 0; i < points.length; i++) {
+            Point p = points[i];
+            if (imageBounds.contains(p.x, p.y)) {
+                pixels[i] = getPixel(buffer, p, pixelStride, rowStride);
```
sarav: I'm beginning to wonder if there's any guarantee that this buffer we get from ImageReader is guaranteed to be an uncompressed image. What is the projections writes the image into the ImageReader in PNG format? Then, directly accessing the buffer isn't going to work?

Also, taking a closer look at the Bitmap docs and grabScreen() I think a simpler code with the same benefits (the main benefit of this change is to avoid copyPixelsFromBuffer) would be to use [this constructor](https://developer.android.com/reference/android/graphics/Bitmap.html#createBitmap(int[], int, int, android.graphics.Bitmap.Config)) when creating the bitmap in grabScreen? Since this constructor is supposed to produce and "immutable" Bitmap, I think it doesn't do any buffer copying. The current code needs copyFromBuffer because we first create a mutable Bitmap and then set the pixel values. Thoughts?

While also led me to another confusion. The [Bitmap.createBitmap() calls we use in OcrHelper](https://developer.android.com/reference/android/graphics/Bitmap.html#createBitmap(android.graphics.Bitmap, int, int, int, int)) to cut out the pieces of the bitmap are supposed to return an immutable Bitmap. However, inside replaceColors, we still getPixels, change the values and setPixels? So, how the heck are we able to mutate an immutable Bitmap? How is .setPixels even allowed on an immutable object? Or am I misunderstanding the meaning of immutable?

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -144,4 +114,55 @@ private Bitmap getBitmap(ByteBuffer buffer, int pixelStride, int rowPadding) {
...
+    }
+
+    //Inspired by http://stackoverflow.com/a/27655022/53974.
+    private static @ColorInt int getPixel(ByteBuffer buffer, Point pos, int pixelStride, int rowStride) {
+        int offset = pos.y * rowStride + pos.x * pixelStride;
+        int pixel = 0;
+        pixel |= (buffer.get(offset) & 0xff) << 16;     // R
+        pixel |= (buffer.get(offset + 1) & 0xff) << 8;  // G
+        pixel |= (buffer.get(offset + 2) & 0xff);       // B
+        pixel |= (buffer.get(offset + 3) & 0xff) << 24; // A
```
sarav: Yeah, the image reader using the ARGB format should mean that you can assume the pixel buffer stores values in ARGB format.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
```diff
@@ -144,4 +114,55 @@ private Bitmap getBitmap(ByteBuffer buffer, int pixelStride, int rowPadding) {
...
+            Timber.e("Error thrown in grabPixels() - acquireLatestImage()");
+            Timber.e(exception);
+        }
+
+        if (image == null) {
+            return null;
+        }
+
+        Rect imageBounds = new Rect(0, 0, image.getWidth(), image.getHeight());
+        @ColorInt int[] pixels = new int[points.length];
```
sarav: What does @ColorInt int [] mean? I can see that it's some annotation, but how does it help?

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
sarav: Also, I think you should just use Color.argb() instead of reimplementing it.

sarav:
Done with the review. Is there any way to assign a review to me in github where I can review and then close it? I had to dig through my read notifications to find this PR. And sometime I miss review requests. Being able to assign review tasks to me would be helpful.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: You have, but you said you want to change the timer we use *after* the big refactoring.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: Clearing fields of the instance which is going to be GCed is should be useless (under some conditions), but I'm doing it for consistency. Releasing resources is different.

Finalizers are a parody of destructors and aren't invoked timely (Google `finalizers bad` for info), and `VirtualDisplay` doesn't even have a finalizer. Yes, Java is worse than C++ here. (Better-working alternatives include a `Thread` and a `ReferenceQueue`).

https://developer.android.com/reference/android/hardware/display/VirtualDisplay.html#release()
> Because a virtual display renders to a surface provided by the application, it will be released automatically when the process terminates and all remaining windows on it will be forcibly removed. However, you should also explicitly call release() when you're done with it.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: I'm using it for documentation value; I'm not sure anybody implemented the appropriate warnings when converting from `int` to `@ColorInt int`.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: Thanks for `Color.argb()`.
But note the image reader uses the `PixelFormat.RGBA_8888` format, not a non-existing `PixelFormat.ARGB_8888`—which is one reason why this code actually reorders the bytes.
Viceversa, the bitmaps we create use `Bitmap.Config.ARGB_8888` instead of a non-existing `Bitmap.Config.RGBA_8888`.

(Not that I realized either when I copy-pasted this, otherwise I'd have called this out in a comment).

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: Re mutable bitmaps: the implementation of `createBitmap` we use will often return a mutable bitmap—in fact, if I trust JavaDocs of other invoked `createBitmap` methods,  the `createBitmap` overload you mention will **always** return a mutable bitmap, unless the original bitmap is immutable and can be reused.

Blaisorblade: > Also, taking a closer look at the Bitmap docs and grabScreen() I think a simpler code with the same benefits (the main benefit of this change is to avoid copyPixelsFromBuffer) would be to use this constructor when creating the bitmap in grabScreen?

That constructor takes a int array in ARGB format. Our `Image` returns a ByteBuffer in RGBA format, from which we *might* be able to get a `byte[]` array (again in RGBA), not an `int[]` array (in ARGB). As discussed [below](https://github.com/farkam135/GoIV/pull/396#discussion_r78292182), I don't see a way to match the formats. And I don't think a `byte[]` can be cast to an `int[]` without JNI somewhere.

I've looked into `BitmapFactory` multiple times, but according to its docs (and to the available options) doesn't seem to be addressed at our use case but at files, and StackOverflow seems to agree (http://stackoverflow.com/q/26673127/53974; I wouldn't trust it if I hadn't checked the docs).

I'm not sure why the APIs are designed to *require* a conversion, but this conflict has enough precedent to give rise to a confusing Wikipedia page (which I don't really trust):
https://en.wikipedia.org/wiki/RGBA_color_space#Representation.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: The points are valid for portrait images, not for landscape images, and we totally got crashes from landscape images when users rotated their phones.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: I thought about that, but here there's little in common.

Blaisorblade:
> Is there any way to assign a review to me in github where I can review and then close it? I had to dig through my read notifications to find this PR. And sometime I miss review requests. Being able to assign review tasks to me would be helpful.

I can assign PRs to people with commit access, but I can't give that to you.

Blaisorblade:
> Nice changes and good commit splitting.

Thanks!

I've responded to your comments and switched to `Color.argb`. I think zero-copy-reading the buffer isn't gonna work so shouldn't keep from merging this code, but of course if you get it to work you can do a new PR later.

sarav:
Sounds good. Merge.

app/src/main/java/com/kamron/pogoiv/ScreenGrabber.java
Blaisorblade: Those crashes were fixed by checking if the image was portrait/landscape, but we still got further crashes (though, it turns out, not here): http://crashes.to/s/c9a025fd5e9. Opened #420 to track that.

PR #398, opened 2016-09-11
Simplify Nidoran correction

### LINKED ISSUE #381: Use nidoran gender recognition also when nidoran's detected from the candy
We only use gender recognition `if (pokemonName.toLowerCase().contains("nidora")` (see `OCRHelper`), but that fails for renamed pokemons.

But apparently also candies can be misgendered:
https://github.com/farkam135/GoIV/pull/308#issuecomment-244835226

We "just" need to move Nidoran gender recognition at a different point, without duplicating code or doing it twice. `PokemonNameCorrector` doesn't have access to the image (and probably shouldn't) so is not an option. It might be enough to *move* Nidoran gender recognition to the code OCRing candy.

Ideas?

sarav:
We can just check the last character of the OCR text to identify the
proper gender of the Nidoran. We don't need to do a color check. With
this change, we can apply the Nidoran correction to the candy name
before we search for the final pokemon. This will be useful for future
improvements to Pokemon detection.

Fixes #381

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -28,10 +28,8 @@ public class OcrHelper {
     private final int heightPixels;
     private final int widthPixels;
     private final boolean candyWordFirst;
-    private final String nidoFemale;
-    private final String nidoMale;
 
-    private OcrHelper(String dataPath, int widthPixels, int heightPixels, String nidoFemale, String nidoMale) {
+    private OcrHelper(String dataPath, int widthPixels, int heightPixels) {
         tesseract = new TessBaseAPI();
         tesseract.init(dataPath, "eng");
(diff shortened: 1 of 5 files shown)


sarav:
@Blaisorblade Travis is super helpful. It lets me catch all the checkstyle errors since I haven't figured out how to do it on my end.

nahojjjen:
This is close to how it originally worked, but we changed it because:

1. This implementation does not take different nidoran names into account in different languages, will only work in english

2. Depending on what phone & resolution you have, the gender symbol is recognized as different symbols. Female nidoran on my S7 is recognized as NidoranZ, and thus this implementation does not work.

Blaisorblade:
I wanted to say "OCR heuristics should only be changed after extensive testing", but we already have one counterexample. 😢  So I'm afraid we're going to have to fix #381 by keeping the same heuristic.

sarav:
@nahojjjen, this implementation should still work for other languages, but the other issue of OCR giving a different result on each phone is a valid point. I can drop this or if we can get a list of common OCR errors for those gender characters, we could get this to work.

PR #399, opened 2016-09-12
Don't do unnecessary null to empty string conversion

sarav:
We can just check for null instead of an empty string.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -295,7 +295,7 @@ public class Pokefly extends Service {
         intent.putExtra(KEY_SEND_INFO_HP, scanResult.getPokemonHP());
         intent.putExtra(KEY_SEND_INFO_CP, scanResult.getPokemonCP());
         intent.putExtra(KEY_SEND_INFO_LEVEL, scanResult.getEstimatedPokemonLevel());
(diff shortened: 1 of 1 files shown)


Blaisorblade:
BTW, nullness checkers might help, and this question lists a few reasonable checkers: http://stackoverflow.com/q/14076296/53974. 
[This link](http://types.cs.washington.edu/checker-framework/current/checker-framework-manual.html#other-tools) compares a few.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -946,10 +946,11 @@ private Pokemon interpretWhichPokemonUserInput() {
      * screenshot, and then deletes the screenshot.
      */
     private void deleteScreenShotIfRequired() {
-        if (batterySaver && !ssFile.isEmpty()) {
+        if (batterySaver && ssFile != null) {
```
Blaisorblade: So it turns out the existing code can fail on a NullPointerException :-(.

sarav: Huh? Not sure I understand. Bug in my code or did I fix something?

Blaisorblade: Turned out `ssFile.isEmpty()` can crash so I went checking this PR.

PR #400, opened 2016-09-12
Minor text fixes

imapokemon4:
I removed the same unused variables for the other localizations and rebased.

### DIFF
--- a/app/src/main/res/values-cs/strings.xml
+++ b/app/src/main/res/values-cs/strings.xml
@@ -71,7 +71,4 @@
     <string name="atk">ÚTK</string>
     <string name="def">OBR</string>
     <string name="sta">STA</string>
-    <string name="percent_phrase">Percent phrase</string>
-    <string name="iv_range_phrase">IV range phrase</string>
-    <string name="higher_values">Vyšší hodnoty</string>
 </resources>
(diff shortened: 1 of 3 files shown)


Blaisorblade:
Can you name the commit (and the PR) something like "Removed unused strings"? I get the running joke, but looking at git logs is not meant to be fun 😉

imapokemon4:
Not a joke. It's that same commit you just merged.

imapokemon4:
@Blaisorblade is that correct?

DJCrapsody:
Minor text fixes  

http://knowyourmeme.com/memes/pokemon-go-updates-controversy

Blaisorblade:
Thanks to @DJCrapsody for finding the meme!

> It's that same commit you just merged.

It's not "the same". But it seems a revised version. Please don't. The original commit was merged (which was... probably bad), now we can only merge a new one. Two commits with the same title is typically bad (GitHub even shows again the same changes 😢  which doesn't help).

And neither should be called "Minor text fixes", but should have an informative title. If you disagree, please take a look to https://github.com/farkam135/GoIV/blob/master/DEVELOPMENT.md#writing-good-commit-messages and/or to the linked article.

Blaisorblade:
Apologies if my comments sounded overly negative, that was not my intention but might have been unclear. Since the OP argued about why you used git this way, I thought explaining better practices was appropriate. And I honestly thought the title was an intentional joke, I didn't mean to make fun of the patch.

And I agree the idea behind the change is good.

Blaisorblade:
FTR remaining changes have been merged in 5c9c60fefc7ab461dea81f6a094d7b23a718a2bd, so this PR is addressed.

PR #404, opened 2016-09-12
Notification with actions v2

thearaks:
Hi all!
Here's my new Notification actions PR.
It should be OK, all the previous PR bugs are fixed.

I modified Pokefly onDestroy method removing unnecessary code and put super call to the bottom.
I hope it's not too much invasive for this PR.

I'm looking forward to your feedback!

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -68,8 +68,9 @@ public class MainActivity extends AppCompatActivity {
 
     private boolean batterySaver;
 
-    private boolean pokeFlyRunning = false;
     private int trainerLevel;
+
+    private Button launchButton;
 
     private final Point arcInit = new Point();
     private int arcRadius;
@@ -109,6 +110,17 @@ public class MainActivity extends AppCompatActivity {
             }
         }
         return false;
+    }
+
+    private boolean hasAllPermissions() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !Settings.canDrawOverlays(this)) {
+            return false;
+        }
+        if (ContextCompat.checkSelfPermission(MainActivity.this, Manifest.permission.WRITE_EXTERNAL_STORAGE)
+                == PackageManager.PERMISSION_DENIED) {
+            return false;
+        }
+        return true;
     }
 
     @TargetApi(Build.VERSION_CODES.M)
@@ -151,12 +163,12 @@ public class MainActivity extends AppCompatActivity {
         Display disp = windowManager.getDefaultDisplay();
         disp.getRealMetrics(rawDisplayMetrics);
 
-        Button launch = (Button) findViewById(R.id.start);
-        launch.setOnClickListener(new View.OnClickListener() {
+        launchButton = (Button) findViewById(R.id.start);
+        launchButton.setOnClickListener(new View.OnClickListener() {
 
             @Override
             public void onClick(View v) {
-                if (((Button) v).getText().toString().equals(getString(R.string.main_permission))) {
+                if (!hasAllPermissions()) {
                     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !Settings.canDrawOverlays(
                             MainActivity.this)) {
                         Intent intent = new Intent(Settings.ACTION_MANAGE_OVERLAY_PERMISSION,
@@ -168,7 +180,7 @@ public class MainActivity extends AppCompatActivity {
                         ActivityCompat.requestPermissions(MainActivity.this,
                                 new String[]{Manifest.permission.WRITE_EXTERNAL_STORAGE}, WRITE_STORAGE_REQ_CODE);
                     }
-                } else if (((Button) v).getText().toString().equals(getString(R.string.main_start))) {
+                } else if (!Pokefly.isRunning()) {
                     batterySaver = settings.isManualScreenshotModeEnabled();
                     setupDisplaySizeInfo();
                     trainerLevel = setupTrainerLevel(npTrainerLevel);
@@ -180,18 +192,14 @@ public class MainActivity extends AppCompatActivity {
                     } else {
                         startScreenService();
                     }
-                } else if (((Button) v).getText().toString().equals(getString(R.string.main_stop))) {
+                } else {
                     stopService(new Intent(MainActivity.this, Pokefly.class));
                     if (screen != null) {
                         screen.exit();
                     }
-                    pokeFlyRunning = false;
-                    ((Button) v).setText(getString(R.string.main_start));
                 }
             }
         });
-
-        checkPermissions(launch);
 
         LocalBroadcastManager.getInstance(this).registerReceiver(showUpdateDialog,
                 new IntentFilter(ACTION_SHOW_UPDATE_DIALOG));
@@ -274,6 +282,9 @@ public class MainActivity extends AppCompatActivity {
     @Override
     protected void onResume() {
         super.onResume();
+
+        updateLaunchButtonText(Pokefly.isRunning());
+
         settings = GoIVSettings.getInstance(MainActivity.this);
     }
 
@@ -294,10 +305,18 @@ public class MainActivity extends AppCompatActivity {
         Intent intent = Pokefly.createIntent(this, trainerLevel, statusBarHeight, batterySaver);
         startService(intent);
 
-        pokeFlyRunning = true;
-
         if (settings.shouldLaunchPokemonGo()) {
             openPokemonGoApp();
+        }
+    }
+
+    private void updateLaunchButtonText(boolean isPokeflyRunning) {
+        if (!hasAllPermissions()) {
+            launchButton.setText(R.string.main_permission);
+        } else if (isPokeflyRunning) {
+            launchButton.setText(R.string.main_stop);
+        } else {
+            launchButton.setText(R.string.main_start);
         }
     }
 
@@ -311,27 +330,10 @@ public class MainActivity extends AppCompatActivity {
         return "Error while getting version name";
     }
 
-    /**
-     * Checks to see if all runtime permissions are granted,
-     * if not change button text to Grant Permissions.
-     *
-     * @param launch The start button to change the text of
-     */
-    private void checkPermissions(Button launch) {
-        //Check Permissions
-        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !Settings.canDrawOverlays(this)) {
-            launch.setText(getString(R.string.main_permission));
-        } else if (ContextCompat.checkSelfPermission(MainActivity.this, Manifest.permission.WRITE_EXTERNAL_STORAGE)
-                == PackageManager.PERMISSION_DENIED) {
-            launch.setText(getString(R.string.main_permission));
-        }
-    }
-
     @Override
     public void onDestroy() {
-        if (pokeFlyRunning) {
+        if (Pokefly.isRunning()) {
             stopService(new Intent(MainActivity.this, Pokefly.class));
-            pokeFlyRunning = false;
         }
         if (screen != null) {
             screen.exit();
@@ -345,13 +347,8 @@ public class MainActivity extends AppCompatActivity {
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
         if (requestCode == OVERLAY_PERMISSION_REQ_CODE) {
-            if (!Settings.canDrawOverlays(this)) {
-                // SYSTEM_ALERT_WINDOW permission not granted...
-                ((Button) findViewById(R.id.start)).setText(getString(R.string.main_permission));
-            } else if (ContextCompat.checkSelfPermission(MainActivity.this,
-                    Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED) {
-                ((Button) findViewById(R.id.start)).setText(getString(R.string.main_start));
-            }
+            updateLaunchButtonText(false);
+
         } else if (requestCode == SCREEN_CAPTURE_REQ_CODE) {
             if (resultCode == RESULT_OK) {
                 MediaProjectionManager projectionManager = (MediaProjectionManager) getSystemService(
@@ -361,7 +358,7 @@ public class MainActivity extends AppCompatActivity {
 
                 startPokeFly();
             } else {
-                ((Button) findViewById(R.id.start)).setText(getString(R.string.main_start));
+                updateLaunchButtonText(false);
             }
         }
     }
@@ -381,13 +378,7 @@ public class MainActivity extends AppCompatActivity {
     public void onRequestPermissionsResult(int requestCode, @NonNull String[] permissions,
                                            @NonNull int[] grantResults) {
         if (requestCode == WRITE_STORAGE_REQ_CODE) {
-            if (grantResults.length > 0 && grantResults[0] == PackageManager.PERMISSION_GRANTED) {
-                if (Settings.canDrawOverlays(this) && ContextCompat.checkSelfPermission(MainActivity.this,
-                        Manifest.permission.READ_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED) {
-                    // SYSTEM_ALERT_WINDOW permission not granted...
-                    ((Button) findViewById(R.id.start)).setText(getString(R.string.main_start));
-                }
-            }
+            updateLaunchButtonText(false);
         }
     }
 
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -68,6 +68,12 @@ public class MainActivity extends AppCompatActivity {
 
     private boolean batterySaver;
 
+    /**
+     * This button detects if pokeFly is starting.
+     * Such booleans are error-prone, but we only use it to decide what label to show in launchButton.
+     */
+    private boolean pokeFlyStarting;
+
     private int trainerLevel;
 
     private Button launchButton;
@@ -79,8 +85,10 @@ public class MainActivity extends AppCompatActivity {
         public void onReceive(Context context, Intent intent) {
             if (Pokefly.ACTION_START.equals(intent.getAction())) {
                 updateLaunchButtonText(true);
+                pokeFlyStarting = false;
             } else if (Pokefly.ACTION_STOP.equals(intent.getAction())) {
                 updateLaunchButtonText(false);
+                pokeFlyStarting = false;
             }
         }
     };
@@ -297,7 +305,7 @@ public class MainActivity extends AppCompatActivity {
     protected void onResume() {
         super.onResume();
 
-        updateLaunchButtonText(Pokefly.isRunning());
+        updateLaunchButtonText(Pokefly.isRunning() || pokeFlyStarting);
 
         settings = GoIVSettings.getInstance(MainActivity.this);
     }
@@ -313,7 +321,9 @@ public class MainActivity extends AppCompatActivity {
      * Starts the PokeFly background service which contains overlay logic.
      */
     private void startPokeFly() {
-        launchButton.setText(R.string.main_stop);
+        pokeFlyStarting = true;
+        updateLaunchButtonText(true); //Should be subsumed by onResume later, which seems to be always called in
+        // testing, but I'm not sure that's guaranteed --@Blaisorblade
 
         int statusBarHeight = getStatusBarHeight();
         Intent intent = Pokefly.createIntent(this, trainerLevel, statusBarHeight, batterySaver);
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -74,6 +74,16 @@ public class MainActivity extends AppCompatActivity {
 
     private final Point arcInit = new Point();
     private int arcRadius;
+    private final BroadcastReceiver pokeflyStateChanged = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (Pokefly.ACTION_START.equals(intent.getAction())) {
+                updateLaunchButtonText(true);
+            } else if (Pokefly.ACTION_STOP.equals(intent.getAction())) {
+                updateLaunchButtonText(false);
+            }
+        }
+    };
     private final BroadcastReceiver showUpdateDialog = new BroadcastReceiver() {
         @Override
         public void onReceive(Context context, Intent intent) {
@@ -201,6 +211,10 @@ public class MainActivity extends AppCompatActivity {
             }
         });
 
+        LocalBroadcastManager.getInstance(this).registerReceiver(pokeflyStateChanged,
+                new IntentFilter(Pokefly.ACTION_START));
+        LocalBroadcastManager.getInstance(this).registerReceiver(pokeflyStateChanged,
+                new IntentFilter(Pokefly.ACTION_STOP));
         LocalBroadcastManager.getInstance(this).registerReceiver(showUpdateDialog,
                 new IntentFilter(ACTION_SHOW_UPDATE_DIALOG));
 
@@ -339,6 +353,7 @@ public class MainActivity extends AppCompatActivity {
             screen.exit();
         }
 
+        LocalBroadcastManager.getInstance(this).unregisterReceiver(pokeflyStateChanged);
         LocalBroadcastManager.getInstance(this).unregisterReceiver(showUpdateDialog);
         super.onDestroy();
     }
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -313,7 +313,7 @@ public class MainActivity extends AppCompatActivity {
      * Starts the PokeFly background service which contains overlay logic.
      */
     private void startPokeFly() {
-        ((Button) findViewById(R.id.start)).setText(R.string.main_stop);
+        launchButton.setText(R.string.main_stop);
 
         int statusBarHeight = getStatusBarHeight();
         Intent intent = Pokefly.createIntent(this, trainerLevel, statusBarHeight, batterySaver);
@@ -402,7 +402,7 @@ public class MainActivity extends AppCompatActivity {
      */
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     private void startScreenService() {
-        ((Button) findViewById(R.id.start)).setText(R.string.accept_screen_capture);
+        launchButton.setText(R.string.accept_screen_capture);
         MediaProjectionManager projectionManager = (MediaProjectionManager) getSystemService(
                 Context.MEDIA_PROJECTION_SERVICE);
         startActivityForResult(projectionManager.createScreenCaptureIntent(), SCREEN_CAPTURE_REQ_CODE);
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -80,6 +80,7 @@ import io.apptik.widget.MultiSlider;
 
 public class Pokefly extends Service {
 
+    public static final String ACTION_START = "com.kamron.pogoiv.ACTION_START";
     public static final String ACTION_STOP = "com.kamron.pogoiv.ACTION_STOP";
     private static final String ACTION_SEND_INFO = "com.kamron.pogoiv.ACTION_SEND_INFO";
 
@@ -329,6 +330,11 @@ public class Pokefly extends Service {
     @Override
     public void onCreate() {
         super.onCreate();
+
+        running = true;
+
+        LocalBroadcastManager.getInstance(this).sendBroadcast(new Intent(ACTION_START));
+
         displayMetrics = this.getResources().getDisplayMetrics();
         initOcr();
         windowManager = (WindowManager) getSystemService(WINDOW_SERVICE);
@@ -485,6 +491,8 @@ public class Pokefly extends Service {
 
     @Override
     public void onDestroy() {
+        LocalBroadcastManager.getInstance(this).sendBroadcast(new Intent(ACTION_STOP));
+
         LocalBroadcastManager.getInstance(this).unregisterReceiver(displayInfo);
         LocalBroadcastManager.getInstance(this).unregisterReceiver(processBitmap);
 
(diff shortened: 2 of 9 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -72,7 +72,7 @@
     <string name="main_not_numeric">%1$s is not a number.</string>
     <string name="main_invalide_trainerlvl">Invalid Trainer Level!</string>
     <string name="notification_title">GoIV Running - Level %1$s</string>
-    <string name="notification_text">Tap to open</string>
+    <string name="notification_open_app">Open app</string>
```
Blaisorblade: Minor comment: why do we need to change this string (hence retranslate it) if the meaning is the same? Is it a matter of space?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -295,13 +309,42 @@ private void startPokeFly() {
...
+    private boolean hasAllPermissions() {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M && !Settings.canDrawOverlays(this)) {
+            return false;
+        }
+        if (ContextCompat.checkSelfPermission(MainActivity.this, Manifest.permission.WRITE_EXTERNAL_STORAGE)
+                == PackageManager.PERMISSION_DENIED) {
+            return false;
+        }
+        return true;
+    }
```
Blaisorblade: Memo to self: This method can be simplified to `return !cond1 && !cond2;`, but it's a minor matter.

Blaisorblade:
From looking at the code (with my limited Android expertise knowledge), it looks very reasonable, but this requires testing.
The actions for the intents you add seem idempotent, so nothing bad should happen even if those are duplicated.

The only Travis failure is from Checkstyle for line length > 120, hopefully you don't mind fixing that? We've had to enforce it to keep the coding style consistent.

It'd be nice if you also reformatted the commit according to Git guidelines—see here:
https://github.com/farkam135/GoIV/blob/master/DEVELOPMENT.md#writing-good-commit-messages

That doc also explains how to install the Checkstyle plugin if you want... though it's missing info on configuring it.

app/src/main/res/values/strings.xml
thearaks: I changed the string because now that this text is inside an action button it's obviuos it needs to be touched to start the action. Also, shorter text fits better inside the notification.
Also "click here to..." or "tap here to..." are not good UI/UX examples, actions should be self-explanatory through a clear design language. A notification action does the trick!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: I kept the if statements separated to emphasize that we are checking two different permissions. A unified condition would also decrease readability.
Anyway, if you think it's better to simplify the statement I'll fix that ASAP! ;)

thearaks:
I'll add the checkstyle plugin... I don't think I wrote lines > 120 chars, but I probably moved some old long lines around so Travis got upset with my PR!
I need to be more careful and meticulous before submitting ;)
I'll update the PR ASAP!

Thank you @Blaisorblade for all your suggestions!

sarav:
Btw, Travis error report tells exactly what failed even if it's a checkstyle failure. Just click "Details" above and then scroll to the bottom.

sarav:
@thearaks Would you mind splitting up your big commit into smaller "bite-sized" ones? Take #393 as an example.

XML String renames could be one commit, cleaning up the "if" conditions into functions could be another, etc.

I'll take a look tomorrow about how you actually handle the start/stop actions from the notification to the service.

thearaks:
Thank you @sarav, I'll try to split the commits and use better messages; it'll takes some time, in the meanwhile I look forward to your review on the stop Intents from the Notification to Pokefly!

For the checkstyle it should be straightforward, it won't take much!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: Fine then, it's a matter of preference and expectations (especially regarding readability). ✅

sarav:
@thearaks thinking more about it, do we even need this feature now that we added #405 ? The only reason I can think of people wanting to stop/start GoIV is due to the high battery drain. But with that significantly optimized, maybe users won't have a need for this. Or maybe we can at least wait to make a release without this feature, but with #405 and then ask users if they still need this notification actions for stop/start?

We could still add notification actions for incrementing and decrementing Trainer level.

thearaks:
@sarav I'd still merge this since it cleans up a lot of code and notification actions doesn't bother anyone (they even get collapsed if the notification bar gets too cluttered).
Also, it provide handful shortcuts to open GoIV and close it without the need to switch to it via the recent/multitasking virtual button or via the phone launcher.
This PR also enhance the notification style and its attributes so it integrates better with latest API levels and looks a lot nicer. Since we need a notification for startForeground() in Pokefly, why don't use a good looking one? The current one is pretty dull and so "Gingerbread".
And, event with the important enhancements from #405, I'd still use this notification to close GoIV when I'm done with all the stats calculations without exiting the game.
Don't forget that the average user is pretty concerned about running services / freeing RAM: he doesn't like having open apps (think about the compulsive "clear all" gesture on the recent window) so, given this option, they might feel more in control of the situation, without annoying any other user that would just ignore it.
And IMHO, having a fully integrated and styled notification as Google guidelines prescribe, makes this app feel a lot more comprehensive, polished and well-crafted.

Also, adding trainer lever increment/decrement would be very easy with this PR as a starting point, even if I don't find that much useful: this two button would steal a lot of space in the notification bar, and this space could be used better with other notification more important for the user. Also, especially on higher levels, it's not so common to level up, it might take a month or more so the "Open app" action might be enough since it gets you to the MainActivity level spinner with just one swipe down on the status bar and one tap on the notification action.
I don't think it's a good practice to add notification actions for unfrequent tasks, so I'd leave that out unless many users explicitly feels the need for a +1 level quick button!

sarav:
Sorry, I wasn't clear in my earlier comment. I wasn't suggesting we should throw away all the changes. I agree in terms of using a nicer notification, etc. I was just questioning the start/stop actions on the notification (that's kinda what the PR title is all about).

My main concern was that the users will start using the start/stop option way more often than they need to (now that we'd have much lower battery drain) -- exactly like the obsessive "clear all" that you mention that Google itself doesn't recommend (there are other good UI clutter reasons for it). Just because the user wants some obsessive behavior doesn't mean we should provide. 

> I'd still use this notification to close GoIV when I'm done with all the stats calculations without exiting the game.

However, I think the above point gives a reasonable argument for having the start/stop actions. I'm still not too keen on it, but I won't oppose it strongly. Let's see what @Blaisorblade or others think about this.

> Also, especially on higher levels, it's not so common to level up.

Fair enough reason. I definitely won't be needing this realistically. But maybe the lower level users of the app might. Again, don't have a strong opinion. Btw, we should not have an "Open app" action (I mean a button). Just clicking the notification should take us to the app.

Btw, you rattled off a bunch of other benefits of this PR -- so, that clearly demonstrates the need to split up that commit :) Looks like you did and pushed some commits. I'll take a look -- it might take a day or two though. Hope you don't mind waiting. Just too busy at work and I had to address review comments in my PR today.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -343,21 +345,29 @@ private static PokemonNameCorrector initCorrectorFromPrefs(PokeInfoCalculator po
...
-            /* Assumes MainActivity initialized ScreenGrabber before starting this service. */
-            if (!batterySaver) {
-                screen = ScreenGrabber.getInstance();
-                startPeriodicScreenScan();
-            } else {
-                screenShotHelper = ScreenShotHelper.start(Pokefly.this);
+        if (intent != null) {
+            if (ACTION_STOP.equals(intent.getAction())) {
+                if (screen != null) {
+                    screen.exit();
```
sarav: Why are you doing this here and not in onDestroy()? Doesn't stopSelf() below trigger onDestroy() anyway?

thearaks:
Ok @sarav I'll wait for your review!

Anyway, the actions are only "Open app" and "Stop" so we won't support users obsessive-complulsive behaviors!

Overall I think this is a nice addition that doesn't hurt anyone :)

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -343,21 +345,29 @@ private static PokemonNameCorrector initCorrectorFromPrefs(PokeInfoCalculator po
...
-            createIVButton();
-            createArcPointer();
-            createArcAdjuster();
-            /* Assumes MainActivity initialized ScreenGrabber before starting this service. */
-            if (!batterySaver) {
-                screen = ScreenGrabber.getInstance();
-                startPeriodicScreenScan();
-            } else {
-                screenShotHelper = ScreenShotHelper.start(Pokefly.this);
+        if (intent != null) {
```
sarav: To avoid excessive indentation, I'd prefer writing this as:

```
if (intent == null) {
  return START_STICKY;
}

if (ACTION_STOP.equals(....)) {

} else if ( ) {

}
```

Btw, any reason you don't write it as intent.getAction().equals(ACTION_STOP). Personally, they way you've written it reads a bit odd for me. It's like having to write 5 = i instead of writing i = 5. This last part is not a strong opinion.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
thearaks: I followed MainActivity example: there the screen is exited and then stopService is called.
I thought that following the same pattern would be better.
If we put that exit call in onDestroy, it will need to be removed from MainActivity or it'll be called twice when the user taps the Stop button in the activity (don't know if that may lead to unexpected behaviors)... Or, at least, this is what I understood given my little knowledge of ScreenGrabber :)

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -2,6 +2,7 @@
 
 import android.Manifest;
 import android.annotation.TargetApi;
```
sarav: I'm not going to force you on this one, but I feel like this commit can still be split up into meaningful smaller commits.

Your commit text you have is kind of an evidence for that. Those 4 things you list aren't tightly related at all. You could easily split this up into 4 different commits.

Here's some explanation to encourage to split this up into commits. There are a couple of reasons for asking to do this:
1. Much easier to review.
2. Easier to drop a commit and push the rest if one of the changes has a bug that's not easy to fix.
3. More importantly, if we find any bugs in the future in one of your split up commits, we can easily revert that one commit instead of reverting one large commit and losing all the features/benefits because one of them was broken.
4. Reverting smaller commits is also a lot less likely to conflict in the future than one large commit.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -145,12 +147,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
 
-        Button launch = (Button) findViewById(R.id.start);
-        launch.setOnClickListener(new View.OnClickListener() {
+        launchButton = (Button) findViewById(R.id.start);
+        launchButton.setOnClickListener(new View.OnClickListener() {
 
             @Override
             public void onClick(View v) {
-                if (((Button) v).getText().toString().equals(getString(R.string.main_permission))) {
+                if (!hasAllPermissions()) {
```
sarav: Bit of a habit from C coding -- I like functions that you call to come before the caller. Just my 2 cents.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
thearaks: I know about the unnecessary diffs, but the code needed to be refractored.
The null check on the Intent is needed in both cases so embraces the two other checks.

I use CONSTANT.equals(...) as many Google employees do since it avoids NullPointerExceptions: the constant will never be null while the compared value could (since it's out of our control, it comes from a system broadcast)!
Anyway it could be reversed and tested: if in the normal lifecycle of the service null action intents are never received for me it's ok to change!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -174,20 +176,15 @@ public void onClick(View v) {
...
+                } else {
                     stopService(new Intent(MainActivity.this, Pokefly.class));
                     if (screen != null) {
                         screen.exit();
                     }
-                    pokeFlyRunning = false;
-                    ((Button) v).setText(getString(R.string.main_start));
                 }
             }
         });
```
sarav: Thanks for the "if" check clean ups! Much nicer to read.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -295,13 +295,42 @@ private void startPokeFly() {
...
         }
     }
 
+    private boolean isPokeFlyRunning() {
+        ActivityManager manager = (ActivityManager) getSystemService(Context.ACTIVITY_SERVICE);
+        for (ActivityManager.RunningServiceInfo serviceInfo : manager.getRunningServices(Integer.MAX_VALUE)) {
+            if (Pokefly.class.getName().equals(serviceInfo.service.getClassName())) {
+                return true;
+            }
+        }
```
sarav: This seems like an expensive and inefficient way to check if the service is running. This function should be there, but I think it might actually be better to have Pokefly set/unset some flag. Who knows how many services are running on a users phone?!

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: They way I wrote it does handle null correctly for both the ifs. Notice the return inside the null check.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Wait, we are exit()ing screen both here and in MainActivity? That's not nice.

sarav:
Ok, I did some decent amount of review/comments. Looks like you are responding to them already. I don't want to comment at the same time as you are respond -- then you might miss some of my comments.

Since I have already made quite a few comments, I'll wait for you to respond/address those and I'll review again after you upload again or if you insist nothing has to change 😄 (hopefully not).

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: FWIW we used to use some flag, not sure it worked well.

> boolean static variabled that you set/unset when the service starts/stops.

Not static please. If new instances of the Activity are created they shouldn't keep any old value. Docs aren't clear that happens, but posts like http://android-developers.blogspot.de/2009/01/avoiding-memory-leaks.html seem to imply that's possible, and so does e.g. https://developer.android.com/training/basics/activity-lifecycle/recreating.html. Not that the docs are very explicit.
Getting a good answer on StackOverflow seems impossible though.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
thearaks: We are exit()ing screen only once but in different use cases:
A) The user stops Pokefly from MainActivity: MainActivity calls exit()
B) The user stops Pokefly from Notification: Pokefly calls exit()
A and B never happen at the same time :)
Anyway, I think too that it's a bit strange to have that call in 2 different places. So I accept suggestions on how to disambiguate this...

app/src/main/java/com/kamron/pogoiv/Pokefly.java
thearaks: @sarav I see now on my PC (from mobile it was harder to read)
I'm going to refactor that code to follow your example above! It will be much cleaner!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: @sarav You are completely right about this one.
I probably didn't split up this commit into 4 different one just because I hadn't the time to do that.
There's an easy way to split it now?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: I'm sorry @sarav, I can't understand this note... (I'm sorry, I'm not a native speaker)
Can you explain again? :)

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: I don't like that code too, but it's one of the most reasonable ways to accomplish that check.
It doesn't seem to be computationally intensive but I'm open to ideas and suggestions on this!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: @sarav is suggesting `hasAllPermissions` to appear in the file before (that is, above) `onClick`, because `onClick` calls `hasAllPermissions`.

thearaks: Oh, I understand now! I also loved to write in C, but since I started to write Android code I "forgot" that rule.
If you want I can move hasAllPermissions up before onCreate, and I'd do the same with isPokeFlyRunning.
But then I should do the same with startPokeFly and startScreenService... :/

@sarav @Blaisorblade what should I do?

Blaisorblade: Feel free to do that for the new code (I'm indifferent, and @sarav's was phrased as a slight preference). I don't think you have to.

But please don't move around the existing code, both because it's not part of this PR and to reduce conflicts with other changes.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
Blaisorblade: This is a bit tricky, I don't recommend it if you're not comfortable with `git`, nor should you apply the suggestions blindly.
If you want to learn the best approach for doing it, read on; that can be helpful or even essential when contributing to bigger projects. If that's too much, it's OK 😉 .

To split a commit into multiple ones, you can combine `git reset`, `git add -p` and `git commit` in the right sequence. For a commit at the base of a PR, you also need to do an interactive rebase with `git rebase -i`. To see how to combine those (in a very good explanation), the best source I've found is https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Changing-Multiple-Commit-Messages (there are answers on http://stackoverflow.com/q/6217156/53974 but they're less clear).

When you split a commit, each commit should not only make sense but at least still compile—so after making a commit, use `git stash` to "stash away" the uncommitted changes and try compiling the result. Trying to ensure that "Every commit compiles" helps when using `git bisect`.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: @Blaisorblade  I'm saying a boolean and static function inside pokefly. I don't think that was the case before. Also, this looping code will still have the race conditions that a boolean might. A flag inside Pokefly guarantees that it'll flow the service state pretty accurately.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Couple of options:

1. Move the exit from MainActivity to here, then both intents would trigger same path? Not ideal, but better than 2 exits() I think.

2. Cleanest, but not easiest to figure out: Figure out a good way to move screen.init() to Pokefly. I was too confused with the existing code that does a lot of intent passing and request codes to get the projection manager  (or whatever it was called) to start recording. I don't think it mandates the need of an Activity to be able to get that to work. It might just need Context (which both Service and Activity are). Also, the code was in a much worse state when I created ScreenGrabber -- so I didn't move it screen init to Pokefly.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Great. Thanks!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: As @Blaisorblade said, my "split into 4 commits" should not be taken to be an exact request you have to follow blindly. It's just seems too big to be one commit. Maybe you can only split it into 2 or 3 commits if you are to meet the other requirements @Blaisorblade mentioned (like having to compile and not be completely broken, etc).

I personally would do exactly what @Blaisorblade said, but if you are new to git, stash and "rebase -i" could be very disorienting. In which case, I'd recommend making a temp branch that's a copy of this. Then reset this branch to master and cherry-pick the commits one at a time from the temp branch. Once you cherry-pick a commit and it's the latest commit in your current branch, you can split it up by doing:

1. git reset HEAD~1 (this will remove the commit from log, but will leave the changes on files)
2. git diff (this should show that the above worked)
3. git add -p (this show pieces of code changes close to each other and asks you if you want to add it or not. Only add the ones you want to go into split commit 1)
4. git commit (fill up the test and commit it)
5. go to step 3 to finish split commit 2, 3, 4...

Hope this helps.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: Yeah, not a strong opinion (this isn't C after all), but if you don't mind do this for the new code.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
sarav: Now on a PC, I can try a longer comment.

If I'm not mistaken, the existing code uses flags in MainActivity that can easily get out of sync with Pokefly in case it's killed for memory, etc. With notification actions to stop Pokefly, it WILL get out of sync.

The commit above tries to solve this by looking at all the services running on the system. This is very expensive in terms of CPU and memory usage and doesn't really solve the problem completely.

What if user in the MainActivity screen, but uses notification to stop the service and then presses the button in the MainActivity to trigger this MainActivity code. If done fast enough (and threads get context switched out), it can cause the function to return a wrong value. The absolute correct code would need locking and synchronization I think or some kind of message queue to serialize all actions (Android might be doing this for us depending on how many threads it creates per task -- I don't have a confident answer).

Anyway, my suggestion was something like this in Pokefly. You don't have to use it and it's not significantly more race-condition proof than your implementation, but I think it'll be just as correct as the above implementation and less resource hungry.

```
public class Pokefly extends Service {

private static boolean running = false;

onCreate() {
  running = true; // Say running while you are starting.
  // do init stuff
}

onDestroy() {
  // do exit stuff
  running = false; // Say running as long as you haven't stopped.
}

public static boolean isRunning() {
  return running;
}
```

sarav:
Btw, rebase to master before you push again. I made some changes to Pokefly that might cause conflicts here.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: Ok @sarav!
I used that pattern before.
I'll include the static flag ASAP!

thearaks:
@sarav Already rebased successfully on my local repo ;)
Before pushing I'd like to include all the suggestions and comments from this thread, it will take some time because I'm a little busy lately. Sorry!

sarav:
Thanks for accepting the feed back and working on it. I'll wait for your update. I hope the git help from  @Blaisorblade or me helps.

Blaisorblade:
Replying to @sarav:
> @thearaks Not trying to discourage or be a downer, but thinking more about it, do we even need this feature now that we added #405 ? The only reason I can think of people wanting to stop/start GoIV is due to the high battery drain. 

FWIW: even with #405 GoIV has non-nil overhead — it uses the screen recording API (which seems far from zero-copy) and scans the screen when I touch it. I'll still stop GoIV before a gym fight—but switching to GoIV takes ages will suspend PoGo and can force it to be reloaded.

And I guess everybody who hasn't audited the app is justified in closing it before typing a password.

> Also, adding trainer lever increment/decrement [as notification action]

I'm pretty sure increment is much more useful than decrement since your level will never decrease while playing.

nahojjjen:
I think that even with the app only scanning the screen on clicks, being able to stop from notification is worthwhile. Both to make the user feel more in control, and to close it when you know you dont need to scan for a long while - for instance when i stop playing pogo, and start browsing reddit.

sarav:
Replying to @Blaisorblade 

I did agree later on that it make sense to have this. So, not sure why we are replying to my older points.

> FWIW: even with #405 GoIV has non-nil overhead — it uses the screen recording API (which seems far from zero-copy) 

Remember how you read the producer would be blocked if the buffers are full? I'm hoping that means no copying is being done unless we start fetching the image when the touch event happens. I agree it's still not nothing because people touch the screen often enough. Also, during gym fights, it's very unlikely GoIV will actually grab a screen because of the initial scan delay/throttling.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -485,6 +491,8 @@ private void hideInfoLayoutArcPointer() {
 
     @Override
     public void onDestroy() {
+        LocalBroadcastManager.getInstance(this).sendBroadcast(new Intent(ACTION_STOP));
+
```
sarav: Hmm... do you need isRunning() if you have to use Intents? I haven't fully gone through the code, but just pointing out for now.

If the receiver of the intent checks for Pokemon.isRunning() they'll get true. Is that the intended design? I hope this intent doesn't get MainActivity to try and stop this service but instead just updates the notification or something. Otherwise, it could try to stop the Pokefly service while it's already OnDestroying -- not sure if that's an issue in Android or not.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -68,11 +68,22 @@
...
     private final Point arcInit = new Point();
     private int arcRadius;
+    private final BroadcastReceiver pokeflyStateChanged = new BroadcastReceiver() {
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (Pokefly.ACTION_START.equals(intent.getAction())) {
+                updateLaunchButtonText(true);
+            } else if (Pokefly.ACTION_STOP.equals(intent.getAction())) {
+                updateLaunchButtonText(false);
+            }
```
sarav: Can't we just handle it with one Intent? And just pass true/false based on Pokefly.isRunning()?

app/src/main/java/com/kamron/pogoiv/Pokefly.java
thearaks: This intent is used to notify to MainActivity that the service is stopping so it need to update the UI (launch button). This happens when the user stops the service from the notification action while the GoIV app is running below the notification bar!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: Yes, we can. If we choose this option then the new Intent will need to be launched only after Pokefly has updated its _running_ variable value!

sarav: Good point. Not sure which is better.

sarav:
@thearaks and @Blaisorblade 

None of my comments are serious enough to warrant holding back this PR if we want to make a release ASAP.

And we can always tweak it later if we need to. But I didn't do a good review of the first commit (all the if checks). hoping someone else can if it's urgent.

Blaisorblade:
I should have tested this earlier—this looks rather cool, but I have two minor issues:

1. after pressing START, the button does some rapid change but comes back to START and enabled until the service actually starts. If you think the button press didn't work, and click START again, the service will stop right away. This can only be triggered if you don't use "Launch Pokemon Go", but I hope this can be addressed. (I'll give it a try).

2. what I would I'd have loved is for "stop" to keep the notification (but make it dismissable) and turn "stop" into "start". But this is already cool as-is and this missing feature should not prevent merging.

Also, I like this because I sometimes want to stop GoIV while running PoGo, and also restart it while running PoGo. But at least for Gym fights that's not needed, as @sarav says — you need a touch and then some time without touches before the screen is scanned.

I'm also reviewing the code meanwhile.

Blaisorblade:
I pushed a small cleanup and a potential fix for cosmetic issue 1 above. Comments on that @thearaks @sarav?

Thanks to @thearaks for enabling committers to push changes!

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -68,11 +68,30 @@
...
+        @Override
+        public void onReceive(Context context, Intent intent) {
+            if (Pokefly.ACTION_START.equals(intent.getAction())) {
+                updateLaunchButtonText(true);
+                pokeFlyStarting = false;
+            } else if (Pokefly.ACTION_STOP.equals(intent.getAction())) {
+                updateLaunchButtonText(false);
+                pokeFlyStarting = false;
+            }
+        }
```
sarav: 1. Combine this into one intent that checks "Pokefly.isRunning()"
2. Let's call that intent "UPDATE_BUTTON"
3. "Enable" the button again (see comment in startPokefly() )

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -274,6 +304,9 @@ public boolean onOptionsItemSelected(MenuItem item) {
     @Override
     protected void onResume() {
         super.onResume();
+
+        updateLaunchButtonText(Pokefly.isRunning() || pokeFlyStarting);
```
sarav: 1. Just send an "UPDATE_BUTTON" intent to ourselves instead of calling this function.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -288,19 +321,29 @@ private int getStatusBarHeight() {
      * Starts the PokeFly background service which contains overlay logic.
      */
     private void startPokeFly() {
-        ((Button) findViewById(R.id.start)).setText(R.string.main_stop);
+        pokeFlyStarting = true;
+        updateLaunchButtonText(true); //Should be subsumed by onResume later, which seems to be always called in
```
sarav: 4. Don't call updateLaunchButton() here.
5. Disable the button instead to avoid further touch events.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -465,43 +491,66 @@ private void hideInfoLayoutArcPointer() {
...
```
sarav: 1. Move this to the bottom after setting "runnin = false"

sarav:
@Blaisorblade, only after installing it do I see that it goes from "Start" -> "Accept Screen Capture" -> "Start" -> about a second of delay -> "Stop". I don't think anyone is going to quickly press "Start" again. By the time they realize that the it has gone back to "Start" and react, it'd have gone back to "Stop". So, I don't think this is a big issue. However, I'll still try to see if I can fix it.

Blaisorblade:
> I don't think anyone is going to quickly press "Start" again.

Thanks for fixing it in #438. FWIW, I did this instinctively as a user and only then rationalize it.

Issue #300, opened 2016-08-30
Make non-battery saver mode more efficient


sarav:
I've been searching online the past few days for ways to detect any touch on the screen (without caring/getting the info where exactly the touch was -- I'm sure this won't be possible due to security concerns) and there are various suggestions. I'm not which of them necessarily work. If anyone here knows the answer to this already or wants to write some test app to see if this can be done, it'll be good to hear the results.

**IF** we can know that a touch actually happened (I think someone else mentioned it here or on reddit) then we can have the screen scan for pokemon screen to happen only during touches. We'll still have to throttle touch based scanning since I'm sure people tap way faster than the every .75 second scan we are doing now. But that should be easy to do.

I'm hoping that if we don't acquire images from the mediaprojection and let it fill up the 2 image queue in the ImageReader, that the GPU or whatever will stop filling in the image (which I think is the most power hungry part). But maybe not -- because then "acquireLatestImage()" won't be latest anymore. In any case, even if the GPU/whatever is filling up the buffer with the screen content, at least not reading it and checking it every 0.75 seconds should help quite a bit too.

Even though I want to get screenshot/battery saver mode to work, the normal mode is a lot more convenient. So, I'd ideally like to reduce it's power consumption to a point that the benefits outweigh the need to use battery saver mode (we'll still leave it in).

Long story short, anyone willing to figure out how to detect screen touches? Well, even better would be to detect the Pokemon screen without doing any screenscan, but that doesn't look like it's feasible.

farkam135:
Ah interesting, What if the pokefly service displayed a full screen transparent overlay. If we put an onclick on it, shouldnt that act as a way to detect screen touches?

sarav:
AFAIK, I don't think you can let the touch go thru. But I could be wrong.

sarav:
I mean, AFAIK, you can't detect the touch and let it go thru. It might be either or (feels like a security issue otherwise). But maybe accessibility permissions does?

rpgrca:
You mean, something similar to the AMOLED pixel off applications in the Play Store, which display an overlay on top of the screen? I'm guessing it detects clicks and passes them through, but my experience while playing Ingress wasn't that good, the game became clunky since some long pressed were cut, some slides were not detected, etc.

sarav:
I don't think the pixel off or filter applications process any touch. Apparently views can mark themselves as not focusable and won't get any events. 

I don't think lag should be an issue if implemented correctly.

sarav:
[...]

PR #405, opened 2016-09-13
Power optimization: Scan screen only when user touches it

sarav:
This addresses #300 

The current implementation scans the screen every 750 milliseconds. This
is pretty expensive in terms of power consumption.

Since the screen mode (map, gym, pokemon, etc) doesn't change unless the
user actually touches the screen, we just need to scan the screen around
the time the screen is touched.

Also, since the user touches the screen a lot during a gym battle, add
some throttling to scan the screen only after the last touch if multiple
touches happen in quick succession (within 750 milliseconds).

This should significantly improve the power.

Since there are some animations which take time when the user opens a
pokemon switches pokemon, scan a couple of times before stopping.

**Caveat:** The touch events only come when the user touches the screen and not when the lift the finger. So, if they do a long drag (2.5 seconds), the scanning would finish before they lift the finger. But this hasn't really happened to me when I use the game UI normally. Also, in that case, they just need to touch the screen once to trigger the scan again. So, it's not that bad for the enormous power savings.

**Result:** Not sure if it's placebo, but I feel like the device is running cooler. Also, this is the first time I was able to start GoIV and then go back to the game that was already running without having to restart it.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -429,14 +429,16 @@ public class Pokefly extends Service {
      * Scans the device screen to check area[0] for the white and area[1] for the transfer button.
      * If both exist then the user is on the pokemon screen.
      */
-    private void scanPokemonScreen() {
+    private boolean scanPokemonScreen() {
         @ColorInt int[] pixels = screen.grabPixels(area);
 
         if (pixels != null) {
             boolean shouldShow =
                     pixels[0] == Color.rgb(250, 250, 250) && pixels[1] == Color.rgb(28, 135, 150);
             setIVButtonDisplay(shouldShow);
+            return shouldShow;
         }
+        return false;
     }
 
     private boolean infoLayoutArcPointerVisible = false;
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -65,8 +65,6 @@ import com.kamron.pogoiv.widgets.PokemonSpinnerAdapter;
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Map;
-import java.util.Timer;
-import java.util.TimerTask;
 
 import butterknife.BindView;
 import butterknife.ButterKnife;
@@ -115,7 +113,6 @@ public class Pokefly extends Service {
     private ScreenShotHelper screenShotHelper;
     private OcrHelper ocr;
 
-    private Timer timer;
     private Point[] area = new Point[2];
 
     private boolean infoShownSent = false;
@@ -125,6 +122,14 @@ public class Pokefly extends Service {
     private ImageView ivButton;
     private ImageView arcPointer;
     private LinearLayout infoLayout;
+
+    private LinearLayout touchView;
+    private WindowManager.LayoutParams touchViewParams;
+    private Handler screenScanHandler;
+    private Runnable screenScanRunnable;
+    private static final int SCREEN_SCAN_DELAY_MS = 1000;
+    private static final int SCREEN_SCAN_RETRIES = 3;
+    private int screenScanRetries;
 
     private PokeInfoCalculator pokeInfoCalculator;
 
@@ -355,7 +360,7 @@ public class Pokefly extends Service {
             /* Assumes MainActivity initialized ScreenGrabber before starting this service. */
             if (!batterySaver) {
                 screen = ScreenGrabber.getInstance();
-                startPeriodicScreenScan();
+                watchScreen();
             } else {
                 screenShotHelper = ScreenShotHelper.start(Pokefly.this);
             }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -364,26 +368,59 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
             public void run() {
-                handler.post(new Runnable() {
-                    public void run() {
-                        scanPokemonScreen();
-                    }
-                });
+                if (screenScanRetries > 0) {
+                    scanPokemonScreen();
+                    screenScanRetries--;
+                    screenScanHandler.postDelayed(screenScanRunnable, 500);
```
Blaisorblade: I hadn't realized we could have used `postDelayed` instead of `TimerTask` *all this time* 😢

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -364,26 +368,59 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
+                        | WindowManager.LayoutParams.FLAG_WATCH_OUTSIDE_TOUCH,
+                PixelFormat.TRANSLUCENT);
+        touchViewParams.gravity = Gravity.LEFT | Gravity.TOP;
+
+        touchView.setOnTouchListener(new View.OnTouchListener() {
+            @Override
+            public boolean onTouch(View v, MotionEvent event) {
+                if (event.getActionMasked() == MotionEvent.ACTION_OUTSIDE) {
+                    screenScanHandler.removeCallbacks(screenScanRunnable);
+                    screenScanHandler.postDelayed(screenScanRunnable, 750);
```
Blaisorblade: Could use `screenScreenDelay` rather than `750`.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -126,6 +123,13 @@
     private ImageView arcPointer;
     private LinearLayout infoLayout;
 
+    private LinearLayout touchView;
+    private WindowManager.LayoutParams touchViewParams;
+    private Handler screenScanHandler;
+    private Runnable screenScanRunnable;
+    private final int screenScreenDelay = 750;
```
Blaisorblade: I'd make it static and add the unit of measure, so `screenScreenDelayMs`.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: This is the 500 I'd replace by `SCREEN_SCAN_RETRY_DELAY_MS`.

Blaisorblade:
This looks good. I've added a couple of comments. And overall I might prefer having this in a separate small class since (1) we can (2) this would encapsulate a small state machine (3) Pokefly is too big, so this small state machine is lost amid a bunch of others.

But I won't insist on that too much since it is in part a matter of preference.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: This was the whole point of adding this variable, but I forgot to use it. Did you test it? How well did it work for you?

Blaisorblade: Yeah I guessed, `screenScreen` was also telling ;-). Testing now...

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -126,6 +123,13 @@
     private ImageView arcPointer;
     private LinearLayout infoLayout;
 
+    private LinearLayout touchView;
+    private WindowManager.LayoutParams touchViewParams;
+    private Handler screenScanHandler;
+    private Runnable screenScanRunnable;
```
Blaisorblade: The code would be easier to navigate if this were final or we used a non-anonymous subclass.

sarav: Why final? I want to release them when the service isn't running or not allocate them when we are using battery saver mode.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: Working late at night doesn't help :)

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: This was supposed to be 750 ms too. So, just used the same final variable.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
sarav: I'm choosing to keep it this way for the above reason. Let me know if you have a strong opinion/reason for changing it.

sarav:
@Blaisorblade regarding your comment about moving this code to a different class, that'll unfortunately force us to add an intent to take the screenshot. I agree in principle, but I think we should decide this after seeing how Pokefly looks after we move the dialog box UI code out -- there's SO MUCH of it!

I almost want to say we shouldn't allow any UI changes until these are addressed.

Blaisorblade:
> I almost want to say we shouldn't allow any UI changes until these are addressed.

That was the plan, but last we tried there was nobody doing any work (because as you told me you were on vacation). If you have the time to try we can schedule that, keeping into account open PRs.

Blaisorblade:
I've pushed a few extra commits on `optimize-screen-grab-mod` (haven't removed the logging yet), that's enough to fix it for me. I hope that works well enough.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: I only meant final for the `screenScanRunnable`—it's just one instance. But it's a minor thing either way.

Blaisorblade:
BTW we should maybe use this: https://github.com/blog/2247-improving-collaboration-with-forks

sarav:
I took a quick look. Yeah, I had already implemented and threw away the code to stop scan it the button status changed. Because there are cases like when you swipe between Pokemon, where if you do it slowly enough, the button can disappear and appear again. 

So, I think all you need is an increase in retry count or the delay (looks like it just takes a while for the UI to settle for you).

Are you okay with just increasing the retry by 1? It's still way more efficient that what we do today.

sarav:
I'm more curious about fixing the null pointer exception even if it didn't cause an issue for you.

Also understood you weren't suggestion a new activity, but I think we'd still have to use intents nonetheless to avoid abstraction breaking hacks. I'll look into it later as a separate refactoring improvement. As is, I don't think I made it any worse.

sarav:
I need to fix some typos, but on mobile now. Will fix later in about 5 hrs.

Blaisorblade:
> Are you okay with just increasing the retry by 1? It's still way more efficient that what we do today.

With 3 it still had issues sometimes (when loading the first pokemon), using 4 (or a bigger delay) should be better.
I'm mainly afraid of discoverability and I'm sure some users will think it broke if it doesn't work. Even if we say it first thing in the changelog and announcement.

> Also understood you weren't suggestion a new activity, but I think we'd still have to use intents nonetheless to avoid abstraction breaking hacks. I'll look into it later as a separate refactoring improvement. As is, I don't think I made it any worse.

Can be later. I'd give the class a pointer to Pokefly. Can try myself.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -364,26 +368,59 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
+
+        touchView = new LinearLayout(this);
+        touchViewParams = new WindowManager.LayoutParams(
+                1,
+                1,
+                WindowManager.LayoutParams.TYPE_PHONE,
+                WindowManager.LayoutParams.FLAG_NOT_FOCUSABLE
+                        | WindowManager.LayoutParams.FLAG_NOT_TOUCH_MODAL
+                        | WindowManager.LayoutParams.FLAG_WATCH_OUTSIDE_TOUCH,
+                PixelFormat.TRANSLUCENT);
```
Blaisorblade: Can we get by with TRANSPARENT? We need to be fully transparent anyway...
```
    /** System chooses a format that supports translucency (many alpha bits) */
    public static final int TRANSLUCENT = -3;

    /**
     * System chooses a format that supports transparency
     * (at least 1 alpha bit)
     */
    public static final int TRANSPARENT = -2;
```

sarav:
Ok, I'll increase retires and change to transparent and test it. And I'll push it.

I might later try increasing the delay between retries (and reducing the no. of retries left) if the button state changes.

sarav:
**Warning:** Rambling comment.

> Can be later. I'd give the class a pointer to Pokefly.

Giving a pointer to Pokefly and making the methods public is what I meant by "abstraction breaking hacks". If it needs to be that tightly coupled, then I'm not sure there's much benefit to moving it out of Pokefly. Pure logic, like OCR, scanning the screen, etc I can see why it's good to move it out. Also, the main point of Pokefly is to stay in background, check for the right conditions and throw up the IV button -- so this really should stay in Pokefly.

If anything (as in move anything other than UI code out of Pokefly), we could/should move the screen identifying code into a pure logic class that'll identify the screen we are on (Map, pokemon, gym, etc). Say, ScreenIdentifier.java. We'd initialize it with the screen dimensions, a screengrabber object (to getPixels from), etc and then later do stuff like screenIdentifier.isPokeScreen().

If I were to ever move the screen watching code out of Pokefly in a clean way. I'd want to move this into a separate class and make it behave like ScreenShotHelper. Similar to screenshot helper looking at some events to go find the pokemon bitmap (screenshot) to scan, this new class would look at the touch events, check if the screen has a pokemon and if so, send the process-bitmap intent. But at that point, it'll be more of the case of not showing an IV button and directly calculating the IV and displaying a floating IV result that someone can touch to expand. That way, it'd work just as well for screenshot or automated screen scans. At that point, there's really nothing to do in Pokefly (since we won't be showing the IV button either).

sarav:
> I took a quick look. Yeah, I had already implemented and threw away the code to stop scan it the button status changed. Because there are cases like when you swipe between Pokemon, where if you do it slowly enough, the button can disappear and appear again.

@Blaisorblade, took another look at your code. My statement about shouldn't be an issue. The code I added was to stop scanning if the button went from "shown" to "not shown" or "not shown" to "shown". But you are stopping only when showing the button. Since there's no way make the IV button appear and then exit from the Pokemon screen in one touch/swipe event, I think your change should be fine. I pulled in your commits into my branch, cleaned up the logs and trying them out. If it works fine, I'll update this PR to include your changes too.

sarav:
@Blaisorblade Travis seems to have thrown up in his mouth. Can you revive him?

Blaisorblade:
Merged #410, confirmed that's the issue. To fix Travis please rebase on latest master.

PR #410, opened 2016-09-13
Upgrade dependencies

Blaisorblade:
This is needed to build with version 37 of "Android support repository", and Travis already seems to default to that version at least according to this log:
https://travis-ci.org/farkam135/GoIV/builds/159657295

Also @Eremiell got stuck onto this when trying to install.

Here is the install of the offending version:
https://travis-ci.org/farkam135/GoIV/builds/159657295#L927-L932

This might require other devs to update their tools. Conversely, I've updated and need this to build. This PR is to give a heads up to all.

BTW: tooling is being stupid here, why can't the older library be downloaded from some maven repositories? Bah.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -127,8 +127,8 @@ android {
  * annotation processing)
  */
 dependencies {
-    compile 'com.android.support:appcompat-v7:24.2.0'
-    compile 'com.android.support:recyclerview-v7:24.2.0'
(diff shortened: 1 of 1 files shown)


sarav:
Yeah, I had to update to these versions on my end anyway for it to compile. Please merge this. I think this is why #405 is failing in Travis.

PR #415, opened 2016-09-15
Overlay gravity only top if on input screen

### LINKED ISSUE #406: Wrong gravity for dialog
To reproduce:
1. press IV button
2. expand appraisal (this will move the floating input window upward)
3. press Check IV (the window will move back down)
4. press button to inspect combinations (window stays down)
5. press back, and *the window moves incorrectly up* instead of staying down.

This sounds like the kind of bug you prevent by just separating the windows.

nahojjjen:
This commit makes the moveOverlayUpOrDownToMatchAppraisalBox method
always move the overlay down if the user is not on the input screen, as
it's not possible to end up in a situation where the appraisal box is
both intended to be expanded and not exist.

Fixes #406

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -842,7 +842,8 @@ public class Pokefly extends Service {
             return;
         }
 
-        moveOverlay(appraisalBox.getVisibility() == View.VISIBLE);
+        //move up if on input screen & appraisal box is open, else move down
+        moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE);
     }
 
     private void adjustArcPointerBar(double estimatedPokemonLevel) {


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -842,7 +842,14 @@ private void moveOverlayUpOrDownToMatchAppraisalBox() {
...
 
-        moveOverlay(appraisalBox.getVisibility() == View.VISIBLE);
+        if (inputBox.getVisibility() == View.VISIBLE) {
+            boolean appraisalExpanded = appraisalBox.getVisibility() == View.VISIBLE;
+            moveOverlay(appraisalExpanded);
+        } else { //always move overlay down if not on input screen. Appraisal box cant be open if user is not in
+            // input screen
+            moveOverlay(false);
+        }
+
```
Blaisorblade: Should this be `moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE);` (with the same comment)?

Blaisorblade:
Thanks for taking a look!

app/src/main/java/com/kamron/pogoiv/Pokefly.java
nahojjjen: > moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE);

That would check two booleans:

1. the user is viewing the input box
2. the user has expanded the appraisal box.

Checking both of those is not necessary, as it's not possible for the user to be viewing the appraisal box if the input box is not visible. This is because the appraisal box is **within** the input box, so if the input box is invisible, the appraisal box is always invisible.

Blaisorblade: Nope, because `&&` is short-circuiting — the second would be tested only if the first is true. It's just more concise.

nahojjjen: I didn't mean how the program ran the check, I thought you meant that I should add moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE); to replace line 850, which wouldn't help much, which I explained by referring to that the possibility would never be possible. 

Do you mean that the if-else block should be replaced by 
moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE);?

In that case, the answer is yes, I think it could be.

Blaisorblade: Yep I talked about the whole if-else, sorry for the ambiguity.

nahojjjen:
Amended commit to change to your suggestion.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -842,7 +842,8 @@ private void moveOverlayUpOrDownToMatchAppraisalBox() {
             return;
         }
 
-        moveOverlay(appraisalBox.getVisibility() == View.VISIBLE);
+        //move up if on input screen & appraisal box is open, else move down
+        moveOverlay(inputBox.getVisibility() == View.VISIBLE && appraisalBox.getVisibility() == View.VISIBLE);
```
Blaisorblade: LGTM.

PR #416, opened 2016-09-15
Automatically set estimate pokemon to evolution of scanned

### LINKED ISSUE #223: Additional information starts from "maxed up" pokemons, & Q about its range
This probably requires more feedback: I'm L27, so when I choose advanced information in any pokemon, it would show me the CP it would have if I pumped it all the way up to L28.5, even for a L20 bulbasaur. Wouldn't it be more useful to automatically show the data for the immediately superior evolution at the current pokemon's level instead?

This is based in a) there's no way I'd spend 49 candies and 60500 stardust pushing such bulbasaur til L28.5 without knowing its moveset will be worth unless it's already at its maximum evolution; and b) I'm more interested in knowing if I evolved this bulbasaur, will he get a higher CP amount than the ivysaur I already got to spend the candies?

Also, if I choose "ivysaur", the amount of candies should be updated with the 25 candies needed to go from bulbasaur to ivysaur (and if I were to choose venasaur, it should also contain the 100 candies needed to jump from ivysaur to venasaur). Right now, it gets stuck at 0 candies.

In other words, I think clicking on Additional information for this bulbasaur should select Ivysaur, L20 (current level) and 25 candies (cost to upgrade from bulbasaur to ivysaur).

nahojjjen:
When scanning, in the evolution and powerup estimation box, the evolved
form of the pokemon will now automatically be selected by default, as
that is the most likely selection the user wants.

For example, if scanning a caterpie, metapod will be automatically
selected. If the user scans a pokemon that has no evolution, the pokemon
itself will be selected.

This change is related to #223
EDIT by @Blaisorblade: This change indeed fixes #223.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1186,37 +1186,56 @@ public class Pokefly extends Service {
      * pokemon evolution and level set by the user.
      */
     private void populateAdvancedInformation(IVScanResult ivScanResult) {
-        double goalLevel = seekbarProgressToLevel(expandedLevelSeekbar.getProgress());
-        int intSelectedPokemon =
-                extendedEvolutionSpinner.getSelectedItemPosition(); //which pokemon is selected in the spinner
+        double selectedLevel = seekbarProgressToLevel(expandedLevelSeekbar.getProgress());
         ArrayList<Pokemon> evolutionLine = pokeInfoCalculator.getEvolutionLine(ivScanResult.pokemon);
+        Pokemon scannedPokemon = ivScanResult.pokemon;
 
-        Pokemon selectedPokemon;
-        if (intSelectedPokemon == -1) {
-            selectedPokemon = ivScanResult.pokemon;//if initialising list, act as if scanned pokemon is marked
-            for (int i = 0; i < evolutionLine.size(); i++) {
-                if (evolutionLine.get(i).number == selectedPokemon.number) {
-                    intSelectedPokemon = i;
-                    extendedEvolutionSpinner.setSelection(intSelectedPokemon);
-                    break;
-                }
-            }
-        } else {
-            if (evolutionLine.size() > intSelectedPokemon) {
-                selectedPokemon = evolutionLine.get(intSelectedPokemon);
-            } else {
-                selectedPokemon = evolutionLine.get(0);
-            }
+        initSpinnerIfNecessary(scannedPokemon);
+        //Since initSpinner might change the selected pokemon when it's setting the default, setting selectedPokemon
+        // is done after initIfNecessary.
+        Pokemon selectedPokemon = evolutionLine.get(extendedEvolutionSpinner.getSelectedItemPosition());
 
-        }
+        //Below block sets the text boxes in the powerup and evolution estimate box
+        setEstimateCpTextBox(ivScanResult, selectedLevel, selectedPokemon);
+        setEstimateCostTextboxes(ivScanResult, selectedLevel, selectedPokemon);
+        exResLevel.setText(String.valueOf(selectedLevel));
+        setEstimateLevelTextColor(selectedLevel);
 
-        extendedEvolutionSpinner.setEnabled(extendedEvolutionSpinner.getCount() > 1);
+    }
 
+    /**
+     * Sets the candy cost and stardust cost textfields in the powerup and evolution estimate box. The textviews are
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -1186,24 +1186,34 @@ private int levelToSeekbarProgress(double level) {
...
+            int selectIndex = 0;
+            for (Pokemon evoPoke : evolutionLine) { //find the index of the pokemon in evolution line
+                if (evoPoke.number == selectedPokemon.number) {
                     break;
                 }
+                selectIndex++;
             }
+            if (hasEvolution) { //select the next pokemon in the spinner if evolution exists
+                selectIndex++;
+            }
```
Blaisorblade: So IIUC these three lines are the actual bugfix, and everything else is just a refactoring, right? Next time, it'd be nice to separate those changes, it's really hard to review them together.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -1186,24 +1186,34 @@ private int levelToSeekbarProgress(double level) {
...
+            extendedEvolutionSpinner.setSelection(selectIndex);
+
         } else {
-            if (evolutionLine.size() > intSelectedPokemon) {
-                selectedPokemon = evolutionLine.get(intSelectedPokemon);
+            //spinner has been previously initialized, dont change selection unless user managed to bug around
+            // interface and tries to set the selection outside the possible evolution line.
```
Blaisorblade: I'd avoid having "bug around interface"—somebody like me seeing the comment would consider removing the check. But IIRC this happens easily, that is whenever you scan some pokemon A and then another pokemon B, if B's evolution line is smaller than A's.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
nahojjjen: Yes, you could implement this feature by adding the boolean hasEvolution and then doing intSelectedPokemon++ instead of moving a lot of stuff around like I did.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
nahojjjen: This was what was confusing me, I previously did this code, but now that I re-read it I didn't quite understand why I'd done as I did. Ive amended the commit and removed the else clause, that can never occur if the spinner list is updated correctly before the selection is made. (Which makes sense, why did I change the selection before changing the list contents before?)

nahojjjen:
Ive included a refactor in this change now, Im sorry this was not what you wanted, but the method was difficult to read and understand, so I think it was due time to break it down. I will read through this again later when I'm not as tired.

skaforey:
Looking forward to this one

nahojjjen:
I have redone the commit now.

Blaisorblade:
I agree with you that the current code is rather messy.

But if a commit is already too big and hard to understand, adding a refactoring to the same commit makes things worse. I've been looking at this code for a while and I still haven't managed to convince myself that either is correct.

I'm closing the PR and will try splitting it properly, reusing your code. I hope by doing that I'll also figure out whether the changes are correct, whether the existing code has any other bugs to fix, and more.

nahojjjen:
> But if a commit is already too big and hard to understand, 

The commit was not too big, it was messy - those are not the same thing. I agree that in most cases that you should keep refactorings and new features as separate commits, but in situations like this I firmly believe that when redoing a part of the code, changing something minor like this is completely harmless.

>  I've been looking at this code for a while and I still haven't managed to convince myself that either is correct.

After the refactor, the code is very straightforward. I think you're taking correctness checking to an absurd level.

> I'm closing the PR and will try splitting it properly, reusing your code. I hope by doing that I'll also figure out whether the changes are correct, whether the existing code has any other bugs to fix, and more.

I don't see a point in doing that.

Overall I don't agree with your decision to close this PR with the motivation you've written out.

Blaisorblade:
> After the refactor, the code is very straightforward. I think you're taking correctness checking to an absurd level.

Yeah, the new code was straightforward, but that doesn't tell me whether the new code forgets something. That's a general principle, and we've already had experience with that.

And just reading the new code doesn't tell me I need to check whether bfddc6d09e1609b2d10c84ef69d77d1deb92c595 makes sense.

It seems correct (like the rest) because `extendedEvolutionSpinner.getSelectedItemPosition()` is only `-1` when we actually change the pokemon list—but I haven't been able to confirm that from either the docs or reading the source.

PR #418, opened 2016-09-15
Adds bottom margin to manual pokemon select

### LINKED ISSUE #345: Hidden input field when searching for pokemon
We merged #308 with a small bug to fix (https://github.com/farkam135/GoIV/pull/308#issuecomment-2439511000):

When typing in the field, the input dialog is covered by the keyboard (what we had forever for CP and HP and then fixed); this can be worked around by enabling appraisal.

nahojjjen:
Moves around code lines as suggested by reformat code in android studio,
and adds a margin to the text input so that it's not covered by the
keyboard when you type.

The only new thing in this commit is really the margin.

Visually this is not great nor optimal, but it makes things easier.

Fixes #345 - the fix is kinda hacky, so a better fix could be implemented, but it's better than nothing imo.

### DIFF
--- a/app/src/main/res/layout/dialog_input_poke_picker.xml
+++ b/app/src/main/res/layout/dialog_input_poke_picker.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
-              android:orientation="horizontal"
               android:layout_width="match_parent"
-              android:layout_height="match_parent">
+              android:layout_height="match_parent"
+              android:orientation="horizontal">
 
 
     <Spinner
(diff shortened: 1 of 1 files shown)


nahojjjen:
unnamed

Picture of change

Blaisorblade:
@oskomonzon @DJCrapsody I'd like you to review this change ;-)

oskomonzon:
I know the soft keyboard is an issue but I'm not sure about this approach because there's no way to be sure it will fit most displays.
In one of my devices the inputs never get covered by the keyboard, but they do on smaller devices. My father haves a device similar to mine but due to his age he uses a way taller keyboard, same thing: it get's covered...

I'm still not sure why the windowSoftInputMode adjustPan setting is not working fine BUT I SUSPECT it's related to the use of a scrollview in that instance. I'd like to give it a try without those before merging this change.

oskomonzon:
Nailed it:
screenshot_2016-09-17-18-54-19

After looking for a while I spotted the issue: Remove "WindowManager.LayoutParams.FLAG_LAYOUT_IN_SCREEN | " from layoutParams declaration instead \:D/

PD: Scrollview can stay.

oskomonzon:
I made a new PR as requested by @Blaisorblade 
https://github.com/farkam135/GoIV/pull/422

Blaisorblade:
Closing after alternative fix. Thanks @nahojjjen anyway!

PR #421, opened 2016-09-17
Fix crash when no external storage is available

Blaisorblade:
Close http://crashes.to/s/c99c9d8a9c5. But this might lead to duplicated
assets if the external folder becomes unavailable. Ideas?

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1442,7 +1442,11 @@ public class Pokefly extends Service {
     }
 
     private void initOcr() {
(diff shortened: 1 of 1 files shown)


Blaisorblade:
This is probably related to http://crashes.to/s/4f080aecfd7.

DJCrapsody:
Seems like we need [this workaround](http://stackoverflow.com/questions/32774226/getexternalfilesdirnull-returns-null-even-with-read-external-storage-permissio)

Blaisorblade:
That's for when `WRITE_EXTERNAL_STORAGE` is not requested, so it doesn't necessarily apply. Maybe we should still add it as a fallback/alternative.

Also, none of those answers actually explains what's going on, though one has a workaround.

PR #422, opened 2016-09-17
Ensure soft keyboard does not obstruct layout

oskomonzon:
Following the discussion at PR #418
Should work as intended now.


sarav:
I was very confused on what this was. The right word is "obstructing". "Obfuscating" means something like "making it blurry/vague" (not the right definition). Obfuscating is what "Proguard" does with the function and variable names in the complied bytecode.

PR #425, opened 2016-09-18
Add PoGo black screen issue to FAQ section

### LINKED ISSUE #424: Black app screen in PoGo after starting GoIV
After starting the GoIV service, PoGo is completely black and unresponsive when switched back to, and must be restarted to use. I've tried to figure out a consistent way to make this happen, but I can't seem to find one. I think it only seems to occur when PoGo has been open for awhile, and it doesn't seem to matter whether it's manually or automatically switched to after starting the GoIV service. I'd initially turned off the auto-switch, but I just had the black app screen again, so I don't think it's related to that.

voltaek:
According to the answer I got in issue #424, this is a common question, so it should go in the FAQ for visibility.

Fix #424


DJCrapsody:
LGTM

PR #429, opened 2016-09-18
Register button handler after completing setup

### LINKED ISSUE #183: Possible race condition at startup
I have just seen GoIV crash at startup when pressing Start/Stop fast enough.

While reviewing other code, `MainActivity.onCreate` sets up a handler for the launch button *before* the rest of the setup is finished. Should probably be avoided (by changing order) to make sure this does not cause trouble.

Blaisorblade:
Fix #183.


nahojjjen:
The code change looks reasonable, and it seems to work on my phone. It's difficult to test the difference it makes in reality though.. But testing race conditions usually is :)

Blaisorblade:
I'm not even sure the race condition is surely there—but with the new code I'm sure this isn't a problem, and the change is small, so I preferred to fix this.

PR #430, opened 2016-09-18
Miscellanous code cleanups/fixes

Blaisorblade:
This PR is a grab bag of changes for review—probably shouldn't be in a single PR. Most should be obvious, some aren't.
This happens to include #421 for now, but that should be kept separate.

To split out:
- commits dealing with `ssFile`.
- deletions of dead code could be combined.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -18,7 +18,6 @@ import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
 import android.preference.PreferenceManager;
-import android.provider.MediaStore;
 import android.provider.Settings;
 import android.support.annotation.NonNull;
 import android.support.v4.app.ActivityCompat;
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -402,21 +401,4 @@ private void startScreenService() {
...
-            cursor.moveToFirst();
-            return cursor.getString(column_index);
-        } finally {
-            if (cursor != null) {
-                cursor.close();
-            }
-        }
-    }
-
 }
```
sarav: Oh thanks! I thought I deleted it when I moved all the stuff to ScreenshotHelper.

app/src/main/java/com/kamron/pogoiv/ScreenShotHelper.java
```diff
@@ -24,7 +24,7 @@
     private ContentObserver mediaObserver;
     private ContentResolver contentResolver;
 
-    private String getRealPathFromUri(Context context, Uri contentUri) {
```
sarav: Looks like I missed a couple of things I meant to delete. :-/

nahojjjen:
I have not tested if the bugs you're referring to are fixed, but I'll just say that I tested the changes, and it seems to be working on my device, so nothing obvious has been broken. So lgtm. 👍

Blaisorblade:
I've added the misc fixes on master directly, the rest is in #421 and #434. Closing this one.

PR #408, opened 2016-09-13
Rewrite of the pokemon identification logic

nahojjjen:
This rewrite adds the ability to identify pokemon by evolution cost + candy name. The full chain of logic used to determine the pokemon in this implementation is:
1. check if the nickname perfectly matches a pokemon
2. check if closest candyname + evolution cost perfectly matches a pokemon
3. check if there's a cached result for the scanned pokemon name
4. get the pokemon with the closest name within the evolution line guessed from the candy

Known issue: Does not handle multiple evolutions such as eeveelutions

PR #431, opened 2016-09-18
Rewrite pokemon identification logic using candy evolution cost, V2

sarav:
@nahojjjen Here's your cleaned up PR #408 

Do you really think this clean up isn't worth the time? It's so much more simpler to read and review.
3 commits instead of the 18 (according to github) you had!

I did the following:

- Rebased everything to master as of now.
- Fixed the conflicts in the commits that had the conflict instead of fixing it in a separate merge commit
- I think you did some file renames and unrenames to avoid some conflicts. Dropped those.
- Squashed you formatting/checkstyle fixes into the commits that added them in the first place.

In the end we should probably squash the last 2 commits that update PokemonNameCorrector.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -144,6 +144,35 @@ public class OcrHelper {
     }
 
     /**
+     * Get the evolution cost for a pokemon, example, weedle : 12
+     * If there was no detected upgrade cost, returns -1
+     *
+     * @param pokemonImage The image of the full pokemon screen
+     * @return the evolution cost, or -1 on scan failure
+     */
+    public int getPokemonEvolutionCostFromImg(Bitmap pokemonImage) {
+        Bitmap evolutionCostImage =
+                Bitmap.createBitmap(pokemonImage, (int) (widthPixels * 0.625), (int) (heightPixels * 0.86),
+                        (int) (widthPixels * 0.2), (int) (heightPixels * 0.05));
+        String hash = "candyCost" + hashBitmap(evolutionCostImage);
+        String stringCacheEvoCandyCost = ocrCache.get(hash);
+        int result = -1;
+        if (stringCacheEvoCandyCost == null) { //if no cached result
+            //the dark color used for text in pogo is 76,112,114
+            evolutionCostImage = replaceColors(evolutionCostImage, 76, 112, 114, Color.WHITE, 40, false);
+            tesseract.setImage(evolutionCostImage);
+            stringCacheEvoCandyCost = tesseract.getUTF8Text().replace("S", "5").replace("s", "5");
+            try {
+                result = Integer.parseInt(stringCacheEvoCandyCost);
+            } catch (NumberFormatException e) {
+                stringCacheEvoCandyCost = "-1";
+            }
+            ocrCache.put(hash, stringCacheEvoCandyCost);
+        }
+        return result;
+    }
+
+    /**
      * Get the hashcode for a bitmap.
      */
     private String hashBitmap(Bitmap bmp) {
@@ -348,7 +377,8 @@ public class OcrHelper {
         String candyName = getCandyNameFromImg(pokemonImage);
         Optional<Integer> pokemonHP = getPokemonHPFromImg(pokemonImage);
         Optional<Integer> pokemonCP = getPokemonCPFromImg(pokemonImage);
+        int pokemonUpgradeCost = getPokemonEvolutionCostFromImg(pokemonImage);
 
-        return new ScanResult(estimatedPokemonLevel, pokemonName, candyName, pokemonHP, pokemonCP);
+        return new ScanResult(estimatedPokemonLevel, pokemonName, candyName, pokemonHP, pokemonCP, pokemonUpgradeCost);
     }
 }
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -119,9 +119,9 @@ public class OcrHelper {
                 allpixels[i] = replaceColor;
             }
         }
-
-        myBitmap.setPixels(allpixels, 0, myBitmap.getWidth(), 0, 0, myBitmap.getWidth(), myBitmap.getHeight());
-        return myBitmap;
+        Bitmap clone = Bitmap.createScaledBitmap(myBitmap, myBitmap.getWidth(), myBitmap.getHeight(), false);
+        clone.setPixels(allpixels, 0, myBitmap.getWidth(), 0, 0, myBitmap.getWidth(), myBitmap.getHeight());
+        return clone;
     }
 
     /**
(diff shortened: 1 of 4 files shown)


nahojjjen:
> Do you really think this clean up isn't worth the time? It's so much more simpler to read and review.

I think it's worth it in bigger projects / professional situations, but I think that most of the time for a project like this, it saves time to be a little sloppy. Squash merging my commit would have made it into a single commit, so history would not have been very messy. This does however help with code review.

There are some checkstyle errors in this pr, do you want me to fix them with a new commit, or will you amend this pr?

sarav:
I think we'll need to review this change. That's why I wanted to clean up. Also, if you are confident about the candy cost reading code, just merge that code and we can review and merge the Pokemon name corrector patch after (it looks more complicated to review).

Just make a copy of this branch and do your thing and push a PR. I'll delete this one. As far as I remember, you still had some issues to fix?

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -144,6 +144,35 @@ private double getPokemonLevelFromImg(Bitmap pokemonImage, int trainerLevel) {
...
+            evolutionCostImage = replaceColors(evolutionCostImage, 76, 112, 114, Color.WHITE, 40, false);
+            tesseract.setImage(evolutionCostImage);
+            stringCacheEvoCandyCost = tesseract.getUTF8Text().replace("S", "5").replace("s", "5");
+            try {
+                result = Integer.parseInt(stringCacheEvoCandyCost);
+            } catch (NumberFormatException e) {
+                stringCacheEvoCandyCost = "-1";
+            }
+            ocrCache.put(hash, stringCacheEvoCandyCost);
+        }
```
Blaisorblade: This code will always return `-1` if the string is found in cache.

sarav: Good catch. So, if things were working for @nahojjjen -- does it mean it was never matching cache?!

sarav: Looks like me might have fixed it in a later commit.

nahojjjen: Sarav are you or me changing this?

Blaisorblade: Turns out this code is rewritten in a later commit.

sarav: I think you fixed it later.

sarav: Wait, maybe it won't. In any case, you fixed it later for this one. I'll wait and see what other issues are reported. I should be able to fix most. But I'll do it in batches of 24hrs of reports.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -148,28 +148,77 @@ private double getPokemonLevelFromImg(Bitmap pokemonImage, int trainerLevel) {
...
+            evolutionCostImage = evolutionCostImageCanAfford;
+        }
+
+        //If not cached or fully evolved, ocr text
+        int result;
+        tesseract.setImage(evolutionCostImage);
+        String ocrResult = fixOcrLettersToNums(tesseract.getUTF8Text());
+        try {
+            result = Integer.parseInt(ocrResult);
+            ocrCache.put(hash, ocrResult);
```
sarav: I'm thinking of rearranging this code so that we'll always store in OCR independent of it's parseable or not. There's no reason to repeat the OCR later just because interger conversion failed.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -148,28 +148,77 @@ private double getPokemonLevelFromImg(Bitmap pokemonImage, int trainerLevel) {
      * If there was no detected upgrade cost, returns -1
      *
      * @param pokemonImage The image of the full pokemon screen
-     * @return the evolution cost, or -1 on scan failure
+     * @return the evolution cost, or -1 on no upgrade cost, or -999 on scan failure
      */
     public int getPokemonEvolutionCostFromImg(Bitmap pokemonImage) {
+        int error = -999;
```
sarav: Does -999 have a special meaning elsewhere outside this function? I'm trying to see if this can be cleaned up or if we need to define a "final" to denote errors. But returning errors isn't really good Java style. That's what exceptions are for.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: I'd have the cache map to Integer, or to Optional<Integer> if we use Optional.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: @sarav you can set the author to somebody else explicitly—git does that automatically when rebasing.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: Exceptions are usually recommended for exceptional errors, not for all of them.

The rest of this code is now using Guava's `Optional` instead of special defaults—it's impossible to use `Optional.absent` as an integer 😉 . (The only caveat is that you should never use `.get`without checking `isPresent`). While that's not as good as a proper `Optional` from  Java 8, I'll take what I can get ;-).

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
nahojjjen: If the integer conversion failed, then it's not a valid value that can ever be used. When the value is recovered from the cache, the line 162

 > if (stringCacheEvoCandyCost != null) return Integer.parseInt(stringCacheEvoCandyCost);

Would crash the program. There is no scenario where we would want to save the unusable value as far as i can see.

Blaisorblade: Then we could change that line. But the caching assumes OCR is deterministic given the same input, so we should remember that we can't OCR this image, not try again to get a new failure.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -119,9 +119,9 @@ private Bitmap replaceColors(Bitmap myBitmap, int keepCr, int keepCg, int keepCb
                 allpixels[i] = replaceColor;
             }
         }
-
-        myBitmap.setPixels(allpixels, 0, myBitmap.getWidth(), 0, 0, myBitmap.getWidth(), myBitmap.getHeight());
-        return myBitmap;
+        Bitmap clone = Bitmap.createScaledBitmap(myBitmap, myBitmap.getWidth(), myBitmap.getHeight(), false);
```
Blaisorblade: I think the goal is to create a separate Bitmap, but using `createScaleBitmap` without needing to rescale the bitmap can't be the best answer; that also makes `setPixels` redundant. I think we want to create an empty bitmap and then use `setPixels`, as in (untested):
```java
Bitmap clone = Bitmap.createBitmap(myBitmap.getWidth(), myBitmap.getHeight(), myBitmap.getConfig());
clone.setPixels(allpixels, 0, myBitmap.getWidth(), 0, 0, myBitmap.getWidth(), myBitmap.getHeight());
```

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -156,14 +234,15 @@ private String hashBitmap(Bitmap bmp) {
...
+        return src.replace("1", "l").replace("0", "o").replace("5", "s").replace("2", "z");
     }
 
     /**
      * Correct some OCR errors in argument where only numbers are expected.
      */
     private static String fixOcrLettersToNums(String src) {
-        return src.replace("O", "0").replace("l", "1").replace("Z", "2");
+        return src.replace("S", "5").replace("s", "5").replace("O", "0").replace("o",
+                "0").replace("l", "1").replace("I", "1").replace("i", "1").replace("Z", "2");
```
Blaisorblade: We might want to lowercase the string in all callers, instead of needing both uppercase and lowercase replacements.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
sarav: Yeah I know I can set it explicitly... But I didn't want to keep doing it or accidentally forget to. But like you said rebase doesn't touch author. So, that should be fine.

DJCrapsody:
Ah okay. So this PR is the discussion thread now. Maybe a better PR title?

nahojjjen:
What's currently happening with this pr?

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: Instead of this change, one can use #485.

Blaisorblade:
Rebased as #502.

PR #434, opened 2016-09-18
Improve handling of snapshot path and fix bugs with it

Blaisorblade:
Split out from #430. Gonna merge based on the existing review there.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1543,12 +1543,12 @@ public class Pokefly extends Service {
         @Override
         public void onReceive(Context context, Intent intent) {
             Bitmap bitmap = (Bitmap) intent.getParcelableExtra(KEY_BITMAP);
+            if (bitmap == null) {
+                return;
+            }
             String ss_file = intent.getStringExtra(KEY_SS_FILE);
             if (ss_file == null) {
                 ss_file = "";
-            }
-            if (bitmap == null) {
-                return;
             }
             scanPokemon(bitmap, ss_file);
             bitmap.recycle();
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -255,7 +255,7 @@
     private Optional<Integer> pokemonCP = Optional.absent();
     private Optional<Integer> pokemonHP = Optional.absent();
     private double estimatedPokemonLevel = 1.0;
-    private @NonNull Optional<String> ssFile = Optional.absent();
+    private @NonNull Optional<String> screenShotPath = Optional.absent();
```
sarav: Path is ambiguous -- whether it's dir or file. That's why I had used File. But not a big deal.

sarav:
I'm still getting a crash on isEmpty() in deleteIfRequired() when I click "checkIV"

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -991,9 +990,9 @@ private Pokemon interpretWhichPokemonUserInput() {
      * screenshot, and then deletes the screenshot.
      */
     private void deleteScreenShotIfRequired() {
-        if (batterySaver && !ssFile.isEmpty()) {
```
Blaisorblade: @sarav This is the only call to `isEmpty` and gets removed here. So I'm not sure about your message—and if it's something else that is crashing in `deleteScreenShotIfRequired`, I'm not sure what.

> I'm still getting a crash with Google Now on Tap share on isEmpty() in deleteScreenShotIfRequired() when I click "checkIV"

Blaisorblade:
@sarav Happy to look at a stacktrace. Re Optional, see

https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained#optional

sarav:
This is what happens when I try to test while falling asleep! I had build the apk, rebased to master and then pushed. This fix works. My bad.

PR #438, opened 2016-09-19
Add actions to notifications, V3

### LINKED ISSUE #268: Add Notification Action Buttons

sarav:
Updating #404 to avoid some UI twitches and clean up the intents a bit.

EDIT: to summarize:

> This adds the "Open app" and "Stop" actions and improves the overall notification style and attributes on latest Android versions.

Closes #404.
Supersedes #371 (its first version).
Fixes #268.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -307,6 +307,7 @@ public class MainActivity extends AppCompatActivity {
      * Starts the PokeFly background service which contains overlay logic.
      */
     private void startPokeFly() {
+        launchButton.setText(R.string.main_starting);
         launchButton.setEnabled(false);
 
         int statusBarHeight = getStatusBarHeight();
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -312,7 +312,7 @@ public class MainActivity extends AppCompatActivity {
      * Starts the PokeFly background service which contains overlay logic.
      */
     private void startPokeFly() {
-        ((Button) findViewById(R.id.start)).setText(R.string.main_stop);
+        launchButton.setText(R.string.main_stop);
 
         int statusBarHeight = getStatusBarHeight();
         Intent intent = Pokefly.createIntent(this, trainerLevel, statusBarHeight, batterySaver);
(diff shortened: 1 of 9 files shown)


sarav:
@Blaisorblade @thearaks -- take a look please.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -296,7 +291,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
     protected void onResume() {
         super.onResume();
 
-        updateLaunchButtonText(Pokefly.isRunning());
+        LocalBroadcastManager.getInstance(this).sendBroadcast(new Intent(Pokefly.ACTION_UPDATE_UI));
```
Blaisorblade: I wondered if we still need `updateLaunchButtonText` for some path, but apparently (from testing) the button is now updated correctly anyway, even when resuming. 👍  @thearaks is there some scenario where this is needed?

Blaisorblade: Ah sorry I get it: the point is that this intent triggers an UI update, but delays it: `onResume` can be called after we send the intent to start Pokefly, the new intent is delivered *after* the intent to start Pokefly (because intents are delivered in order). The ordering is not really documented but likely enough to work: http://stackoverflow.com/a/28513424/53974. I'll add a comment after merging this.

thearaks:
For me it's ok!

Just a note, we can think about this after merging:
I'd add an extra to the ACTION_UPDATE_UI Intent that says if  service is starting or stopping so that MainActivity doesn't have to call isPokeflyRunning().

Blaisorblade:
> I'd add an extra to the ACTION_UPDATE_UI Intent that says if service is starting or stopping so that MainActivity doesn't have to call isPokeflyRunning().

We can think about that later. Off the cuff, calling `isPokeflyRunning` takes IMHO less code and maybe less runtime than an extra parameter (and has no opportunity for bugs rather than a small one), even though conceptually passing a boolean together is simpler than needing a callback.

PR #441, opened 2016-09-19
Prepare for releasing 3.2

Blaisorblade:
The changelog is based on @nahojjjen's, but he said we might want to take another look.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,39 +3,35 @@
 ## 3.2.0
 
 ### Features
-* More appraisal narrowing:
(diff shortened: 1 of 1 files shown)


Blaisorblade:
If this is OK I recommend a squash merge—the separate commits only highlight the changes to the original.

sarav:
Squash merge seems like the right thing to do for a change log.

Blaisorblade:
Whoops, I screwed that up :facepalm:

sarav:
How? By not squashing? Everything else seems fine.

Issue #388, opened 2016-09-09
Pick colors for IV highlighting


Blaisorblade:
In #336 we implemented better IV highlighting—which colors to use was still an open discussion. We should figure that out.

Blaisorblade:
Pinging @Remiscan since he already did a few experiments.

Remiscan:
Here's what I tested and the reasoning behind it:

- Light Green / Dark Green : http://i.imgur.com/b1rgsOE.png

I tested this one because I felt that 13-14 and 15 should both be in different shades of green, because they both represent good IVs. I thought the contrast between the light green text and the background was a bit low, so I tried making the light green darker.

- Green / Dark Green : http://i.imgur.com/2MrQO4k.png

But then the two shades of green were too similar, so I tried using blue instead, since the spectrum of light goes red-yellow-green-blue, it felt like it could work.

- Green / Turquoise : http://i.imgur.com/XSG56kJ.png

I still wanted 13-14 and 15 to look close to each other, because the difference in IV is really small, so I picked turquoise instead of blue, because it's closer to green.

- Green / Blue : http://i.imgur.com/HUvqkHl.png

Then i just tried with blue, because it makes it easier to differentiate 13-14 and 15, and I thought people might prefer that.

- Dark Green / Blue : http://i.imgur.com/qT99AJZ.png

And I tried the original dark green with blue again.

I think I would be ok with any of them, except the green / dark green one. I'd vote for turquoise but that's just because I like the color :p 

I guess it depends on whether we want the colors to show the "closeness" between 13-14 and 15, or to just be easy to distinguish. I think the "easy to distinguish" ones (the green / blue ones) would probably be better, if people feel that blue is a natural progression of the red-yellow-green spectrum.

Blaisorblade:
> I think the "easy to distinguish" ones (the green / blue ones) would probably be better, if people feel that blue is a natural progression of the red-yellow-green spectrum.

Personally, I find easy to distinguish most important, especially accounting for sunlight and non-perfect screens (or less extreme variants).

Paxtez:
For the each stat IV, why not use a gradient like the catch color circle?  0 pure red, 15 pure green, with 7/8 being yellow?  Simple, matches the theme used by the game.

Blaisorblade:
I think we already decided to follow the appraisal system and there are good arguments there—many care about the difference between 14 and 15.

The catch color uses red-orange-yellow-green, so it does have enough colors. But yellow on white background is unreadable.

[...]

PR #442, opened 2016-09-19
Change colors for IV highlighting

Remiscan:
After much discussion in #336 and #388, no consensus has been reached about which colors to use to highlight IVs on the results screen. However, I propose that we use these red/orange/green/blue colors - http://i.imgur.com/jenqZv5.png - for the following reasons:
- of the few people who gave their opinion about the colors, a (very small) majority prefered the darkgreen/blue combination.
- I think it makes IVs easier to distinguish compared to the colors that are currently used.
- red/orange/green/blue still follows the natural progression of the spectrum of light, so blue feels like it naturally means "better than green".
- I tested 3 color combinations (the one from #336, lightgreen/darkgreen from #388 and the one from this pull request) on a color blindness simulator (Coblis). This PR looks a bit better than the one from #336, and the lightgreen/darkgreen one could be almost impossible to distinguish for some people.
- Green and blue from this PR both have better contrast on the white/grey background than lightgreen from #336.

Does anyone disagree?

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GuiUtil.java
+++ b/app/src/main/java/com/kamron/pogoiv/GuiUtil.java
@@ -10,10 +10,10 @@ import android.widget.TextView;
  */
 public class GuiUtil {
 
-    private static final String RED = "#8A0808";
-    private static final String ORANGE = "#DBA901";
-    private static final String LIGHT_GREEN = "#A0A53F";
-    private static final String GREEN = "#088A08";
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Using colorblind testing is really cool, and the colors make sense to me. So if there are no comments this goes in into 3.2 (it also doesn't affect translations, so it's safe this late).

Blaisorblade:
Mentioned in chat and opinions were neutral to positive, so merged.

PR #451, opened 2016-09-20
Show button if starting on pokemon screen (#437)

### LINKED ISSUE #437: GoIV doesn't show button if starting on pokemon screen
Happening with the latest code.

1. Start Pokemon Go without starting GoIV service. 
2. Go to a Pokemon screen.
3. Go to GoIV, make sure "Launch Pokemon Go" option is enabled. Start the service.
4. Go back to Pokemon Go.

The button does not appear until the screen is pressed. If the "Launch Pokemon Go" option is enabled, it should do a first capture to check if the user has started GoIV while having a pokemon already selected. Minor issue since any touch will force the button to appear.

Blaisorblade:
Fix #437 in a very hacky/simple way: if we're starting Pokemon Go, trigger a
(delayed) scan at startup.

I'm not sure if/when we should merge this: it appears to work on my Nexus 7. Before that I had 10 screens from PoGo in a row though—I'm rather sure I hadn't installed this change yet so it should be unrelated, but I still wonder, so:

Q1: could this increase the frequency of black screens? EDIT: Further testing suggests no.

Also, technically, we'd want the scan to be started after MainActivity activates Pokemon Go (through an intent). But this code worked on my devices (testing on my Nexus 5), and in the worst case (except for Q1) it just won't fix #437 which is just an (annoying) cosmetic issue.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -386,6 +386,9 @@ public class Pokefly extends Service {
             if (!batterySaver) {
                 screen = ScreenGrabber.getInstance();
                 watchScreen();
+                if (GoIVSettings.getInstance(getBaseContext()).shouldLaunchPokemonGo()) {
+                    triggerDelayedScan();
+                }
             } else {
                 screenShotHelper = ScreenShotHelper.start(Pokefly.this);
             }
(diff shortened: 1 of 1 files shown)


Blaisorblade:
@sarav opinions?

sarav:
This is the same fix I would have written -- so the code itself and the idea are good by me.

But the fix assumes that the phone is fast enough to open the PoGo app in the Pokemon screen within a few seconds (4 seconds I believe) of STARTing the service. I'm not sure how true that assumption is across devices. So, it'll fix it for some but might not fix it for some.

I'd rather just document this in the README or FAQ and just mark the issue as WONTFIX and call it a day. We can say something like:

"If the IV button doesn't show up in a Pokemon screen, try to tap the screen once to see if it shows up".

Blaisorblade:
We should totally document this very clearly (hence #452) before release.

> But the fix assumes that the phone is fast enough to open the PoGo app in the Pokemon screen within a few seconds (4 seconds I believe) of STARTing the service. I'm not sure how true that assumption is across devices. So, it'll fix it for some but might not fix it for some.

This is tested on my (slow) phone, and the scenario is "the PoGo app is already open". But yes it's a hack.

Blaisorblade:
Closing because of https://github.com/farkam135/GoIV/issues/437#issuecomment-248278572.

PR #454, opened 2016-09-20
Update Gradle to 2.2.0

DJCrapsody:
More info in [release notes](https://developer.android.com/studio/releases/gradle-plugin.html)
REMINDER: Update to Android Studio 2.2!


sarav:
Give us a few days please. Also, why are you forcing an update to 2.2.0? Wouldn't it be used automatically if someone has a newer version?

Blaisorblade:
I tagged this as 3.3.0—let's not merge this before a release.

> Also, why are you forcing an update to 2.2.0? Wouldn't it be used automatically if someone has a newer version?

There are two updates here, for a Gradle plugin and for Android Studio. IIUC:

1. The new plugin seems to work with old Android Studio. After release we can standardize on 2.2 if it works well.
2. This config change is needed to use the new Gradle plugin IIUC. I'm no Gradle expert, but [this quote](https://developer.android.com/studio/releases/gradle-plugin.html) and other stuff suggests that no, the build file picks a specified Gradle version—you can let Gradle pick but that's not recommended, in a word because it makes the build unreproducible:

> Caution: You should not use dynamic dependencies in version numbers, such as 'com.android.tools.build:gradle:2.+'. Using this feature can cause unexpected version updates and difficulty resolving version differences.

PR #457, opened 2016-09-21
Add PidgeySpam calculation (PokeSpam)

NightMadness:
Hi Team,
As part of playing PoGo on a daily basis I find myself starring pokemon to prepare for my lucky egg usage where I do the following calculation:
1. Take the # of candy and divide that with the candy cost. Your answer should be rounded down to the nearest whole number. 
2. Take the # from step 1 and divide by 3. This makes the counting faster because there are 3 pokemon to a row. Any remainder will let me know that there are extra pokemon are left. 

Example: 
    PidgeySpam candy = 536, Candy Cost = 12
    Result = 44 pokemon OR 14 rows + 2 more

I did the following:
1. Get Candy amount from OCR
2. Modifed screen to include candy
3. Calculate how many pokemon we can evolve 
    and Calculate how many pokemon fit in a row
4. Display PokeSpam on screen
5. English translation
6. Added unit tests

Pokemon Spam Examples: 
https://pokeassistant.com/main/pidgeyspam?locale=en
http://www.pidgeycalc.com/

Please check the candy amount  on your phone and let me know that it works, my phone is Samsung S7 Edge. Also, this is my first gitHub commit and first android code commit/write, please let me know if i need to redo,adjust,setup,do something a different way.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -295,10 +295,14 @@ public class OcrHelper {
             try {
                 //If "/" comes at the end we'll get an array with only one component.
                 String[] hpParts = pokemonHPStr.split("/");
-                String hpStr =
-                        hpParts.length >= 2
-                                ? hpParts[1]
-                                : hpParts[0];
+                String hpStr;
+                if (hpParts.length >= 2) {
+                    hpStr = hpParts[1];
+                } else if (hpParts.length == 1) {
+                    hpStr = hpParts[0];
+                } else {
+                    return Optional.absent();
+                }
 
                 return Optional.of(Integer.parseInt(fixOcrLettersToNums(hpStr).replaceAll("[^0-9]", "")));
             } catch (NumberFormatException e) {
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -4,7 +4,6 @@ import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.support.annotation.NonNull;
 import android.util.LruCache;
-import android.widget.Toast;
 
 import com.google.common.base.Optional;
 import com.googlecode.tesseract.android.TessBaseAPI;
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -336,12 +336,27 @@ public class OcrHelper {
 
 
     /**
-     * Gets the candy amount from a pokenon image.
+     * Gets the candy amount from a pokemon image.
      *
      * @param pokemonImage the image of the whole screen
      * @return candyAmount the candy amount, or blank Optional object if nothing was found
      */
     private Optional<Integer> getCandyAmountFromImg(Bitmap pokemonImage) {
+
+        //disable getting CandyAmount if we dont have pokespam running
+        GoIVSettings Settings;
+        try {
+            //Anything better then getInstance(null)? should we let it crash?
+            Settings = GoIVSettings.getInstance(null);
+        } catch (Exception ex) {
+            //probably a crash due to getInstance
+            return Optional.absent();
+        }
+
+        if (Settings == null || !Settings.isPokeSpamEnabled()) {
+            return Optional.absent();
+        }
+
         Bitmap candyAmount = Bitmap.createBitmap(pokemonImage,
                 (int) Math.round(widthPixels / 1.515), (int) Math.round(heightPixels / 1.44),
                 (int) Math.round(widthPixels / 5.0), (int) Math.round(heightPixels / 38.4));
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -339,7 +339,7 @@ public class OcrHelper {
      * Gets the candy amount from a pokenon image.
      *
      * @param pokemonImage the image of the whole screen
-     * @return the candy name, or "" if nothing was found
+     * @return candyAmount the candy amount, or blank Optional object if nothing was found
      */
     private Optional<Integer> getCandyAmountFromImg(Bitmap pokemonImage) {
         Bitmap candyAmount = Bitmap.createBitmap(pokemonImage,
@@ -356,7 +356,7 @@ public class OcrHelper {
         }
         candyAmount.recycle();
 
-        if (pokemonCandyStr.length()>0) {
+        if (pokemonCandyStr.length() > 0) {
             try {
                 return Optional.of(Integer.parseInt(fixOcrLettersToNums(pokemonCandyStr).replaceAll("[^0-9]", "")));
             } catch (NumberFormatException e) {
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -4,6 +4,7 @@ import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.support.annotation.NonNull;
 import android.util.LruCache;
+import android.widget.Toast;
 
 import com.google.common.base.Optional;
 import com.googlecode.tesseract.android.TessBaseAPI;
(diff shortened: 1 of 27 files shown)


sarav:
Sorry, this seems completely outside the scope/point of GoIV. I don't think this will be or should be merged.

DJCrapsody:
Sounds good but out of scope. Could be kept in view for future release in 3.4.0 or 3.5.0.

Read CONTRIBUTING.md for code style and commit guidelines.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -938,6 +949,12 @@ private boolean parseNumericInputs() {
         } catch (NumberFormatException e) {
             return false;
         }
+        //do not require pokemon candy to be filled
+        try {
+            pokemonCandy = Optional.of(Integer.parseInt(pokemonCandyEdit.getText().toString()));
+        }  catch (NumberFormatException e) {
+            pokemonCandy = Optional.of((int)0);
```
Blaisorblade: I'd have expected `Optional.absent()` here rather than `Optional.of((int)0)`.

Blaisorblade:
Hi @NightMadness ! First of all, thanks for the contribution—the code looks good and you win a ⭐  for being the first to add real unit tests at all. I'd love to have your contributions. @DJCrapsody's right about code style and commit guidelines but hopefully that's easy to fix and we can help.

Regarding scope, yes this is different. But it is a useful feature, I also do this calculation often.

Somebody proposed having a plugin system on Reddit. In that system, this would be a separate plugin. But that system is not done yet, is it? If that comes to light, this feature should move there.

The only concerns are user confusion and maintenance effort.
Regarding user confusion, we plan to add to expanded results more numbers that need explanation.
Regarding maintenance, 

Anyway, this can't go in for 3.2.0 because we're releasing soon and we're de facto in feature freeze—new features would be cool but we don't want to risk bugs right away.

phate89:
"PidgeySpam candy = 536, Candy Cost = 12
Result = 44 pokemon OR 14 rows + 2 more"
That's not how it works.. If you evolve one pokemon they give you back 1 candy.. So only the first truly uses 12 candies.. with 536 pidgey candies you can evolve 47 pidgeys

sarav:
@NightMadness, As @Blaisorblade said, this would be an excellent option if we have plugin support. Sadly we don't. Don't mean to discourage you from contributing to this project though.

NightMadness:
@phate89 Yes you are right, during spam i usally dont tranfer my pokemon since it take extra seconds. I agree some people would like to see that, other people might want other features like how much time is left for egg and candy left. This why I moved the code to an object so new functions can be added in future.

@Blaisorblade and @sarav thank you for keeping a positive attitude and helping me out. While i do agree plugin is where it should go in the future, there are other features like gym helped that we want to include. I see GoIV as an all out tool to help the player play better which utilize ocr to avoid user input. I think we should include it for now and have it writen so it be easier to refactor to a plugin when plugins come to effect, this will also will highlight where we need plugins to be have acess to the software (ocr phase.. dialog.. dispay.. etc).  Let me know what you think.

Even if pokespam doesnt go to this project I would still like to contribute. I hope this is just the beggining.

phate89:
@NightMadness 
the extra candy is not from transferring. If you evolve a pokemon they give a candy for free without transferring. If you transfer you get another one.
For example with 23 candies you can evolve 2 pidgeys without transferring anything. You evolve the first with 12 candies and they give back a candy so you can evolve again without transferring the first pidgeotto-

NightMadness:
@phate89 Ahh, I wasnt aware, that mean cost of evolve is actually N-1 once you can have at least N candies. N being amount of candy need to evolve.  Wow good to know!

DJCrapsody:
Shit. GitSkarios crashed when I typed 1k characters. I'm going to type it again.

Regarding PidgeySpam (or PokeSpam) formula, formula should be:
initialCost: `n`
effectiveCost: `n - 1` (1 candy for evolving)
effectiveCostWithTransfer: `n - 2` (1 candy for transferring)
holdingCandy: `h`

To get the correct algorithm, the code should:
1. Check if `h >= n`
2. If 1 is `true`, check if user would like to transfer after; evolving; else inform user `Not enough candy to evolve.`
3. If 2 is `true` for transfer, `FLOOR(h / effectiveCost)`; else `FLOOR(h / effectiveCostWithTransfer)` (this will be numberOfEvolvable)

However, this will only get how many Pokemon you can evolve with your current holding candy.

To calculate number of Pokemon to evolve for max XP yield and other info, you need more formula:
holdingPokemon: `p`
extraCandyextraCandy: `p - numberOfEvolvable` (1 candy for each extra Pokemon transferred, this can be used to show how many Pokemons to be transferred before evolving if `extraCandy >= 0`)
requiredCandy: `p * effectiveCost` OR `p * effectiveCostWithTransfer`
actualCandy: `h + extraCandy`
actualEvolvable: `FLOOR(actualCandy / requiredCandy)`
candyLeft: `actualCandy - requiredCandy`

phate89:
@DJCrapsody that formula isn't correct.
the first transfer will always cost 1 more so the right formula is 'FLOOR((h-1) / effectiveCost)'
for example if you have 22 pidgey candy with your formula you will get 2 but it's wrong because the first cost 12 and it will give you back 1 leaving you with 11. that's not enough for a 2nd evolution.
the same applies to the other formulas

DJCrapsody:
@phate89, read the formula again, n - 2 comes in when you get 1 from evolving AND 1 from transferring.

Just in case, the steps:
22 candy and 2 Pokemon
11 candy and 1 Pokemon after evolution (22 - 12 + 1)
12 candy and 1 Pokemon after transferring (11 + 1)
0 candy after evolving last Pokemon

phate89:
@DJCrapsody read again your formula. you used `FLOOR(h / effectiveCost)`
with 22 and not transferring you have:
`h = 22`
`effectiveCost = 11`
`FLOOR(22 / 11)`
that give you 2 evolution when you only can evolve one

DJCrapsody:
I see where you are coming from. The +1 should be done after evolution rather than -1 from cost. Will edit the formula.

DJCrapsody:
@phate89 revised formulas.

NightMadness:
How about
(Total candy - bonus) / (evolution cost - bonus)
Floor ((22 - 1) / (12 - 1)) = (21 ÷ 11) = 1

Blaisorblade:
@NightMadness Not enough alone, you need @DJCrapsody's variant. Example:
Candies ->Evolutions
11 -> 0,
12 -> 1,
22 -> 1,
23 -> 2

phate89:
@Blaisorblade for that you can simply check before doig the math:
```
int evolutions = 0;
if (h >= cost)
  evolutions = Math.floor((h-1) / (cost - 1);
```

DJCrapsody:
ok I got the final formula. took me a bus ride to ensure its correct.

Blaisorblade:
Sounds good. This PR already sets up testcases; I'd propose to add testcases from the discussion.
The other requirement is an option to enable this.

NightMadness:
I dont think we need to check that h>=cost. Math should do that for us.
if (cost-1 != 0)
  evolutions = Math.floor((h - 1) / (cost - 1));
Floor((11-1))/11= floor (10/11) = 0

DJCrapsody:
It's good practice to check `h >= cost` when you translate it to code. You can `return -1` and show a toast without any calculations.

Blaisorblade:
1. We're IMHO past the point for debate—time for code with testcases.
2. @NightMadness is right on the formula, not sure what I was thinking. See point 1.
3. `cost` should be non-0 at that point.
4. Returning `-1` to mean `0` seems like it complicates things.
5. If one formula is always correct, let's keep that. We don't even need floating-point it seems, integer division should do the trick (not that *one* division is critical).

DJCrapsody:
Alright, here's a [working prototype](http://pastebin.com/Fhm5AArM) of PokeSpam that can be compiled on [Online Java IDE](https://www.compilejava.net).

Blaisorblade:
Man thanks! (But why do you use that rather than Android Studio?)

DJCrapsody:
Because there may be one or two lurkers out here who doesn't use it that may be interested in this. :p

Blaisorblade:
I see, but please don't keep torturing yourself with tools that destroy your work!

DJCrapsody:
Yeah damn. Clicking the back button wouldn't show the "You're leaving the page!" warning. :(

NightMadness:
You are right, I will add option to enable/disable.

sarav:
@NightMadness -- when it's disabled, it shouldn't change the UI in anyway. UI already takes up the whole screen on my phone.

NightMadness:
Hi Team,
What is the best way for accessing settings from OCR helper I used GoIVSettings.getInstance(null); which seems wrong to me.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -336,12 +336,27 @@ private String getCandyNameFromImg(Bitmap pokemonImage) {
...
+        GoIVSettings Settings;
+        try {
+            //Anything better then getInstance(null)? should we let it crash?
+            Settings = GoIVSettings.getInstance(null);
+        } catch (Exception ex) {
+            //probably a crash due to getInstance
+            return Optional.absent();
+        }
+
+        if (Settings == null || !Settings.isPokeSpamEnabled()) {
```
Blaisorblade: Remember local variables start lowercase, uppercase is for classnames, so `Settings.isPokeSpamEnabled` alone looks like a static method call...

Re `getInstance(null)`, I'll see what we can do, but a crash should not be possible.

sarav:
@NightMadness, what settings does OCR need to know? And if it's needed, pass it in as part of the constructor.

DJCrapsody:
Do a rebase (see CONTRIBUTING.md) in the future, I'm seeing duplicated commits.

NightMadness:
@sarav just seeing if PokeSpam Enabled, if its enabled it will not check or return candy amount
@DJCrapsody  hmmm something here went wrong, i was working on a different branch testing creating Plugins, i think i merged it by accident.

sarav:
@NightMadness -- just pass that boolean when you init OCR. We don't need to support run time changing of that flag

Blaisorblade:
> @DJCrapsody hmmm something here went wrong

Common problem: if you merged master into your branch and *then* do a rebase, you get "copies" of the commits on master. But once this is ready I guess we can help cleaning up the commits.

NightMadness:
Hi Team, 
Code completed, can you take a look at it? 
I had trouble with Merges but I finished the PluginHelper code.

This will allow add plugins in a separate folder without effecting Settings or results Display screen. 
There are triggers at key points that will hopefully allow for us to not be depended on making changes on PokeFly for new features.

If we want to revert without the plugin helper this would be the commit to revert to https://github.com/farkam135/GoIV/pull/457/commits/7dedc5ed0977e5a5cc0ed676f340a44e1dbc6c80

Please let me know what you think. Do you think we should continue with this route?

Blaisorblade:
Hi @NightMadness ! It's cool that you've worked on adding a plugin system, but I suspect I'd like to hold off on that for now.

I'm rebasing your PR without the plugin system—not with git rebase, I tried and it doesn't do the right thing on this convoluted history. Please wait for me.

Blaisorblade:
I've opened #486. I'll implement suggestion https://github.com/farkam135/GoIV/pull/457#issuecomment-249409932 on top.

PR #461, opened 2016-09-21
Ensure GoIV survives "Don't keep activities"

Blaisorblade:
If I enable "Don't keep activities" (tested under Android 7 on my Nexus 5X), GoIV dies immediately after going away from the MainActivity, which is to be expected from the description.

This PR fixes the problem, and generally enables GoIV to keep working if Android kills MainActivity (which it can do at will). Conversely, Android will try to keep Pokefly active till possible, because it is considered a foreground (hence important) activity, thanks to our persistent notification (technically, thanks to `startForeground`), according to [relevant docs](https://developer.android.com/reference/android/app/Service.html#ProcessLifecycle).

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -354,10 +354,6 @@ public class MainActivity extends AppCompatActivity {
 
     @Override
     public void onDestroy() {
-        if (Pokefly.isRunning()) {
-            stopService(new Intent(MainActivity.this, Pokefly.class));
-        }
-
         LocalBroadcastManager.getInstance(this).unregisterReceiver(pokeflyStateChanged);
         LocalBroadcastManager.getInstance(this).unregisterReceiver(showUpdateDialog);
         super.onDestroy();
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -357,9 +357,6 @@ public class MainActivity extends AppCompatActivity {
         if (Pokefly.isRunning()) {
             stopService(new Intent(MainActivity.this, Pokefly.class));
         }
-        if (screen != null) {
-            screen.exit();
-        }
 
         LocalBroadcastManager.getInstance(this).unregisterReceiver(pokeflyStateChanged);
         LocalBroadcastManager.getInstance(this).unregisterReceiver(showUpdateDialog);
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -368,11 +368,7 @@ public class Pokefly extends Service {
         running = true;
 
         if (ACTION_STOP.equals(intent.getAction())) {
-            if (screen != null) {
-                screen.exit();
-            }
             stopSelf();
-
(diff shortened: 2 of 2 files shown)


Blaisorblade:
Review by @sarav.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -357,9 +357,6 @@ public void onDestroy() {
         if (Pokefly.isRunning()) {
             stopService(new Intent(MainActivity.this, Pokefly.class));
         }
-        if (screen != null) {
-            screen.exit();
-        }
```
sarav: Hmmm... I thought I had asked @thearaks to remove this when he added Pokefly.isRunning(). I guess not.

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -451,6 +447,10 @@ private void unwatchScreen() {
         screenScanHandler.removeCallbacks(screenScanRunnable);
         screenScanRunnable = null;
         screenScanHandler = null;
+        if (screen != null) {
+            screen.exit();
+            screen = null;
+        }
```
sarav: I'd prefer it isn't inside this function. Call from onDestroy directly or something? This is just about watch/unwatching the touch events.

sarav:
This PR gives us a "duh!" moment. Obvious on hindsight kinda PR. Made one minor comment. Good to go after that.

Blaisorblade:
Note: I can't reproduce this easily, but the first time I start with this patch, the screen goes black or does strange stuff when I start GoIV (and then goes back to normal).

sarav:
Let me give it a shot maybe tomorrow? In 24 hrs?

app/src/main/java/com/kamron/pogoiv/MainActivity.java
thearaks: When I first read this code, I tought this was an incorrect behavior.
I think Pokefly should run even if MainActivity gets destroyed.
@sarav I don't remember about your request, maybe I just lost it between the "notification" PR comments...
Anyway, +1 for me!

app/src/main/java/com/kamron/pogoiv/Pokefly.java
Blaisorblade: 👍

sarav:
@Blaisorblade To make sure I understand this correctly, are you saying this change is creating some issue that's hard to reproduce for you?

Blaisorblade:
> To make sure I understand this correctly, are you saying this change is creating some issue that's hard to reproduce for you?

@sarav Yeah, but this was unrelated to the change—it's #466.

sarav:
@Blaisorblade Just as a FYI, this is why swiping away was sufficient to kill GoIV completely and restart it. WIth this change, that won't be possible. People will have to forcestop. This PR is still correct though.

thearaks:
@sarav That also meant that if the MainActivity was destroyed to free up system memory, the Service would do the same (even if the user still needs it).
Anyway, now with the "stop" notification action is possible to stop Pokefly at any time, even after MainActivity gets swiped away from the recent apps. 
Also, since it's a service, I think it should always survive to MainActivity destroys! :)

Blaisorblade:
Agreed with both—I wouldn't have merged this without "Stop" from the notification.

PR #462, opened 2016-09-21
Fix warning/error with String formatting

Blaisorblade:
Android Studio 2.2 considers this a serious warning and even marks it as
an error. While the code works perfectly fine it's overly general (and
might be inefficient). So let's fix the warning.


DJCrapsody:
I'm fixing this with my next focus, resource refactoring.

PR #468, opened 2016-09-23
Refactor resources and string fixes

### LINKED ISSUE #450: Drop Japanese language support
The discussion was lost somewhere on discord.

So... We haven't really implemented Japanese support (didn't include jap.traineddata, no Japanese translation other than Pokemon names).

Reason for dropping is because there aren't any requests for Japanese and translator for Japanese strings.

Adding back Japanese support would be quite simple but only if the above criteria are met.

DJCrapsody:
Split base stats from integer.xml to their respective files
Move colors and dimens into respective files
Drop Japanese locale support
Remove redundant files
Simplify string formats (i.e. use %s instead of %1$s if only 1 string)
Fix wrong string formats (i.e. %d for trainer_level)
Minor text fixes (REALLY!)

Fixes #450


app/src/main/res/values/strings.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools">
-    <string name="pokemon029">Nidoran♀</string>
-    <string name="pokemon032">Nidoran♂</string>
+    <string name="pokemon029" translatable="false">Nidoran♀</string>
+    <string name="pokemon032" translatable="false">Nidoran♂</string>
```
Blaisorblade: Are we still using these? We should have stopped.

app/src/main/res/values/strings.xml
```diff
@@ -103,11 +103,11 @@
...
     <string name="evolution">Evolution</string>
     <string name="level">Level</string>
-    <string name="level_num">Level %1$.1f</string>
+    <string name="level_num">Level %.1f</string>
     <string name="refine">Refine</string>
     <string name="refine_no_progress">Refinement didn\'t improve results</string>
-    <string name="last_scan">Last scan: %1$s</string>
-    <string name="possible_iv_combinations" tools:ignore="PluralsCandidate">"%1$d possible IV combinations"</string>
+    <string name="last_scan">Last Scan: %s</string>
+    <string name="possible_iv_combinations" tools:ignore="PluralsCandidate">\"%d possible IV combinations\"</string>
```
Blaisorblade: Again `\"'. Will test this.

Blaisorblade:
There's one issue. Just tested this change, as a representative:
```
diff --git a/app/src/main/res/values/strings.xml b/app/src/main/res/values/strings.xml
index 73486af..fbbde67 100644
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -107,7 +107,7 @@
     <string name="refine">Refine</string>
     <string name="refine_no_progress">Refinement didn\'t improve results</string>
     <string name="last_scan">Last scan: %1$s</string>
-    <string name="possible_iv_combinations" tools:ignore="PluralsCandidate">"%1$d possible IV combinations"</string>
+    <string name="possible_iv_combinations" tools:ignore="PluralsCandidate">\"%1$d possible IV combinations\"</string>
     <string name="too_many_iv_combinations">Too many IV combinations</string>
     <string name="refining_instructions">Increase IV calculation precision by leveling up and comparing two scans of the same Pokémon</string>
     <string name="appraisal_refining">Appraisal Refining</string>
```

Result:
image

app/src/main/res/values/strings.xml
DJCrapsody: It's being used in initOcr, but since we're dropping Jap support, I'm moving this to code.

DJCrapsody:
@nahojjjen somehow you've added all the redundant quotes 😱

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -28,10 +28,10 @@
     private final int heightPixels;
     private final int widthPixels;
     private final boolean candyWordFirst;
-    private final String nidoFemale;
-    private final String nidoMale;
+    private final String nidoFemale = "Nidoran♀";
+    private final String nidoMale = "Nidoran♂";
```
Blaisorblade: No, these names can be translated. The plan was to fetch them from the pokemon array rather than duplicating it.

Blaisorblade:
Having commits doing multiple things means it's hard to cherry-pick the bits that matter. I'm merging #462 meanwhile (since it's a clear improvement), since this one isn't converging to a fix.

DJCrapsody:
Will be resolving the conflicts. Still trying to get a hang of committing the changes bit by bit (seems just weird to commit just 1 tiny change).

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: Turns out these particular names never get translated apart from Japanese, out of sheer luck... But still there's no need to duplicate these constants. Fixing soon.

DJCrapsody: I'll see what I can do. The only translation for the Nido family is from English to Japanese (or generally non-latin character languages).

Blaisorblade:
I have other ideas on other bits though... For instance, I don't think it's an improvement to drop `"` to add `\`.

DJCrapsody:
Oops I just saw that. and I broke something so I'll be fixing that by doing an amendment.

app/src/main/res/values-fr/strings.xml
```diff
@@ -100,11 +100,11 @@
     <string name="stardust_cost">Coût en Poussière</string>
     <string name="evolution">Évolution</string>
     <string name="level">Niveau</string>
-    <string name="level_num">Niveau %1$.1f</string>
+    <string name="level_num">Niveau %.1f</string>
     <string name="refine">Raffiner</string>
     <string name="refine_no_progress">Le raffinement n\'a pas amélioré les résultats</string>
-    <string name="last_scan">Dernier scan : %1$s</string>
```
Blaisorblade: I'm sorry to inform you, that's not how French punctuation works.

app/src/main/res/values-fr/strings.xml
```diff
@@ -9,8 +9,8 @@
         <item name="valor">"Bravoure"</item>
         <item name="instinct">"Intuition"</item>
     </string-array>
-    <string name="place1">"Dans l\'ensemble, ton Pokémon&#8230;"</string>
-    <string name="place2">"Ses stats&#8230;"</string>
+    <string name="place1">Dans l\'ensemble, ton Pokémon&#8230;</string>
+    <string name="place2">Ses stats&#8230;</string>
```
Blaisorblade: Based on my understanding of the syntax, this looks like actually a bugfix (the current code will show a `\'` to the user). @Remiscan @toinux or does this look right?

app/src/main/res/values-fr/strings.xml
DJCrapsody: @Remiscan please fix the punctuation

app/src/main/res/values-fr/strings.xml
DJCrapsody: The quotation marks appeared on every single file including default (English). I don't think is part of syntax. \ is added so ' can appear right.

Remiscan: `<string name="place1">"Dans l\'ensemble, ton Pokémon&#8230;"</string>`
This code does not show a `\'` on my phone. Which is weird, because I think I remember having to delete the `\` from the strings that are between quotation marks for this exact reason.

DJCrapsody: I'll go ahead and remove \?

app/src/main/res/values-fr/strings.xml
Remiscan: In French there is a space both before and after a colon. There's also a space before lots of punctuation marks like ! ? and ;

DJCrapsody: I see have noted

app/src/main/res/values-fr/strings.xml
Remiscan: @DJCrapsody This page https://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling says we can choose to use the quotes or the \ but there is no need for both, so you could just remove the \ as you said.

DJCrapsody: We'll keep the \, because appraisal text are actual quotes so that'll keep things consistent

DJCrapsody:
All set

DJCrapsody:
All set. RTM

app/src/main/res/values-fr/strings.xml
```diff
@@ -113,7 +113,7 @@
     <string name="stamina_hp">PV</string>
     <string name="missing_inputs">PC et PV requis.</string>
     <string name="ivtext_no_possibilities">Aucune combinaison d\'IV possible pour ces données. Veuillez vérifier les informations.</string>
-    <string name="checking_for_update">Recherche de mise à jour…</string>
+    <string name="checking_for_update">Recherche de mise à jour&#8230;</string>
```
Blaisorblade: Why??

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
Blaisorblade: I've dealt with this.

app/src/main/res/values-fr/strings.xml
DJCrapsody: I'm pretty sure my commit fixed this.

Blaisorblade:
I've extracted out a few bits which were clearly good. PR coming.

I guess these sound worthwhile:
* Split base stats from integer.xml to their respective files
* Move colors and dimens into respective files
* Remove redundant files

For the others:
* Simplify string formats (i.e. use %s instead of %1$s if only 1 string)
I'm ambivalent on that. The rule we actually need is "never use just %s if there are two arguments" because translations might need to show arguments in different orders.

* Minor text fixes (REALLY!)
Doing them together with tons of other changes got in the way of merging the rest.

nahojjjen:
The project has changed so much that this PR has become very outdated, and parts of it has been fixed in other PRs. I'm closing this PR to clean up the Pull request log.

PR #469, opened 2016-09-23
Add perfection and hp estimates to evolution box

nahojjjen:
This pr adds two new fields in the powerup and evolution box in the results window. HP and %CP vs max iv.

unnamed

I'm probably not going to add tier to the results, and instead just add it as an option for user configurable clipboard in another pr.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
@@ -323,7 +323,7 @@ public class PokeInfoCalculator {
      * @return An integer representing how much hp selectedpokemon with ivscanresult stamina ivs has at selectedlevel
      */
     public int getHPAtLevel(IVScanResult ivScanResult, double selectedLevel, Pokemon selectedPokemon) {
-        double lvlScalar = Data.CpM[(int) (selectedLevel * 2 - 2)];
+        double lvlScalar = Data.CpM[Data.levelToLevelIdx(selectedLevel)];
         int highHp = (int) Math.max(Math.floor((selectedPokemon.baseStamina + ivScanResult.highStamina) * lvlScalar),
                 10);
         int lowHp = (int) Math.max(Math.floor((selectedPokemon.baseStamina + ivScanResult.highStamina) * lvlScalar),
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
@@ -312,4 +312,23 @@ public class PokeInfoCalculator {
 
         return list;
     }
+
+    /**
+     * Get how much hp a pokemon will have at a certain level, including the stamina IV taken from the scan results.
+     * If the prediction is not exact because of big possible variation in stamina IV, the average will be returnred.
+     *
+     * @param ivScanResult    Scan results which includes stamina ivs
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
```diff
@@ -312,4 +312,23 @@ private Pokemon getLowestEvolution(Pokemon poke) {
...
+     * Get how much hp a pokemon will have at a certain level, including the stamina IV taken from the scan results.
+     * If the prediction is not exact because of big possible variation in stamina IV, the average will be returnred.
+     *
+     * @param ivScanResult    Scan results which includes stamina ivs
+     * @param selectedLevel   which level to get the hp for
+     * @param selectedPokemon Which pokemon to get Hp for
+     * @return An integer representing how much hp selectedpokemon with ivscanresult stamina ivs has at selectedlevel
+     */
+    public int getHPAtLevel(IVScanResult ivScanResult, double selectedLevel, Pokemon selectedPokemon) {
+        double lvlScalar = Data.CpM[(int) (selectedLevel * 2 - 2)];
```
Blaisorblade: There's `Data.levelToLevelIdx(selectedLevel)` for this—but I should use this also for the existing code. Fixing this...

Blaisorblade:
Looks rather good! I'm making a couple minor fixups but I like the code. And the CP index makes sense to me.

app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
```diff
@@ -323,7 +323,7 @@ private Pokemon getLowestEvolution(Pokemon poke) {
      * @return An integer representing how much hp selectedpokemon with ivscanresult stamina ivs has at selectedlevel
      */
     public int getHPAtLevel(IVScanResult ivScanResult, double selectedLevel, Pokemon selectedPokemon) {
-        double lvlScalar = Data.CpM[(int) (selectedLevel * 2 - 2)];
+        double lvlScalar = Data.CpM[Data.levelToLevelIdx(selectedLevel)];
```
Blaisorblade: I should probably add a Data.getCpM(double level) method for this.

Blaisorblade:
LGTM — are you fine with those fixups?

sarav:
@nahojjjen -- so, you gave up on the %perfect IV? That's different from %perfect CP right? Because CP biases more for atk?

sarav:
I wish we had gotten the HP info in 3.2.0. If we haven't merged a lot of stuff since 3.2.0, should we release a 3.2.1 with this?

nahojjjen:
Sorry for slow replies, Ive been bedridden sick since morning.

@Blaisorblade I think the changes you made are great and I have no complaints or issues with them.

@sarav I motivated removing the proprietary "battle score" after researching how different the results between battle score and CP was - the difference was not big enough that I felt that forcing a new standard on the users. While comparing different max cp does bias the results in favor of attackers, the defenders still get a decent score.

Blaisorblade:
@sarav 3.2.1 might be an option if we have time, but there's also the crashes on Fabric right now... But yes, HP could be good.

app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
Blaisorblade: I did this now.

PR #478, opened 2016-09-24
Gradle clean up

DJCrapsody:
Minor clean up to gradle
Removed constrainlayout dependency, most should be on AS2.2, doesn't seem to affect build

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -127,35 +127,33 @@ android {
  * annotation processing)
  */
 dependencies {
+    /**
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Even this PR should be two commits, so one could merge the cleanup separately from dropping the dependency. I'd merge it though if the commit description at least documented what it was doing—the PR description is not accessible from Git history.

Blaisorblade:
Cherry-picked this onto master (with fixups). Closing. Thanks!

PR #244, opened 2016-08-28
Remove unnecessary SharedPrefs writes from settings

harkin:
the `PreferenceFragment` can do all the saving for us

this can be merged independent of any changes to how settings are accessed

PR #480, opened 2016-09-25
Custom clipboard

nahojjjen:
This PR adds the ability to customise what GoIV adds to the clipboard after a pokemon scan.

Left to do: 
Move the edit button from the main screen to the settings
Add the ability for different clip-settings depending on if the result is a range or a single iv.


nahojjjen:
The PR is ready for testing.

app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
```diff
@@ -56,6 +62,22 @@ public void setPlayerTeam(int value) {
...
+
+    public void setClipboardPreference(ArrayList<ClipboardToken> tokens) {
+        SharedPreferences.Editor editor = prefs.edit();
+        String saveString = "";
+        for (ClipboardToken token : tokens) {
+            saveString += token.getStringRepresentation();
+        }
+        editor.putString(GoIVSettings.GOIV_CLIPBOARDSETTINGS, saveString);
+        editor.commit();
+    }
```
Blaisorblade: To check: right now we don't explicitly commit other settings, check if we can be consistent here. Also, `commit` is discouraged.

app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
```diff
@@ -231,6 +231,26 @@ public CPRange getCpRangeAtLevel(Pokemon pokemon, int lowAttack, int lowDefense,
...
+     * @return int representing the average cp of the high and low estimate of the pokemon
+     */
+    public double getAverageCPAtLevel(Pokemon pokemon, int lowAttack, int lowDefense, int lowStamina, int highAttack,
+                                   int highDefense, int highStamina, double level) {
+        CPRange range = getCpRangeAtLevel(pokemon, lowAttack, lowDefense, lowStamina, highAttack, highDefense,
+                highStamina, level);
+        return (range.high + range.low) / 2;
+    }
+
+    /**
```
Blaisorblade: Possible simplification: make this a method on CPRange.

Blaisorblade:
Overall I like this a lot! I have some minor issues I'm happy to fix. I also think tokens are going to need unit testing, otherwise there's no chance to not have bugs in there.
For instance, right now, name truncation didn't work.

I also have issues with the design of pokemon name truncation (it should IMHO be automatic).

Blaisorblade:
Other issues:
- PoGo doesn't show black Unicode numbers. Not our fault, but we should consider work around it if it happens for all.
- the UIV token picks a combination even when many are possible.
- I'd like a token to show an IV% range if one is needed, a single IV if the value is certain.

app/src/main/res/layout/activity_main.xml
```diff
@@ -58,6 +58,16 @@
 
         <LinearLayout
             android:layout_width="match_parent"
+            android:layout_height="wrap_content">
+            <Button
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="Edit Clipboard Setting"
+                android:onClick="navigateToClipboardModifier"/>
```
Blaisorblade: Could somebody take a look at the margins here? I think we need some, not sure why we don't have a good default.

app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
Blaisorblade: I was thinking of #244, but it seems that trick doesn't work here.

Blaisorblade: I've taken care of all that needed changing.

nahojjjen: I agree that we should be consistent with how settings are saved, but I currently dont completely understand how it's done, so I implemented it in this way. If anyone wants to change it to be consistent with the rest of the app, I wouldnt mind.

nahojjjen:
@Blaisorblade 
> the UIV token picks a combination even when many are possible.
I picked that it always gives the lowest iv combination because of personal preference, but I agree that it would probably be a good idea to generalise for other people with other ideas.

>PoGo doesn't show black Unicode numbers. Not our fault, but we should consider work around it if it happens for all.

Not completely true, it's not PoGo, its device specific. For instance, black unicode works fine on my device.
unnamed

I'd like a token to show an IV% range if one is needed, a single IV if the value is certain. EDIT: I probably should add it.

I actually wanna solve this in a more general way, I want the user to be able to have two different settings for when there's a single or multi-result. That way you can create even more specific settings. For instance, I plan on having the IV display with black unicode if the scan was a single result.

Blaisorblade:
Gotta figure out what's wrong with black unicode—for me some characters worked and some didn't...

> I actually wanna solve this in a more general way, I want the user to be able to have two different settings for when there's a single or multi-result.

That might be a smarter idea. I had a few special cases in mind, but I'm sure there's some case I missed. It complicates the interface of course, but it's probably worth it. And this is already a "geeky" corner of the interface anyway.

Can I also add another wish? Don't hate me, but is there a way to reorder the tokens without removing and readding them? With drag-and-drop? If there is, I failed to find it...

nahojjjen:
Afaik there are two different black unicode sets, I might have copypasted a weird set for some of the numbers. I think i used three different sources, one for the 0, one for 1-9 and one for 10-15. This is of course not a good solution, it would be better to generate the string array programmatically by adding the unicode in a loop.

nahojjjen:
As for drag and dropping, that's something that would be great to add. This is only the first version of the ui, it can certainly be polished and extended. I didn't add it in the first version, as I couldnt figure out any trivial way to handle the UI code for it.

We could potentially change the copy-paste ui so that instead of clicking a token in the edit part, it doesnt remove the token, but rather selects it. After that, the description window could transform into an edit field, with remove, move left, and move right as options. That'd be easy to implement. It's not as elegant as a drag and drop solution though.

app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
Blaisorblade: As mentioned, I've taken care of that 😉

Blaisorblade:
I've opened a separate issue about reordering tokens, so we can do it after this PR and not forget it.

Blaisorblade:
> I picked that it always gives the lowest iv combination because of personal preference, but I agree that it would probably be a good idea to generalise for other people with other ideas.

Hm, so you'd like to keep this option in I understand? I'm asking because otherwise (for other reasons) I'd have liked to drop `getLowestIVCombination()`.

Blaisorblade:
> Afaik there are two different black unicode sets, I might have copypasted a weird set for some of the numbers. I think i used three different sources, one for the 0, one for 1-9 and one for 10-15. This is of course not a good solution, it would be better to generate the string array programmatically by adding the unicode in a loop.

So my problem was with a 9. You might want to look at @Eccenux's code for this (with credit of course)—the code has a MIT license, so we'd just need to add it to ours. (I can do that if needed if you're unsure of the details).
https://github.com/Eccenux/UnicodeNumbers/tree/master

Blaisorblade:
Still TODO: 
- [ ] `BaseStatToken.mode` should also be an enum if we keep it
- [ ] not sure that "Prevent NullPointerException in tokens" is actually needed, I'd like a better alternative if possible.
- [ ] support both single and multi-result.

nahojjjen:
@Blaisorblade Regarding eccenux unicode, please do that if you have the time.

oskomonzon:
This is really impressive. Amazing work, @nahojjjen :+1:

Blaisorblade:
It's better to have this PR in than out—I wouldn't dream of not running this branch. Hence merging even with all the outstanding requests above (most of which are enhancements).

PR #483, opened 2016-09-25
Info overlay defaults to appraise box closed

### LINKED ISSUE #432: Input dialog should start with appraisal mode closed
Appraisal mode starts expanded if you used it on your last pokemon, but that means it covers the HP/CP for the new one. @sarav doesn't like it, and I find it annoying because I only use it on a few pokemon, so I don't want it open by default.

[Quoting @sarav](https://github.com/farkam135/GoIV/issues/372#issuecomment-247825059):
> Every time I scan the pokemon, it opens up in appraisal mode that ends up blocking CP and HP. I'll have to fold appraisal mode to see what the HP is (which is barely missed from being blocked) and then open appraisal again and go through the rest of the steps. Adding this feature would make the steps a lot more simpler.

nahojjjen:
Fixes #432

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -845,9 +845,13 @@ public class Pokefly extends Service {
     private static final int MAX_DRAWABLE_LEVEL = 10000;
 
     private void toggleVisibility(TextView expanderText, LinearLayout expandedBox, boolean animate) {
+        setVisibility(expanderText, expandedBox, animate, expandedBox.getVisibility() != View.VISIBLE);
+    }
+
+    private void setVisibility(TextView expanderText, LinearLayout expandedBox, boolean animate, boolean visible) {
         int boxVisibility;
         Drawable arrowDrawable;
-        if (expandedBox.getVisibility() != View.VISIBLE) {
+        if (visible) {
             boxVisibility = View.VISIBLE;
             arrowDrawable = getDrawableC(R.drawable.arrow_expand);
         } else {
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Thanks, but don't we need usually `toggleVisibility` for this? We can skip the animation here, but we need to fix the arrow.

Blaisorblade:
I've just verified the arrow's indeed in the wrong position with this fix.

nahojjjen:
I actually tried to find if we had any methods like this, but couldnt. I blame the size of pokefly / that the method does not have javadoc. Anyway, redid the pr in #484

Blaisorblade:
I agree Pokefly's a mess. To find this, I looked through uses of `appraisalBox`. #484 doesn't do the right thing, but right now this one doesn't work right either because skipping the animation doesn't work :-|

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -853,23 +855,27 @@ private void toggleVisibility(TextView expanderText, LinearLayout expandedBox) {
             arrowDrawable = getDrawableC(R.drawable.arrow_expand);
         }
         expanderText.setCompoundDrawablesWithIntrinsicBounds(null, null, arrowDrawable, null);
-        Animator arrowAnimator = ObjectAnimator.ofInt(arrowDrawable, "level", 0, 10000).setDuration(100);
-        arrowAnimator.start();
+        if (animate) {
+            Animator arrowAnimator = ObjectAnimator.ofInt(arrowDrawable, "level", 0, MAX_DRAWABLE_LEVEL).setDuration(100);
+            arrowAnimator.start();
+        } else {
+            arrowDrawable.setLevel(MAX_DRAWABLE_LEVEL);
```
Blaisorblade: OK, this was the line the clinched the deal (at least in basic testing).

PR #485, opened 2016-09-25
Generalize replaceColors

Blaisorblade:
Refactoring to support #431. (Untested).

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -84,8 +84,7 @@ public class OcrHelper {
      * threshold.
      *
      * @param srcBitmap    The source bitmap to scan.
-     * @param dstBitmap    The destination bitmap. Must have same size as the source, can coincide with it, or be null
-     *                     (in which case it is allocated here).
+     * @param mutateSrc    Indicates whether to mutate srcBitmap or to produce a new one.
      * @param keepCr       The red color to keep
      * @param keepCg       The green color to keep
      * @param keepCb       The blue color to keep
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Should be straightforward, but review by @sarav. (Testing now).

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -92,8 +94,8 @@ private boolean isCandyWordFirst() {
      * @param simpleBG     Whether the bitmap has a simple background
      * @return Bitmap with replaced colors
      */
-    private Bitmap replaceColors(Bitmap srcBitmap, int keepCr, int keepCg, int keepCb, int replaceColor, int distance,
-                                 boolean simpleBG) {
+    private Bitmap replaceColors(Bitmap srcBitmap, Bitmap dstBitmap, int keepCr, int keepCg, int keepCb,
```
sarav: I think a "create copy" or some sort of boolean would be a much clearer API. You also won't have to depend or assume if the input bitmap is the same side.

sarav: I do understand that having "dstBitmap" would allow a caller to reuse the same bitmap, but I'm not sure this is worth it.

Blaisorblade: Good point: the API I'm using allows `dstBitmap` to be `!= null` and `!= srcBitmap`, but we're not using that.

Blaisorblade: I'm overly sleepy but I just added the new API in (untested) 810873854f4e20192119183474bd58b452c8bb54. I can squash with the previous commit.

PR #488, opened 2016-09-26
Adds new Czech strings.

Eremiell:


app/src/main/res/values-cs/strings.xml
```diff
@@ -62,4 +62,7 @@
     <string name="atk">ÚTK</string>
     <string name="def">OBR</string>
     <string name="sta">STA</string>
+    <string name="cp_vs_iv" formatted="false">% CP vs max IV</string>
```
Blaisorblade: That doesn't look Czech? I'll merge if that's on purpose of course.

Eremiell: well, I considered changing it in some way, but honestly all of it it ok in Czech and all alternatives that happened to me are inelegant in some way. could drop the whole line atm, but I decided to keep it if I come will some better option later on.

PR #499, opened 2016-10-02
Encapsulate screen info (from metrics and trainer level)

Blaisorblade:
Two reasons:
- this is tricky logic
- some features might require different screen info (e.g. scanning screenshots from other phones)
- some features might require different trainer level.

But there are still other places using screen dimensions, in ScreenGrabber and in Pokefly. I'm not sure they should be changed though, since at least ScreenGrabber really needs the dimensions from the current screen.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -10,7 +10,6 @@ import android.content.IntentFilter;
 import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.database.Cursor;
-import android.graphics.Point;
 import android.graphics.Rect;
 import android.media.projection.MediaProjection;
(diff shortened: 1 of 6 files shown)


nahojjjen:
I like these changes, but I'd like a javadoc explaining what responsibilities the screeninfo class does if it's not too much trouble. I think there's generally a problem with our code that a lot of methods and classes lack documentation. Since we're not persistent in how we handle this, I don't blame you if you don't feel like adding it yourself.

Blaisorblade:
Thanks for the reminder! I'm happy to get comments on how to improve the code. And I do this often I admit, so it's good you point that out. Especially for this class since its "responsibility" is not trivial to articulate.

sarav:
I'll try to review this later. No promises.

Blaisorblade:
Closing due to inactivity (and me saying goodbye to the project).

PR #502, opened 2016-10-03
Candy cost identification v3

### LINKED ISSUE #381: Use nidoran gender recognition also when nidoran's detected from the candy

Blaisorblade:
Supersedes #408 and #431. I rebased #431 on master, used the new version of `replaceColors` and added Checkstyle fixes.
EDIT: Fixes #381.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -209,11 +209,11 @@ public class OcrHelper {
                 result = 400; //god damn magikarp
                 ocrResult = "400";
             }
-
-            ocrCache.put(hash, ocrResult);
         } catch (NumberFormatException e) {
             result = error; //could not ocr text
+            ocrResult = String.valueOf(error); //Store error code instead of scanned value
         }
+        ocrCache.put(hash, ocrResult);
         return result;
 
     }
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -201,6 +201,10 @@ public class OcrHelper {
         String ocrResult = fixOcrLettersToNums(tesseract.getUTF8Text());
         try {
             result = Integer.parseInt(ocrResult);
+            if (result == 10) { //second zero hidden behind floating button
+                result = 100;
+                ocrResult = "100";
+            }
             ocrCache.put(hash, ocrResult);
         } catch (NumberFormatException e) {
             result = error; //could not ocr text
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -318,17 +318,29 @@ public class OcrHelper {
             tesseract.setImage(name);
             pokemonName = fixOcrNumsToLetters(tesseract.getUTF8Text().replace(" ", ""));
             if (pokemonName.toLowerCase().contains("nidora")) {
-                boolean isFemale = isNidoranFemale(pokemonImage);
-                if (isFemale) {
-                    pokemonName = nidoFemale;
-                } else {
-                    pokemonName = nidoMale;
-                }
+                pokemonName = getNidoranGenderName(pokemonImage);
             }
             name.recycle();
             ocrCache.put(hash, pokemonName);
         }
         return pokemonName;
+    }
+
+    /**
+     * Get the correctly gendered name of a pokemon.
+     *
+     * @param pokemonImage The image of the nidoranX.
+     * @return The correct name of the pokemon, with the gender symbol at the end.
+     */
+    private String getNidoranGenderName(Bitmap pokemonImage) {
+        String returner = "";
+        boolean isFemale = isNidoranFemale(pokemonImage);
+        if (isFemale) {
+            returner = nidoFemale;
+        } else {
+            returner = nidoMale;
+        }
+        return returner;
     }
 
     @NonNull
@@ -365,6 +377,9 @@ public class OcrHelper {
             candyName = fixOcrNumsToLetters(
                     removeFirstOrLastWord(tesseract.getUTF8Text().trim().replace("-", " "), candyWordFirst));
             candy.recycle();
+            if (candyName.toLowerCase().contains("nidora")) {
+                candyName = getNidoranGenderName(pokemonImage);
+            }
             ocrCache.put(hash, candyName);
         }
         return candyName;
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -157,28 +157,80 @@ public class OcrHelper {
      * If there was no detected upgrade cost, returns -1.
      *
      * @param pokemonImage The image of the full pokemon screen
-     * @return the evolution cost, or -1 on scan failure
+     * @return the evolution cost, or -1 on no upgrade cost, or -999 on scan failure
      */
-    public int getPokemonEvolutionCostFromImg(Bitmap pokemonImage) {
+    private int getPokemonEvolutionCostFromImg(Bitmap pokemonImage) {
+        int error = -999;
         Bitmap evolutionCostImage =
                 Bitmap.createBitmap(pokemonImage, (int) (widthPixels * 0.625), (int) (heightPixels * 0.86),
                         (int) (widthPixels * 0.2), (int) (heightPixels * 0.05));
         String hash = "candyCost" + hashBitmap(evolutionCostImage);
+
+        //return cache if it exists
         String stringCacheEvoCandyCost = ocrCache.get(hash);
-        int result = -1;
-        if (stringCacheEvoCandyCost == null) { //if no cached result
-            //the dark color used for text in pogo is 76,112,114
-            evolutionCostImage = replaceColors(evolutionCostImage, true, 76, 112, 114, Color.WHITE, 40, false);
-            tesseract.setImage(evolutionCostImage);
-            stringCacheEvoCandyCost = tesseract.getUTF8Text().replace("S", "5").replace("s", "5");
-            try {
-                result = Integer.parseInt(stringCacheEvoCandyCost);
-            } catch (NumberFormatException e) {
-                stringCacheEvoCandyCost = "-1";
-            }
-            ocrCache.put(hash, stringCacheEvoCandyCost);
+        if (stringCacheEvoCandyCost != null) {
+            return Integer.parseInt(stringCacheEvoCandyCost);
+        }
+
+        //clean the image
+        //the dark color used for text in pogo is approximately rgb 76,112,114 if you can afford evo
+        //and the red color is rgb 255 95 100 when you cant afford the evolution
+        Bitmap evolutionCostImageCanAfford = replaceColors(evolutionCostImage, false, 80, 110, 110, Color.WHITE, 45,
+                false);
+        Bitmap evolutionCostImageCannotAfford = replaceColors(evolutionCostImage, false, 255, 95, 100, Color.WHITE, 30,
+                false);
+
+        boolean affordIsBlank = isOnlyWhite(evolutionCostImageCanAfford);
+        boolean cannotAffordIsBlank = isOnlyWhite(evolutionCostImageCannotAfford);
+        //check if fully evolved
+        if (affordIsBlank && cannotAffordIsBlank) { //if there's no red or black text, there's no text at all.
+            ocrCache.put(hash, "-1");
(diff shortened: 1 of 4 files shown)


Blaisorblade:
I still have comments in #431 to address, and @nahojjjen had some bugs he wanted to address on this before merging it (though I keep forgetting).

nahojjjen:
I think the only "problem" with the current code is that it's not been tested for  non-english and that there are still Log calls here and there. 

The other things I wanted to tweak have already been fixed or are no longer applicable.

Blaisorblade:
One thing you IIRC kept talking about were issues with pokemon with multiple evolutions. If that's not relevant I'll take a look at my other remarks.

nahojjjen:
Yea, multiple evolution pokemon will return a perfect match for candy, for multiple pokemon. But the current implementation will pick the first alternative. I don't think this is a major issue because even with all pokemon, it's quite a small part that are affected, and the user can still manually correct it.

But you're right that this was an issue I forgot Id brought up. 

How would this be solved? We could make it so the mandy match guesser skips on multi-ev pokemon. That wouldnt change much, as it'd still try the next guesser -to find the closest name within the ev-line. I dont think that would return a more accurate result often enough that the added code complexity for handeling this edge case would be worth it.

nahojjjen:
Ive actually got to correct myself, I looked back and noticed that I also wrote that there was occasionally an issue with scanning the candy cost 100, because the last zero would be hidden partially behind the button, and that occasionally caused the scan to miss-identify the pokemon. 

Ill try to see if I can make a fix for the 100-scan issue, and remove the log-calls.

nahojjjen:
Ive fixed the issue with scanning 100, and I also noticed that Nidoran was not correctly identified if it was renamed, because the nidoran image recognition method wasnt used on the candy name, only if the nickname contained "nidora". So I made it so both candy name and pokemon nickname use the nidoran gender image recognition method if they contain"nidora".


If your dont find any errors in review of the changes, I think this will be ready for merge.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -153,6 +153,92 @@ private double getPokemonLevelFromImg(Bitmap pokemonImage, int trainerLevel) {
...
+    private boolean isOnlyWhite(Bitmap refinedImage) {
+        int[] pixelArray = new int[refinedImage.getWidth()];
+
+        //below code takes one line of pixels in the middle of the pixture from left to right
+        refinedImage.getPixels(pixelArray, 0, refinedImage.getWidth(), 0, refinedImage.getHeight() / 2, refinedImage
+                .getWidth(), 1);
+
+        // a loop that sums the color values of all the pixels in the array
+        for (int pixel : pixelArray) {
+            if (Color.red(pixel) != 255 || Color.green(pixel) != 255 || Color.blue(pixel) != 255) {
```
Blaisorblade: I think this condition could be replaced by `pixel != Color.rgb(255, 255, 255)`, like we do elsewhere; that should be faster.

Blaisorblade:
Oh cool, sounds like you fixed #381! Apart from formatting, there's at least https://github.com/farkam135/GoIV/pull/431#discussion_r79311304 to address, for the reasons explained there.

nahojjjen:
I think I addressed what you referred to with #431 with 68b0b0dd6e9b39ae2ace2b0b8b065b32c72402ad

Blaisorblade:
Rebased, now reviewing...

Blaisorblade:
Testing now on Eevolutions, user corrections are broken there and this always selects Vaporeon... user corrections should have priority at least here, and probably in general.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -224,7 +224,7 @@ private double getPokemonLevelFromImg(Bitmap pokemonImage, int trainerLevel) {
         Optional<Integer> result = getPokemonEvolutionCostFromImgUncached(evolutionCostImage);
         String ocrResult;
         if (result.isPresent()) {
-            ocrResult = String.valueOf(result); //Store error code instead of scanned value
+            ocrResult = String.valueOf(result.get()); //Store error code instead of scanned value
```
Blaisorblade: FTR, this crash was my doing.

Blaisorblade:
The problem with user corrections should now be solved.

DJCrapsody:
User corrections should get priority but sometimes [this] (https://www.reddit.com/r/GoIV/comments/57dsl1/320_offline_build_no_longer_scans_renamed_pokemon/) happens.

Blaisorblade:
Hm, good point. As I proposed [there](https://www.reddit.com/r/GoIV/comments/57dsl1/320_offline_build_no_longer_scans_renamed_pokemon/d8rpmqx), maybe we want to lookup name *and* candy *and* candy amount needed for evolution, so two pokemons with the same name but different candy name are not "corrected" to the same result? That, however, would require migrating existing save data for persistence, so it's not 100% trivial.

nahojjjen:
Regarding user corrections, I think they should be added in before stage 4.

     * 1. check if the nickname perfectly matches a pokemon
     * 2. check if candyname + evolution cost perfectly matches a pokemon
     * 3. check if there's a cached result for the scanned pokemon name
     * 4. get the pokemon with the closest name within the evolution line guessed from the candy

So the new logic would be 


     * 1. check if the nickname perfectly matches a pokemon
     * 2. check if candyname + evolution cost perfectly matches a pokemon
     * 3. check if there's a cached result for the scanned pokemon name
     * 3. check if there's a stored user correction
     * 5 get the pokemon with the closest name within the evolution line guessed from the candy


User corrections should not take priority over step 1 & 2, they are "perfect identifiers" & "fail silent" modules. The first check  can only result in service failure if the user manually renamed a pokemon to another pokemon name, and the risk of the second module giving erroneous data is extremely low (It's also fail-silent). 

If we add user corrections as step 1, it causes errorous results every time a pokemon has the same name as another, which is one of the things this new identification module was suppsed to solve. 
For example the described scenario in https://www.reddit.com/r/GoIV/comments/57dsl1/320_offline_build_no_longer_scans_renamed_pokemon/d8rqyrk breaks if user corrections are considered most important.


Basically we have different guessing modules, and they're all supposed to be fail silent except the last one, and the modules should be called in order of accuracy. Stored name corrections is not accurate, because it will have a collision every time a user has the same nickname for two different pokemon, which happens quite often.

nahojjjen:
So reading through and step-by step debugging the current code here's the chain of logic used to determine the pokemon from what I can tell:

1. Scanned nick-string matches previous correction
2. Scanned nick-string matches previous scan result
3. Pokemon name exists in database
4. ~~nickname guess within best guess evo line?~~ - impossible for this guess to ever be used, for dist to be 0 (guess 3) would have triggered
5. candyname+evcost - match the candy name and upgrade cost to a specific pokemon
6. make a wild guess

This is my analysis of these different guessing methods:


image


My suggestion to change the order to 

 * 1. check if the nickname perfectly matches a pokemon
 * 2. check if candyname + evolution cost perfectly matches a pokemon
 * 3. check if there's a cached result for the scanned pokemon name
 * 3. check if there's a stored user correction
 * 5 get the pokemon with the closest name within the evolution line guessed from the candy

remains.

nahojjjen:
I gave the commit i just uploaded a bad name, "accurate results" can be debated. It does however implement the order i suggested, which leads to more accurate results when scanning pokemon with similar (scanned the same) names. I believe users scan similarly named pokemon more often than they scan eeveelutions, and even when they scan eeveelutions, the evolution family will be correct, so the user can select the right poke in the drop down.

It does break eeveelution detection on two of the evolutions (jolteon & flareon) but i think it might be worth to add specific handling for those in the future.

My idea for the split evolution line handling would be that we can automatically find which pokemon are part of a split evolution, and if we detect that one has been scanned we look at the pokemon average color. We'd need to add the average color for each split case, but it would work without the user having to correct manually.

app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
```diff
@@ -41,54 +41,53 @@ public void putCorrection(String ocredPokemonName, String correctedPokemonName)
...
-        /* Go with the cached value; if no entry is found, returning null is correct. */
-        PokeDist cacheGuess = cachedCorrections.get(poketext);
-        if (cacheGuess != null) {
-            return cacheGuess;
-        }
-
         ArrayList<Pokemon> bestGuessEvolutionLine = getBestGuessForEvolutionLine(candytext);
-        PokeDist guess;
+        PokeDist guess = null;
         //Warning: Don't forget to call cacheResult before exiting. Preferably keep a single return point.
```
Blaisorblade: The new code violates this comment—the code I think is fine, can you update the comment?

Blaisorblade:
For eevolutions: can the candy-based (`getCandyNameEvolutionCostGuess`) simply *not* return a sure guess in that case? That would make me happy enough for now. But if that's too hard, I guess the PR could go in as is (modulo Travis style fixes).

A possible further improvement would be that, even in that case, we could still learn what the family is and use it in subsequent guesses, but it's not worth it, especially because it's unnecessary if we do the averaging.

Blaisorblade:
Since Checkstyle passes locally, I'm merging this for now in the interest of progress. Congratulations @nahojjjen!

nahojjjen:
I think your idea of keeping track of which pokemon has branched evolution families and simply not return a certain result in the candyname+upgrade method is very logical

BraisGabin:
Why dont you use the candy name + CP + HP + lvl? I use that and it works perfect with all the Pokémon except caterpie/metapod and weedle/kakuna? At that point I use candy cost (I didn't push this yet). The candy cost is a bit tricky in some phones because the floating button hides a bit the numbers so I don't want to heavily relay on that. An example: https://github.com/BraisGabin/pokescreenshot/blob/master/processing/src/androidTest/assets/screenshots/en/70_weepinbell.png

Blaisorblade:
> Why dont you use the candy name + CP + HP + lvl?

Interesting idea. At this point the user hasn't confirmed yet CP, HP and lvl, and for both our apps they're all error-prone. I hope you have something better for the level, but what do you do with Gastly when it hides the CP and/or the level? In our app we allow correcting CP/HP/lvl after we show the selected pokemon.

Assuming those are perfect, IIUC you end up doing three/four IV scans and picking the only one which has non-zero some possibilities, right?

> The candy cost is a bit tricky in some phones because the floating button hides a bit the numbers so I don't want to heavily relay on that. An example: https://github.com/BraisGabin/pokescreenshot/blob/master/processing/src/androidTest/assets/screenshots/en/70_weepinbell.png

Thanks for pointing that out, we should test against that screenshot.

BraisGabin:
I don't have any user correction implemented yet. And Gatsly is horrible, always. But I think that we can do nearly nothing about that.

This is the method that I use to guess the pokemon: https://github.com/BraisGabin/pokescreenshot/blob/master/processing/src/main/java/com/braisgabin/pokescreenshot/processing/Guesser.java#L28-L47 The pokémon selection is done with the candy name.

Blaisorblade:
> I don't have any user correction implemented yet. And Gatsly is horrible, always. But I think that we can do nearly nothing about that.

That's why we have the dialog, in general—but that's indeed not a problem for picking a pokemon in a family.

> This is the method that I use to guess the pokemon: 

Ah I see, it's enough to try with min. and max. IV—that's smarter and faster than I thought. And when we do get the right candy family (which we almost always do) it's indeed more reliable than the candy cost (and avoids the problems with both Eevolutions and hidden candy cost). I think it's cool.

sarav:
@BraisGabin I'm curious, what is it that GoIV does that doesn't fit your need and caused you to create a separate app? Just asking because if we can meet your need easily, you could be yet another dev to contribute to this project instead of maintain a separate one.

Blaisorblade:
I agree with @sarav, we'd be happy if you wanted to join...

@sarav He didn't know about GoIV:
https://www.reddit.com/r/GoIV/comments/57gedb/it_seems_that_i_reinvented_the_wheel/

> Hi, I just published my new app PokeScreenshot and some redditors pointed me out to this subreddit because this app does the same.

Also, @BraisGabin's code does have some advantages... quoting more:

> In my project I have a separated module that do all the ocr and image processing. I hope it can help. Merge the best things of your processing with the best of mine would be great, but let's see...

PR #503, opened 2016-10-03
Add support for display Pokemon Name with Traditional Chinese.

tofu0913:
I'm working on Traditional Chinese localization for GoIV, but the Pokemon Go app in this region is still in English.
If I translate pokemons.xml into Chinese directly, OCR will be failed.

This PR adds Traditional Chinese pokemon name display support.
A default false boolean setting is added in pokemons xml resource file:
True: will uses English as OCR string, but display pokemon's name localized.
False: will uses localized string for both OCR and display.


DJCrapsody:
We did not add any Chinese traineddata for tesseract, ocr will not work either way. But update when Chinese language is supported, so we can add it in (or you can do it :p)

tofu0913:
Pokemon Go doesn't have Chinese version neither, this PR is still OCR English but allow GoIV to display Chinese.
I don't know is there any country have the same situation? Pokemon Go is in English but not local language.

DJCrapsody:
Yes I'm aware. Maybe we can have translations for the goiv strings instead of the Pokemon names since the game doesn't support Chinese language (yet or ever from last I've read).

app/src/main/java/com/kamron/pogoiv/widgets/PokemonSpinnerAdapter.java
```diff
@@ -45,19 +45,26 @@ public void updatePokemonList(ArrayList<Pokemon> list) {
...
+
     @Override
     public View getView(int position, View convertView, ViewGroup parent) {
-        return getCustomView(position, convertView, parent);
+        LayoutInflater inflater = (LayoutInflater) context.getSystemService(Context.LAYOUT_INFLATER_SERVICE);
+
+        TextView row = (TextView) inflater.inflate(textViewResourceId, parent, false);
+        Pokemon pokemon = pokemons.get(position);
+
+        row.setText(pokemon.getDisplayName());
```
Blaisorblade: We'd need to add a comment explaining the whole point of this method is to show `pokemon.getDisplayName()` instead of `pokemon.toString()` (that is, the name). I also wonder whether we should just change toString instead.

Blaisorblade:
I'm trying to understand this PR and have a few questions. I apologize in advance for any cultural misunderstanding, I'm Italian and had to look up quite a few things.

IIUC, this PR is for locale `zh-rTW`, that is what in English is usually called Taiwan, and there the game is released? Asking because it took me a while to figure that out, and because this must be clear.

And even though the game is in English and has English Pokemon names, you propose to display them to the user in Traditional Chinese? Won't that be confusing for users? Or is that a matter of preference, because users prefer names from the other games to the PoGo ones? I can be convinced but I'd like a clear explanation. Before adding code I'd like to understand there's a valid use case for it.

I have some technical issues with the PR itself, but that's all probably solvable (not trivial) if we want.

tofu0913:
Sorry for lack of explaining.

Pokemon Go app was released at 8/6 in Taiwan, with English version.

As I know, the Pokemon fans in this small place were played Japanese version of Pokemon games on Gameboy/NDS since childhood. Most of players don't know the exact way to pronounce Japanese's Pokemon name at that time, until TV anime is on air. The subs on Anime is also based on Japanese, for example, Dragonite, the official Chinese name in Taiwan we call it "Kuai-Ryu" which is similar with Japanese. Another example, we call it "Ka-Bi" instead of Snorlax. I believe there are many players here are never known Pokemon's English name until Pokemon GO.

I don't think Niantic will give us a localized update in the future. However, we'll be glad to see their names which we familiar with on any 3rd party tools.

Changes I'll made:

- This PR I was tried to do not change the original behavior, usual case goes to `Pokemon.name`, zh_TW case goes to `Pokemon.translatedName`. But you're right, Java will reuse the pointers.

1. Remove null pointer check. `Pokemon.translatedName` will be equal to `Pokemon.name` when usual case.
2. I perfer to rename `Pokemon.translatedName` to `Pokemon.displayName`. it should be easier to understand.
3. `Pokemon.toString()` will return `Pokemon.displayName`. Thanks for #505.

- About user input, my idea is to modify `PokeInfoCalculator.populatePokemon()`, keep both name into pokemap. This will double the size of HashMap when zh_TW, but no affect on other languages.
- I'll keep monitoring if any other things or use cases I missed. Will update this PR later.

Blaisorblade:
Thanks a lot for your explanation! I suspect there might be new players not familiar with the series from earlier, who'll only know the English name, but I won't try to guess how many, so maybe there should be an option. But we can do this on request.

And generally, since this request is somewhat invasive, I'd like you to make extra sure this is safe for other languages. Your proposed changes sounds very good.

> Pokemon.toString() will return Pokemon.displayName. Thanks for #505.

Welcome—but I just reviewed explicit toString uses found by grepping the sources (all other ones appear safe), so I might have missed more uses (I hope not). And this ignores future changes—we'll need JavaDocs on `.name` for people to know when to use which.

> About user input, my idea is to modify PokeInfoCalculator.populatePokemon(), keep both name into pokemap. This will double the size of HashMap when zh_TW, but no affect on other languages.

Can you check if pokemap is used elsewhere? There's one use you want to affect, in `interpretWhichPokemonUserInput`, but I'm not sure that's good in `rememberUserInputForPokemonNameIfNewNickname`—though the only problem is if the OCR reads a "Taiwanese" name, and IIUC Tesseract will never read that.

Blaisorblade:
Hey @tofu0913 ! I imagine you're short on time, but are you still interested in this?

Blaisorblade:
@cfchen Thanks for the cleanup, they match most of my requests. IIUC the only point that is open is this:

> About user input, my idea is to modify PokeInfoCalculator.populatePokemon(), keep both name into pokemap. This will double the size of HashMap when zh_TW, but no affect on other languages.

but I forget what I had in mind—should take a closer look later.

tofu0913:
I've updated the commit, user input is included. Also I've added a setting, this setting will only visible if locale is Traditional Chinese.
If user is using Traditional Chinese, he can choose to use Traditional Chinese Pokemon name or stay in English.

Blaisorblade:
I've just tested this, and the problems in #535 seem all unrelated. Hence merging.

PR #505, opened 2016-10-03
Avoid using toString in lookup

Blaisorblade:
`Pokemon.toString()` should just have a cosmetic function (I proposed
changing it in #503). And there's no need for a lookup. The problem is
only that `getSelectedItem` is declared to return `Object` when in fact
it returns `Pokemon` (part of the involved classes have not been
generified yet).

Testing ongoing.


Blaisorblade:
Testing passed!

Issue #471, opened 2016-09-24
Input box moves way too high when appraisal is opened


sarav:
In some devices (my tablet), the input box moves way too high that the top is kinda going into the notification area. So, clicking the pokemon name to correct the detection is kinda impossible/hard.

Also, even without that issue, I think it would be better to have the bottom of the input box lined up with the 25% point (wild guess, but we should be able to find this from screenshots) of the screen from the bottom than having the top of the input box lined up to the top of the screen. That way, it'd be easier to reach on devices with bigger screen and takes care of this notification bar issue.

sarav:
UI guys: @nahojjjen @oskomonzon @DJCrapsody

nahojjjen:
Im not sure I understand the issue, could you post a screenshot?

sarav:
![screenshot_20160926-152458](https://cloud.githubusercontent.com/assets/247061/18854097/d8069f68-83fd-11e6-9a1b-ced20ba825a6.png)

sarav:
@nahojjjen I couldn't reproduce the issue where the UI was overlapping with the notification area. But the 2nd point should be visible here. On a tablet, the UI moves way too far. It could be a lot further down (easy to reach) while still not blocking the appraisal UI.

nahojjjen:
I think adding logic to "how far up" the overlay should move could be complex, I think the current solution is to set a flag (top) to tell the overlay where it should move, we don't actually tell the overlay to move to a position in percent or pixels.

Unless there's a bug where you actually can't click the input like you suggested in the original post, I think we might want to avoid messing with the overlay position code.

Even if we do decide to implement a custom position for the overlay, it would not be as simple as "25% to the bottom" as some small phones have the overlay take more than 75% of the screen. (Z3 compact)

cubeyz:
Close the appraisal refining bit while you try fix the top parts.

sarav:
Actually, can we make the window dragable vertically so that the user can adjust it for their screen and the app can remember it. How hard would that be?

oskomonzon:
@sarav I already have a working prototype of that. Minor details to sort out yet but it's working fairly well.

NightMadness:
If this is issue fixed we can close this.

nahojjjen:
Yea

PR #507, opened 2016-10-06
Added position handler for appraisal window

oskomonzon:
This is to address #471 and hopefully cater a wide range of users usability concerns.

- [x] Added position handler to input dialog in appraisal mode
- [x] Added position handler logic
- [x] Implemented save, load and visibility logic for the position handler

Tagging @Blaisorblade, @nahojjjen, @DJCrapsody, @thearaks, @sarav and @Eremiell as usual for feedback and code review.


Blaisorblade:
Code LGTM (based on in-browser review). And this should not be very device-specific (though of course you never know), so this should be mergeable as soon as somebody else tests it.

nahojjjen:
Testing it, it seems to be working flawlessly 👍 

To be honest, with how well its working, I wouldn't be against having the window movement listener active even when appraisal is not active, on all of the overlay types.

nahojjjen:
Minor comments that might be worth considering:

1. Since its a button, the user might think the button moves the window up/down on click, and not understand that you have to drag it. Might be worth to Toast "drag to move window" if the user moves the window less than x pixels after the listener is done.

2. The padding is slightly off compared to the rest of the elements in the overlay.

oskomonzon:
I'm glad you guys like it. I've been thinking about @nahojjjen first point and it makes sense, adding a toast upon release if there was no position change.
Or maybe it isn't a terrible idea to have it behave like a toggle between "bottom" and "custom position" if the user just taps on it.

Blaisorblade:
What's the status of this PR? There's the status feedback from Travis to fix, and the comments from @nahojjjen to address. @oskomonzon are you waiting for feedback or just low on time?

oskomonzon:
@Blaisorblade very low of time, I've been out of the city again during the weekend.
I'll add the toast as sugested by @nahojjjen unless any of you guys have another proposal to discuss.

oskomonzon:
Rebased to keep it clean. This are the last changes:

- [ ] Merged and cleaned conflicts with master
- [ ] Added toast to instruct the user to move the window if he just tapped the handler
- [ ] Added new string to the strings resource files for the new toast
- [ ] Small changes to make it Travis compliant

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -968,7 +963,7 @@ private void moveOverlay(Boolean moveUp) {
         WindowManager.LayoutParams newParams = (WindowManager.LayoutParams) infoLayout.getLayoutParams();
         if (moveUp) {
             newParams.gravity = Gravity.TOP;
-            newParams.y = sharedPref.getInt(APPRAISAL_WINDOW_POSITION, 0);
+            newParams.y = sharedPref.getInt(APPRAISAL_WINDOW_POSITION,0);
```
Blaisorblade: that's not good ;-)

oskomonzon: Lol, I was so happy it passed the test.

PR #516, opened 2016-10-18
Fix crash when no external storage is available

Blaisorblade:
Close http://crashes.to/s/c99c9d8a9c5. But this might lead to duplicated
assets if the external folder becomes unavailable.

This is a new PR for #421 — I closed that accidentally when I moved the branch to my fork—consider this an issue, we're still not sure this is the correct approach.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1442,7 +1442,11 @@ public class Pokefly extends Service {
     }
 
     private void initOcr() {
-        String extdir = getExternalFilesDir(null).toString();
+        File externalFilesDir = getExternalFilesDir(null);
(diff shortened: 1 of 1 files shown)


DJCrapsody:
Should also fix the issue with Android 7 devices

Blaisorblade:
For now I consider it as orthogonal, but never had time to look at it closely, so not sure how to fix that.

NightMadness:
LGTM. 
However, there is an edge case where user will have files at both locations (like while the sd card is out start GoIV then put back card)

Blaisorblade:
@NightMadness exactly, but I'm not sure how to fix that.
Maybe it'd be good if the file was in some cache, so that cleaning caches would fix that? But an Android expert might have a better idea. @harkin ? @thearaks?

Blaisorblade:
Merging meanwhile.

PR #517, opened 2016-10-19
Translate custom clipboard (fr)

Remiscan:
I had some time to waste today, so, as usual, I wanted to translate the new strings to French. However, most of them were hardcoded, so I tried to make them translatable. That's what the first 4 commits are about.

Then I noticed the French for "add" didn't fit inside the "add" button, and the layout was a bit weird, so I tried to change it and played with the layout a bit in the last commit. Here's what it looks like on a Newus 6P and a Nexus 5 : http://imgur.com/a/B2XRY - there are probably still problems with the layout on smaller screens. I couldn't test but I wanted to show you what I have so far. There's also a weird issue when you fill one line in the "edit" and "preview" fields, on both the current and my modified layout (something like this happens: http://i.imgur.com/Rc10iFP.png).

Anyway, about the translation -- there are two things I couldn't translate:
- in almost every file from the /java/com.kamron.pogoiv/clipboard/tokens/ folder, there's a line like `public String getCategory() {
      return "Pokémon Name";
  }` and I can't put a string in there without getting an error like "...is not abstract and does not override abstract method...".
- in PokemonNameToken.java, I'd like to translate the Pokémon names in the line `String pokeName = maxEv ? "Dragonite" : "Dratini";` but I don't know how to call strings from the pokemons.xml file.

I still don't know anything about Java and I'm doing this by trial and error so I hope I'm not doing too many stupid things. I'm having fun trying though :p


nahojjjen:
My device is not french & i dont speak french, so I cant really validate if the french is correct... But its true that your change from hardcoded to xml string pointers are something that's needed. It seems to work on my end, but i'd like someone else to also chime in and give an okay / review before i merge.

Blaisorblade:
Thanks a lot for giving it a try! No time for a full review yet, but noticed this:

> in almost every file from the /java/com.kamron.pogoiv/clipboard/tokens/ folder, there's a line like public String getCategory() { return "Pokémon Name"; } and I can't put a string in there without getting an error like "...is not abstract and does not override abstract method...".

It's hard to be sure without actual errors, but did you try changing that to 
`public String getCategory(Context context) {...}` to use the `context` to lookup the translated strings? That would explain it — oversimplifying, the error tells you "hey, `getCategory()` is declared and called without arguments, so we're missing a definition that takes no arguments"—abstract methods have no definition.

I'll also leave the layout review to somebody else.

> My device is not french & i dont speak french, so I cant really validate if the french is correct...

You needn't even mention that 😉—translations are the realm of their authors and I don't really review them either (well, except for Italian).

Remiscan:
I can't fix the whole layout by myself :/ Multiple textViews in the same LinearLayout will stay on the same line even if they reach the edge of the screen, so this http://i.imgur.com/Rc10iFP.png inevitably happens. I couldn't find a way to fix it, so I guess someone better qualified should take care of it.

Moreover, there isn't a lot of room for the Tokens view (particularly when the height of the Preview and Edit ones becomes bigger), so I guess the whole layout would need to be redesigned to avoid all these issues, which is beyond my ability.

Blaisorblade:
> I can't fix the whole layout by myself :/

Since I can't really review/fix layouts: is the current layout in this PR always better than the one before? If not (or you're not sure), I'd ask you to take that commit out (if you want, to a separate PR for somebody to take a look at), otherwise I can't merge the PR. If that's a problem let me know, I can pick the other commits with a couple more minutes...

> Yes, that's actually what gave me the error I quoted. Without 'context', I just get "cannot resolve symbol getString". With 'context', Android Studio gives me these errors:

Fixing that by trial and error is not easy, but one way would be to also change the declaration of `getCategory` in `ClipboardToken`, in all classes implementing it, and in all call sites. But it's not that easy, and I'm not even sure that's the best approach...

Remiscan:
> is the current layout in this PR always better than the one before?

This PR fixes the "add" button in French, and on my devices everything "glitches" the same way as before when you add too many tokens, so I'd say it's slightly better.

Blaisorblade:
So #529 also changes the GUI and the layout, so I guess I can't merge the last commit (without having reviewed in detail though). Without that commit I'd have merged the rest already...

Remiscan:
Last commit deleted :)

Blaisorblade:
@Remiscan Thanks—we still got a conflict with #529, if you have time to fix it I should be able to merge this more quickly 😉

nahojjjen:
Clipboard has changed quite a bit since this pr was issued, and I think I'd like to re-do how clipboardtokens are represented/described in the future, so I believe I should close this PR.

I'm sorry that I did not spend enough time to merge this pr, Im unfortunately having difficulties juggling GoIV and irl.

PR #520, opened 2016-10-20
Re-enable user corrections for eeveelutions

### LINKED ISSUE #518: Cache correction and nick name guess not working for Eevee
Hi Team,
Flareon96 will no longer self correct to Flareon even tough it is cached and guess correctly, instead it reads as Vaporeon.

I think we can change that when getCandyNameEvolutionCostGuess applies to more then one pokemon, we should not override cache/guess correction and me it the last priority.

nahojjjen:
Edited description:

The identification method which uses evolution cost and candy name now returns that it's unsure of the pokemon if it detects that the evolution line branches. (This currently only happens for eevee.) Because of this, is continues testing other guessing methods, and these guessing methods take user corrections into account.

edit: fixes #518

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
@@ -71,17 +71,23 @@ public class PokemonNameCorrector {
         }
 
         //3. check if there's a cached guess for the hash of the bitmap of the nickname
+        //TODO
+        // Cacheguess and userCorrections are two different caches, one checks for the bitmap cache, one
+        // checks for the scanned string cache, however the bitmap cache does not seem to save properly - I'm
+        // disabling this module until it's been fixed. - Though we could maybe remove this guessing module altogether.
+        /*
         if (guess.pokemon == null) {
             PokeDist cacheGuess = cachedCorrections.get(poketext);
             if (cacheGuess != null) {
-                return cacheGuess;
+                guess =  new PokeDist(cacheGuess.pokemon, cacheGuess.dist);
             }
-        }
+        }*/
 
         //4.  If the user previous corrected this text, go with that.
         if (guess.pokemon == null) {
             if (userCorrections.containsKey(poketext)) {
                 poketext = userCorrections.get(poketext);
+                guess = new PokeDist(pokeInfoCalculator.get(poketext), 20);
             }
         }
         //5. All else failed: make a wild guess based only on closest name match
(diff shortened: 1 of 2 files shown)


nahojjjen:
Explenation for deactivating the bitmap module:

From what I can tell from debugging and reading the code, the  remembering of user corrections happens outside the pokemon guessing class - which makes sense, since when the pokemon guesser is running before the user has a change to make any changes.  It actually takes place in pokefly. However, pokefly only saves user corrections related to the OCR'd nickname, and does not save any corrections related to the bitmap. So the bitmap cache never gets corrected by the user. 

Since pokefly currently does not have access to the bitmap cache object as far as i can tell, I can not with my limited knowledge find a trivial way to change the code to make user corrections remembered in the bitmap cache.

Therefor, to fix eeveelution user corrections, the bitmap cache was disabled in this pull request.

nahojjjen:
I forgot a space, so checkstyle failed. :(

Blaisorblade:
TL;DR. can you just drop `cachedCorrections` and everything referring to it (field definition, `cacheResult`, ...)?

I should have done it in #502 since it stopped doing what it should have already before 65daf0327052fcee4a563420bea55ea638a800f7. That commit reorganized things and changed the comments, but the new comment
"//3. check if there's a cached guess for the hash of the bitmap of the nickname"
for the step with `cachedCorrections` is inaccurate—there's no bitmap being hashed.

The old comment was "//1. if there's a cached result for the nickname, return that", and that's more accurate (though maybe confusing).

> Since pokefly currently does not have access to the bitmap cache object as far as i can tell, I can not with my limited knowledge find a trivial way to change the code to make user corrections remembered in the bitmap cache.

Well, I'm not sure one should try. The goal of`cachedCorrections` was to save the result of all other loops computing a pokemon from the nickname (and nothing else). Right now we have fewer loops for that; at best, `cachedCorrections` could cache the result of `getNicknameGuess`, and I'm not sure it's worth it.

nahojjjen:
Was this what you had in mind?

app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
```diff
@@ -19,19 +18,19 @@
...
-    /* We don't want memory usage to get out of hand for stuff that can be computed. */
-    private final LruCache<String, PokeDist> cachedCorrections;
 
     public PokemonNameCorrector(PokeInfoCalculator pokeInfoCalculator, Map<String, String> storedUserCorrections) {
         this.pokeInfoCalculator = pokeInfoCalculator;
         userCorrections = new HashMap<>(pokeInfoCalculator.getPokedex().size());
         userCorrections.putAll(storedUserCorrections);
-        userCorrections.put("Sparky", pokeInfoCalculator.get(132).name);
-        userCorrections.put("Rainer", pokeInfoCalculator.get(132).name);
-        userCorrections.put("Pyro", pokeInfoCalculator.get(132).name);
```
Blaisorblade: Did you mean to drop changes to userCorrections? If not, as I guess, would you restore them?

Blaisorblade:
Apart from that regression, everything fine.

app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
nahojjjen: I actually removed it because it kept correcting my rainer to eevee, even though it's a vaporeon. I should have discussed the change before removing, and I should have kept it in a seperate commit.

My thought was that people will want to know the iv's of an eevee before they decide whether to make it a vaporeon or not, so I think its more likely that a pokemon named "rainer" is an already evolved vaporeon, not an eevee.

Blaisorblade:
Thanks for fixing and merging!

Issue #358, opened 2016-09-03
"CP354" recognized as "CP3 54"


rpgrca:
(Virtually) Latest code is misreading Nidoran CP354 as "CP3 54", so it always shows as 10 in the GoIV screen.

![screenshot from 2016-09-03 19-11-14](https://cloud.githubusercontent.com/assets/15602473/18227878/654b8074-720a-11e6-8a0b-89bf5ac676aa.png)

Not sure how a white space managed to get there, since it isn't in the tesseract's white list...

Blaisorblade:
Tesseract does that constantly. The code for HP handles this better I think by filtering non-digits (rather than cutting away two characters).
I've been afraid to reuse it because, well, these are always heuristics and you're never sure it'll be an improvement. Specifically, "ZP 354" would become "2354".

sarav:
I think we should throw out all spaced from all the OCR text and then do any comparisons. Including pokemon name, candy name, etc. We can remove the spaces in a similar way when we create the pokemap so that the look up continues to work. That should remove a lot of errors IMO. Including this one.

nahojjjen:
@sarav i agree that no ocr should actually need spaces, except for maybe candy recognition, where the space might be helpful in determining where the "candy" portion begins.

Blaisorblade:
But right now Candy is often split into two words. IIRC this is fixed because "Pikachu Can" is still closest to Pikachu.

I've thought of hardcoding the word for "Candy" in the languages we support, though I'm wary of that (right now GoIV works for unsupported languages). Maybe that's the way to go?

Or we could just keep "Candy" in there and hope that the closest thing to "PikachuCandy" is "Pikachu"—but maybe "MewCandy" seems as close to "Dewgong" as to "Mew" (distance = 5), so this isn't necessarily robust.

Maybe we should just log centrally all scanned strings (from users or at least from us) and try new heuristics on those.

sarav:
Having a unique list of user OCR results would be super helpful! Can crashalytics help there?

But we might get a lot of junk because of renames.

I'm tempted to say we break languages we don't support or add a setting to get number of characters in Candy name. Then we can just chop off that many characters. Any idea on the user stats on locales? Wonder how they'd even use it if the text is all in English anyway.

rpgrca:
Another for the backlog: Pidgey, CP300. This one is tricky, if the brightness is high, it gets correctly detected, but if it is not as bright, it gets detected as CPS00.

[...]

Issue #297, opened 2016-08-30
Can't scan CP on 2000 CP Vaporeon


Nizzem:
Tried to scan this Vaporeon on my Oneplus One phone, trainer level 25. CP always showed as 10. I was able to scan CP on the previous level and the level after this.

![screenshot_20160829-222325](https://cloud.githubusercontent.com/assets/20824577/18074610/e1b45a46-6e3b-11e6-9b92-a776a42679a5.png)

SteinerKD:
This happened to me as well with a 2021 Gyarados. First time it showed correctly, then I clicked the top stat for it after which it said there was no combinations for it and CP was locked as 10.
After viewing another Pokémon and going back it showed correctly again (I have not been able to reproduce this though).

Blaisorblade:
@SteinerKD the CP shouldn't be locked—the OCR can totally screw up but it should always be possible to edit it by hand (and if not, that's a bug I'm sure we can fix).

This other thread misreads a CP 200 pokemon (not clear if only that). I'm not convinced there's a relation but I'm watching for it.
https://www.reddit.com/r/GoIV/comments/50ex8f/bug/

SteinerKD:
In my case I couldn't edit the 10cp showing, but maybe I did something wrong and as I can't reproduce the error I can't verify that statement. If I ever get it to happen again I will try more things and try to document it somehow.

rpgrca:
I can confirm I can't OCR a 200CP Pidgey, it always goes to 10CP. It would look as if the OCR reader can't just detect the "200" combination correctly.

Blaisorblade:
@rpgrca Thanks a lot! We just need to replace Z with 2 when we expect numbers, which we already do elsewhere (copy-paste *is* bad). I've pushed to master a fix doing this, and it fixes the problem on my CP200 Seel (hadn't noticed I had one)—do you want to give it a try?

rpgrca:
Yep, it work, thanks.

SteinerKD:
Still happening for me with 2021 Gayrados (says 10cp) and 1380 Kingler (says 1cp).
Both have their IV % in their name, might that affect things? (GoIV 3.1.0)

rpgrca:
[...]

PR #521, opened 2016-10-20
Reverts change to how scanned cp text is parsed

nahojjjen:
This implementation assumes that the tesseract manages to scan "cp" or
"7p" or something similar, which is at least always two characters.

The reverted implementation removed all non-numerical characters, which
did not always work as sometimes "cp" was scanned as "c9" or similar.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -453,6 +453,12 @@ public class OcrHelper {
         tesseract.setImage(cp);
         String cpText = tesseract.getUTF8Text();
 
+        /*
+         * Always remove the two first characters instead of non-numbers: the "CP" text is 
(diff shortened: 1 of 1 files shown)


Blaisorblade:
Thanks for taking care of this! But what about the commit I added?

The old code called `fixOcrLettersToNums` (9d3e716766b44a74ff5eed97d1449bd0ce5551c6), and I assume we still need it (after dropping the first two characters) if numbers are scanned as letters.

BTW, with the new version of `fixOcrLettersToNums` we should fix #358. And readding the call should restore the fix to #297.

Blaisorblade:
And here's the rationale so people like me won't break it 😉 
Beware: I committed both things through the web interface so they're untested yet.

Blaisorblade:
@nahojjjen My "review comments" are in the new commits. Would you like to take this past the finish line?

PR #523, opened 2016-10-22
small refactor of PokeSpam

harkin:
Small refactor of the `PokeSpam` class. 

Made the constants static and changed to follow the normal naming convention. 

Removed the optional parameter from the second constructor and opted to chain them instead.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokeSpam.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokeSpam.java
@@ -1,7 +1,5 @@
 package com.kamron.pogoiv.logic;
 
-import com.google.common.base.Optional;
-
 import lombok.Getter;
 
 /**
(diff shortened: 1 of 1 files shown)


Blaisorblade:
LGTM, thanks! (BTW is the second constructor even used anywhere?)

harkin:
It isn't currently, figured it might be related to the Halloween event running soon but I don't really have context for it; not sure what a PokeSpam is exactly 😅

Blaisorblade:
Maybe that constructor could still be useful... leaving it alone is fine.

That feature computes how many pokemons you can evolve with the available candies, so you can say collect Pidgeys (or other 12-candy pokemons) to evolve with a lucky egg to level up. "PokeSpam" is a name a contributor picked for this contributor.

NightMadness:
Exactly, I added code for to have the option to calculate the extra bonuses in the future.

Like special events or if you trade your pokemon when you do a pokespam, usally it gives you one more candy (right now for halloween is 2).

PR #528, opened 2016-10-25
Fixes crop misalignments for pogo update

nahojjjen:
Also changes the cropping to a new wrapping method which is easier to
read & tweak


Blaisorblade:
Thanks for the refactoring, I like it. It's not clear which values are changed and which aren't, but if the values need to vary across phones, we'll need to collect screenshots from scratch anyway, so there's no reason to keep the old time-tested values.

PR #529, opened 2016-10-25
Clipboard UI upgrade

### LINKED ISSUE #526: Move Edit Clipboard Setting button to Settings screen
Can we add an entry in the Settings screen (after the "Copy to Clipboard" entry) that can be clicked to show the Clipboard settings activity?

I didn't realize we could scroll the Main activity to find a settings button down there.

nahojjjen:
Moves the clipboard configuration from the main activity to the settings

Tweaks several things in the clipboard configuration, it's messier but explains more.

Edit: fixes #526

### DIFF
--- a/app/src/main/res/xml/settings.xml
+++ b/app/src/main/res/xml/settings.xml
@@ -29,8 +29,16 @@
     <SwitchPreference
         android:key="copyToClipboard"
         android:title="@string/copy_to_clipboard_setting_title"
-        android:summary="@string/copy_to_clipboard_setting_summary"
+        android:summary="Automatically add scanned information to your copy &amp; paste clipboard."
         android:defaultValue="false"/>
+
+
+    <Preference
+        android:key="@string/clipboardButton"
+        android:summary="@string/clipboardButton_summary"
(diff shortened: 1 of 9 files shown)


Blaisorblade:
At first sight looks reasonable to me. Feel free to merge if ready.

app/src/main/res/xml/settings.xml
```diff
@@ -29,9 +29,17 @@
     <SwitchPreference
         android:key="copyToClipboard"
         android:title="@string/copy_to_clipboard_setting_title"
-        android:summary="@string/copy_to_clipboard_setting_summary"
+        android:summary="Automatically add scanned information to your copy &amp; paste clipboard."
```
NightMadness: We should probably propograte it back to strings.

Blaisorblade: + 💯  thanks for spotting this. @nahojjjen could you please review your PRs to avoid this before declaring them as ready to merge?

Blaisorblade:
@nahojjjen Thanks for the fix!

PR #531, opened 2016-10-26
Removes hardcoded "user" corrections

nahojjjen:
I dont think these hardcoded suggestions help the users more than they can potentially annoy users, now with the new identification module.

They're initiated every new restart of goiv
They assume that rainer, pyro and sparky are eevees, not vaporeon, flaroeon and jolteon.


Blaisorblade:
Ah I see, makes sense to me! Thanks!

PR #537, opened 2016-10-29
Fixes #463 Crash when scanning screenshot from another device.

### LINKED ISSUE #463: Crash when scanning screenshot from another Android/iPhone phone
Hi Team,
A friend of mind send me a screenshot from an Iphone to check IV, I open the text message, clicked on share then GoIV, GoIV crashed.
Took same screenshot, took a screenshot of the screenshot using google now, sent to GoIV, calculated successfully.

Could this be bug #420 ? 

I have a galaxy S7 edge, stock, not rooted.

Example screenshot that fails
img951676 _Level 22 Cloyster_

also, in order to scan I must enable powersave mode, even if i share the screenshot using the android share button. is this a bug?

NightMadness:
Hi Team,
below is a check for GOIV screenshot, I was able to scan images from iphone.

code:

```
// this should allow processing of images where the displaymetrics don't match, for example a different
// phone, it is known that the red dot might not display correctly.
if (displayMetrics.heightPixels != bitmap.getHeight() || displayMetrics.widthPixels != bitmap.getWidth()) {
     bitmap = Bitmap.createScaledBitmap(bitmap, displayMetrics.widthPixels, displayMetrics.heightPixels,
              false);
}
```

Known issues:
Red dot location MIGHT be off if the image height is different
since red dot is based on the screen, and since the photo display software (andorid photos etc.) probably stretches the display to fill width of the screen that red dot might be located elsewhere especially if height of the screen is significantly different (see https://github.com/farkam135/GoIV/issues/463#issuecomment-250882817 for example), most of the time the height will be the same and the red dot will work. It might be easier to recalculate the metric's when https://github.com/farkam135/GoIV/pull/499 is complete.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1822,6 +1822,14 @@ public class Pokefly extends Service {
             } else {
                 screenShotPath = Optional.absent();
             }
+
+            // this should allow processing of images where the displaymetrics don't match, for example a different
+            // phone, it is known that the red dot might not display correctly.
+            if (displayMetrics.heightPixels != bitmap.getHeight() || displayMetrics.widthPixels != bitmap.getWidth()) {
+                bitmap = Bitmap.createScaledBitmap(bitmap, displayMetrics.widthPixels, displayMetrics.heightPixels,
+                        true);
+            }
+
             scanPokemon(bitmap, screenShotPath);
             bitmap.recycle();
         }


nahojjjen:
So this ensures that no weird out of bound cropping or similar occurs, by forcing the scanned image to be the same size as the android screen? I suppose it makes sense. It's also reasonable for the user to assume that there will be aligning errors, since it's from another phone.
 
I think the PR makes sense, anything you'd like to comment or should I merge?

NightMadness:
Nope, we can merge

PR #538, opened 2016-10-30
Restart GoIV after choosing level

NightMadness:
This PR will Restart GoIV when selecting level.
Once #499 is finished we can reuse commit https://github.com/farkam135/GoIV/commit/656e8e7f66f5e9a66b15cba730b236af086348d5 code to change the Trainer Level (since we will not stop/start we don't need to start Pokemon Go App)


nahojjjen:
Sorry for the late reply.

I like the idea, but would it be possible to keep the new class outside of MainActivity, to keep it from bloating?

NightMadness:
Sure will do

Issue #535, opened 2016-10-28
Testing results on master + PR #503, TODOs for release


Blaisorblade:
Since we're nearing a release, I took a look at the current state.

Noticed a few odd things, which I think have nothing to do with #503:
- Some user corrections were off (probably was my doing), I had a Vaporeon nick mapped to Spearow. Given the code, I must have done that at some point, and forgot when user corrections weren't used as much.
- Eevolutions can be mapped to random pokemons because they're recognized based on the nickname, without using the candy family. That's as expected from what I proposed, it's hard to fix, and gets worked around via user corrections.
- A Nidoking was mapped to Nidoqueen, but that didn't happen after relaunching and the code for Nidoran gender guessing is there also for candies.

I also don't know
- what happened with Android 4.4? @thearaks @DJCrapsody . If the bugs aren't fixed yet, we could include a warning in the changelog.
- IIRC there were a few other bugs which we could fix like #477.

DJCrapsody:
The crash on 4.4 was fixed in #506, and was merged quite some time ago.

nahojjjen:
> Some user corrections were off (probably was my doing), I had a Vaporeon nick mapped to Spearow. Given the code, I must have done that at some point, and forgot when user corrections weren't used as much.

I think a lot of nickname corrections were mapped wrong when the candy was cropped partially wrong, I cant reproduce this, so I dont think it's a problem.

Blaisorblade:
OK, sounds like we should close this (I've also seen work on Eevolutions). I wasn't sure whether to file the issue in the first place, but last time I had such an issue it turned out IIRC I should have done a report to be sure.

PR #539, opened 2016-11-03
Guess eevelotion based on candy

NightMadness:
Hi Team,
This should fix Eevolutions part of #535 
1. https://github.com/farkam135/GoIV/commit/771455b22c5207d3135f89ecd5234425697646ff added get the Pokemon with the closest name within the evolution line (for eevees)
2. https://github.com/farkam135/GoIV/commit/51542c199d14aafe89e15c6c228d5e8b9c324f39 put back user correction during Pokemon check (removed at https://github.com/farkam135/GoIV/pull/531), this does not override user correction.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
@@ -42,8 +42,9 @@ public class PokemonNameCorrector {
      * 1. check if the nickname perfectly matches a pokemon
      * 2. check if candyname + evolution cost perfectly matches a pokemon
      * 3. check if there's a stored user correction for the scanned pokemon name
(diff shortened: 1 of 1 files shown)


nahojjjen:
Im testing this but I'm unsure if I don't understand what it's supposed to do, or if the functionality is disrupted because I have "left over" user corrections remembered on my phone.

How does this change the behaviour on eevee scans? What is supposed to have changed?

NightMadness:
If there is a flareon but is read as f1ereon it will correct it based on bess guess

NightMadness:
It will guess based on eevee line since we know its a type if eevee. Also if we have an flareon called 98% it will take a wild guess based on eevee evolved line (we know from the candy cost it cannot be an eevee)

nahojjjen:
Seems reasonable. I'll merge it for now, and if issues pop up we can fix it in the future. :)

Issue #477, opened 2016-09-24
Unknown crash due to view being null


DJCrapsody:
Not sure what is causing the crash. Could not reproduce, but seems like many are facing this.

> Fatal Exception: java.lang.RuntimeException: Unable to stop service com.kamron.pogoiv.Pokefly@6616382: java.lang.IllegalArgumentException: view must not be null
>        at android.app.ActivityThread.handleStopService(ActivityThread.java:3101)
>        at android.app.ActivityThread.-wrap21(ActivityThread.java)
>        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1448)
>        at android.os.Handler.dispatchMessage(Handler.java:104)
>        at android.os.Looper.loop(Looper.java:148)
>        at android.app.ActivityThread.main(ActivityThread.java:5460)
>        at java.lang.reflect.Method.invoke(Method.java)
>        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
>        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)
> Caused by java.lang.IllegalArgumentException: view must not be null
>        at android.view.WindowManagerGlobal.removeView(WindowManagerGlobal.java:346)
>        at android.view.WindowManagerImpl.removeView(WindowManagerImpl.java:111)
>        at com.kamron.pogoiv.Pokefly.unwatchScreen(Pokefly.java:448)
>        at com.kamron.pogoiv.Pokefly.onDestroy(Pokefly.java:497)
>        at android.app.ActivityThread.handleStopService(ActivityThread.java:3082)
>        at android.app.ActivityThread.-wrap21(ActivityThread.java)
>        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1448)
>        at android.os.Handler.dispatchMessage(Handler.java:104)
>        at android.os.Looper.loop(Looper.java:148)
>        at android.app.ActivityThread.main(ActivityThread.java:5460)
>        at java.lang.reflect.Method.invoke(Method.java)
>        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:726)
>        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:616)

sarav:
Seems to be due to something in the unwatchScreen(). 
If we can reproduce this issue, try adding these line (they look like pointless lines from an example code I saw, so I dropped, but maybe it matters) right after touchView is created:

[...]

PR #540, opened 2016-11-03
changed onStartCommand with intent to START_REDELIVER_INTENT, added error for null intent (Testing needed)

NightMadness:
Hi Team,
below possibly fixes https://github.com/farkam135/GoIV/issues/477 (to address #535), Testing is need, can someone reproduce the bug?

If we have an null intent we throw an error early so we don't need to dig to the log, conversely we can handle nulls for on destroy.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -419,7 +419,9 @@ public class Pokefly extends Service {
     @Override
     public int onStartCommand(Intent intent, int flags, int startId) {
         if (intent == null) {
-            return START_STICKY;
+            //We should not reach here with no intent, this wil error later due to view being null so we throw error now
+            //https://github.com/farkam135/GoIV/issues/477
+            throw new java.lang.IllegalArgumentException("No intent found.");
         }
 
         running = true;
@@ -443,8 +445,9 @@ public class Pokefly extends Service {
                 screenShotHelper = ScreenShotHelper.start(Pokefly.this);
             }
         }
-
-        return START_STICKY;
+        //We have intent data, it's possible this service will be killed and we would want to recreate it
+        //https://github.com/farkam135/GoIV/issues/477
+        return START_REDELIVER_INTENT;
     }
 
     private void watchScreen() {


nahojjjen:
I haven't been involved enough with this part of the code to make a good review, I think blaisorblade handled this? I'm not sure.

Blaisorblade:
Just noticed this—it fits with what @sarav proposed in https://github.com/farkam135/GoIV/issues/477#issuecomment-249744984 and IIRC the idea made sense. So thanks @NightMadness for the PR. BTW, I've seen you've been rather active recently with good submissions, thanks for that too!

To exercise the change, I'd recommend adding a log statement in `onStartCommand` (temporarily) and try loading other apps to fill memory, to see what happens when GoIV is brought out of memory. Specifically, to have a chance of reproducing #477, I'd try 
- start the service
- get it killed by opening lots of apps
- close the other apps, and hope Android restarts the service
- shut it down via the button in MainActivity

You might even want to try this before the patch (to reproduce the crash) and after (to confirm it's gone), because I'm not actually sure that sequence of events is the correct one.

But overall, unless testing finds some regression, I'd tend to merging this.

NightMadness:
I was able to repreduce while causing a crash in pokefly then clickin on stop on the notification, which is not 100% what we want to test but is close. 

Let me know if you are able to repreduce the bug.

Which reminds me stop in notification should probably changed to skip oncreate and ondestory when executing intent.

Blaisorblade:
> Which reminds me stop in notification should probably changed to skip oncreate and ondestory when executing intent.

I'd guess it shouldn't and can't skip `Pokefly.onDestroy`. Is it running `Pokefly.onCreate`?

NightMadness:
### tl;dr
This pr should handle low memory cases better
We need a separate PR/Bug to handle the "stop" notification bug

--------------
if we have low memory android might try try to start the service then go to 
- **onStartCommand** --> **Null intent** (skipped initialization)
- **stop service** --> **onDestory** --> we get an error since we didn't initialize the view.

I'm not sure where onCreate comes in the cycle...

### one way to reproduce a similar the bug using notification
1. Crash the PokeFly service (any bad code will do).
2. Go to the notification and click on "Stop".
3. We see a new crash.

### to fix it 
1. We change to START_REDELIVER_INTENT (this PR) for restart with low memory cases 
2. stop notification is a different issue could be fixed with a new PR/bug.

Blaisorblade:
I think merging this PR is better than not. On this branch I've renamed quite a few pokemons, tried loading other apps (I think without getting GoIV unloaded), without any problem.

An issue about the extra crash might be good (with the stacktrace), but if the extra crash happens only after a crash that's probably *not* the worst thing (it's not even clear whether we want to fix that). I see the goal was to test what happens if the service is closed instead... but I guess, after this PR, the remaining bug only happens after a crash.

PR #542, opened 2016-11-04
Add share button to result screen

nahojjjen:
This pull request adds a share button at the result screen, which allows other applications to recieve a string formated Json blob intent with the pokemon information. Currently, Storimod can support recieving pokemon in this way.


nahojjjen:
Picture showing the share button:

image

NightMadness:
1.I like the idea, I prefer a plugin system but this could be a start.
2. Can we factor some of the stuff out of pokefly. 
3. Add setting to disable (maybe enable share button if supported app is installed).
4. Remove some of the toasts :)

nahojjjen:
>1. I like the idea, I prefer a plugin system but this could be a start.

A plugin system has been discussed, and is a lot of work, I currently dont have the time for such a large project. Im doing this meanwhile.

2. Can we factor some of the stuff out of pokefly. 

This is a constant struggle, we've made pokefly responsible for waaay too much. :) Ill try to move the sharing logic & making the json blob into a separate class.

3. Add setting to disable (maybe enable share button if supported app is installed).

I thought about this, but I personally think the share button is minimal enough to not be too distracting for a normal user, and the show/hide logic checking on the settings would bloat the pokefly class.

4. Remove some of the toasts :)

Will do

NightMadness:
:)

BraisGabin:
What about a more material icon? https://material.io/icons/#ic_share

nahojjjen:
I suppose a more material icon would need to be added as a drawable?

BraisGabin:
Yes. You can add it as a vector drawable.

NightMadness:
Formula for unique pokemon: 
[Name], [Height], [Weight], [List of all possible IV's]

Then we can compare 
I have a pidgey, with height 10 and weight 10, with possible IV 15, 15, 15 or 15, 15, 14
I looks for all the pidgies with this height and weight and loop through the IV's maybe he has the same iv possibility.

List of Possible IV's can be simple JSON or better yet for a unique id we encode each IV using some formula:
we should have 3375 possibilities

I tried to come up with formla but its beyond my math.
Google says we could use Cantor pairing function twice or some other mathematical function that represent 3 integers combination

We can send it as an array or better yet
now we add bits to a binary number for each IV possibility
could look like this  (we should see 3375 zeros) 00000100000000010000000000
now we a convert this binary to a string and we have a unique representation.

then we can just check to see if this bit (for an IV combination) exits when we scan

nahojjjen:
@NightMadness Im not sure what you're saying, are you suggesting I change the uniqueId to use a new algorithm?

NightMadness:
Yeah,  its impossible with 100% to know if its the same pokemon (especially after evolving) the only costants are IV and evelotion line.

NightMadness:
Rewrote my previous answers to make it more clear.

nahojjjen:
Ive discussed about the unique id with storimod people over skype, the motivation for the current implementation was basically that it has a quite high success rate and was very easy to implement :) The risk of two pokemon having the same weight and height are exceedingly low.

Your suggested change would from what I understand limit possible false-detections of duplicates by including extra information in form of IVs. The added information could also potentially allow storimod to detect which pokemon are evolutions. It would however require that weight and height were scanned as separate values, and that the storimod devs implement the extra logic.


tl;dr I guess it could be done, but I'm not sure it's enough of a problem to motivate the implementation time.

NightMadness:
Sounds reasonable

nahojjjen:
The previous commit made the image black

nahojjjen:
Is there anything anyone wants changed more or can I merge this?

PR #543, opened 2016-11-05
if we have no Candy Cost, we still should guess based on evolution line.

NightMadness:
This PR will 
1. move getBestGuessForEvolutionLine to only activate if we need to
2. if candy evolve cost recognition failed (we have an absent for candy upgrade cost), and we are sure about the evolution line, we will guess the name based on evolution line, this is better then a wild guess.


nahojjjen:
I think it'd be good to avoid double assignments for code readability

nahojjjen:
modified in #544

PR #545, opened 2016-11-07
Better notification Start, settings and +1 level  in notification bar

### LINKED ISSUE #504: Start/restart app in notification bar
This way the app can always be open but not always drain battery. 

When someone gets that singular mon they can immediately check with GoIV and name/transfer depending on results and then stop app again without it closing it or switching from PoGo to start again.

NightMadness:
## Better notifications
Fixes #504
1. Added stop button to notifcation (if we stopped from notification)
2. Added settings button. (Only when stopped)
3. Removed Open App button (only room for 2 icons), clicking on GoIV notification will open GoIV. (based on android docs we should not have 2 objects do the same thing)
4. added +1 trainer level button as requested in https://github.com/farkam135/GoIV/issues/268 (will only show if GoIV is started)

screenshot_20161108-143907

Known issues:
1. Notification is destroyed then a new one is created:  Notification is tied to service (to make it a foreground service), so when selfstop is fires it will kill the notification. At api 24 stopForeground(STOP_FOREGROUND_DETACH) should fix that but I don't have API 24 to test.
2. if we are in screen capture mode, start button in notification will open the main activity screen first, then start GoIV, (if you have open pogo option it will start pogo.)

I think we can live with those limitations for now.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/MainActivity.java
+++ b/app/src/main/java/com/kamron/pogoiv/MainActivity.java
@@ -50,8 +50,10 @@ import timber.log.Timber;
 public class MainActivity extends AppCompatActivity {
 
     public static final String ACTION_SHOW_UPDATE_DIALOG = "com.kamron.pogoiv.SHOW_UPDATE_DIALOG";
-    public static final String ACTION_CLICK_ON_BUTTON = "com.kamron.pogoiv.ACTION_CLICK_ON_BUTTON";
-    public static final String ACTION_START_SETTINGS = "com.kamron.pogoiv.ACTION_START_SETTINGS";
+    public static final String ACTION_START_POKEFLY = "com.kamron.pogoiv.ACTION_START_POKEFLY";
+    public static final String ACTION_RESTART_POKEFLY = "com.kamron.pogoiv.ACTION_RESTART_POKEFLY";
+    public static final String ACTION_INCREMENT_LEVEL = "com.kamron.pogoiv.ACTION_INCREMENT_LEVEL";
+    public static final String ACTION_OPEN_SETTINGS = "com.kamron.pogoiv.ACTION_OPEN_SETTINGS";
 
     private static final String TAG = MainActivity.class.getSimpleName();
     private static final int OVERLAY_PERMISSION_REQ_CODE = 1234;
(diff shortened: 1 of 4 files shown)


NightMadness:
Unrelated findings:
1. We might want to modify intent notification flags to FLAG_ACTIVITY_SINGLE_TOP https://developer.android.com/reference/android/content/Intent.html#FLAG_ACTIVITY_SINGLE_TOP
2. looks like selfstop sends back an TransactionTooLargeException exception.

app/src/main/java/com/kamron/pogoiv/MainActivity.java
```diff
@@ -439,4 +440,20 @@ private void startScreenService() {
                 Context.MEDIA_PROJECTION_SERVICE);
         startActivityForResult(projectionManager.createScreenCaptureIntent(), SCREEN_CAPTURE_REQ_CODE);
     }
+
+    /**
+     * Handles custom action intents probably from notification.
+     */
+    @Override
+    public void onNewIntent(Intent intent) {
```
NightMadness: Noticed this doest fire when starting a new instance of the activity (when GoIV settings screen is killed)

NightMadness:
Code ready to merge.
added +1 button as requested in https://github.com/farkam135/GoIV/issues/268

nahojjjen:
Okay, Ive tested this during the day and it seems to be working very well.

My three comments are
 1. it's slightly annoying that the goiv main activity pops up on start
 2. It feels like the notification should be expanded by default - some users dont know that notifications can be expanded
 3. The code is slightly messy- but this does not seem to be your fault, but simply caused by that pokefly and mainactivity are a little messy.

All of these are minor comments, so I'll go ahead and merge this. :)

NightMadness:
1.it's slightly annoying that the goiv main activity pops up on start
Yes i agree. There is no easy way to fix, ill keep checking maybe i can overide the activity manager to skip killing that object.

2. It feels like the notification should be expanded by default - some users dont know that notifications can be expanded
I was able to get this to auto expend by putting priority.max but i think its a bit excessive to make it max priority but thats a fast change if we want. Also we can use media control icons which are visible during collapsed but no text, only icons.

3. The code is slightly messy- but this does not seem to be your fault, but simply caused by that pokefly and mainactivity are a little messy.
Yeah, I tried to reuse componenets as much as I could maybe thats not the best strategy.

PR #547, opened 2016-11-09
Fixes hp scan adding 5 to certain languages

nahojjjen:
In some language, hp is "PS", and the previous implementation
consistently translated the S to a 5 and added it to the scan.


BraisGabin:
My 2 cents: `[^/]+/([0-9 ]+) [^ ]+` and if the previus one fails: `[^/]+/([0-9 ]+).{2}`.

nahojjjen:
I agree that it'd be better to have a solution which would also handle the case where "99 P5" was scanned.

@NightMadness I think your solution would be good, but is a very specific fix for PS, I'd like to have a more general fix if possible. This fix would also not work if P was scanned wrongly.

@BraisGabin I'm not good enough at regex to understand that, could you explain what it does?

Another fix could be to remove the letters after a space, regardless of what it is. So after the split on "/" has happened, we're left with something like 99 XY, where we dont know what XY is. By removing anything after the space, we'd be left with 99. The problem with this approach is that it relies on that the space is detected.

Blaisorblade:
@BraisGabin's regex, `[^/]+/([0-9 ]+) [^ ]+`, also cuts away the part up to the first (only) `/` (which we do before), then accepts at least one (or more) digits or spaces (`[0-9 ]+`), and then cuts away a space and a sequence of non-spaces. So this still requires that the space is detected.

BraisGabin:
Yes, the first regex relies on detecting the space. The space is normally there. But the times that it isn't I use this other regex: `[^/]+/([0-9 ]+).{2}`. It's basically the same but I remove the last two characters `.{2}`.

nahojjjen:
I didn't add your exact regex to the code, but I did try and replicate the behaviour which your regex described.

This is just because I'm not comfortable with dealing with regex, and it looks like we're gonna need to push this fix soon.

NightMadness:
Yeah you can replicate the split to delete after the space without regex

NightMadness:
@BraisGabin  I couldn't get the regex you suggested to work on online regex calculator https://regex101.com/r/wHuzeR/1
Can you come up with regex that applies for most scenarios, maybe handle the / split as well. 


#### text examples:
```
2SS 100 PS
2SS100PS
100 / 100 PS
100 / 100 HP
100 / 1s0 HP
100 /10S hp
100/10S HP
100/10SHP
100/10SPS
100 PS
100PS
100HP
```
#### expected result 100
@nahojjjen i think since its urgent we fix it this is my proposal for the fix
```
//Split on Space, if we found space, take the first part
hpParts = hpStr.split(" ");
if (hpParts.length > 0) {
 hpStr = hpParts[1];
}
```

Blaisorblade:
@NightMadness FWIW regarding that regexp: that calculator seems to have a different syntax, but fixing the warnings as suggested gives `[^\/]+\/([0-9 ]+).{2}` (I'm not sure what syntax variant Java uses, but it's described by the docs).
Problem: it's hard for a regexp alone to convert an `S` to a `5` (your examples call for replacing an `S` with a `0`, but that sounds wrong); if that's possible, I'd still prefer doing something else.

Blaisorblade:
@NightMadness also, the code snippet you show is buggy:
```
hpParts = hpStr.split(" ");
if (hpParts.length > 0) { // This must be > 1, or >= 2
 hpStr = hpParts[1];
}
```

If `hpParts.length == 1` then `hpParts[1]` crashes.

nahojjjen:
The current code on this branch is from what I can tell very similar to the regex brais suggested, and from my manual testing it seems to be working. Could anyone confirm?

NightMadness:
@nahojjjen sorry my mistake for some reason the show changes showed the wrong code your commit is perfect!

BraisGabin:
As @Blaisorblade said, the regex doesn't clean the data. It just extract it. You must perform all the `replace()` before. And, in my code, I couldn't fit all the cases in just one regex. I use two:
1. `[^/]+/([0-9 ]+) [^ ]+` This one is more secure.
2. `[^/]+/([0-9 ]+).{2}` If the first one fails I must go with this one that have some issues, but is the best I can. How to improve this? Train Tesseract with the correct font.

PR #548, opened 2016-11-10
Minor changes to Notification 

NightMadness:
1. Changed plus button to one that is closer color scheme of start button.
2. Added notification sub text "Tap to open GoIV."
3. Naming convention fixes for stings
4. Change visibility to public so notification can be viewed from lock screen
5. Change priority to high, as long as we are the first notification, it will be opened expended (max priority notification are higher priority)


Blaisorblade:
LGTM!

Blaisorblade:
BTW cool that you added the +1 level (in #545), just got to try that out!

PR #549, opened 2016-11-11
Minor changes to #547 (directly to connectStorimod branch)

NightMadness:
**Reverted:** _1. Switch string to Optional<String>_
2. Removed pokemonUniqueID from IVScanResult
3. Added log to PokemonShareHandler
4. Make APPLICATION_POKEMON_STATS constant

@nahojjjen feel free to cherry pick

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -486,14 +486,14 @@ public class OcrHelper {
      * @param pokemonImage the image of the whole pokemon screen
      * @return a string which should remain the same even if you power up a pokemon
      */
-    private String getPokemonIdentifierFromImg(Bitmap pokemonImage) {
+    private Optional<String> getPokemonIdentifierFromImg(Bitmap pokemonImage) {
         Bitmap infoRow = Bitmap.createBitmap(pokemonImage,
                 (int)Math.round(widthPixels * .1f), (int)Math.round(heightPixels / 1.714286f),
                 (int)Math.round(widthPixels * .8f), (int)Math.round(heightPixels / 25.26316f));
         tesseract.setImage(infoRow);
         String uniqueText = tesseract.getUTF8Text();
 
-        return uniqueText;
+        return Optional.fromNullable(uniqueText);
     }
 
 
@@ -547,7 +547,7 @@ public class OcrHelper {
         Optional<Integer> pokemonCP = getPokemonCPFromImg(pokemonImage);
         Optional<Integer> pokemonCandyAmount = getCandyAmountFromImg(pokemonImage);
         Optional<Integer> pokemonUpgradeCost = getPokemonEvolutionCostFromImg(pokemonImage);
-        String pokemonUniqueIdentifier = getPokemonIdentifierFromImg(pokemonImage);
+        Optional<String> pokemonUniqueIdentifier = getPokemonIdentifierFromImg(pokemonImage);
 
         return new ScanResult(estimatedPokemonLevel, pokemonName, candyName, pokemonHP, pokemonCP,
                 pokemonCandyAmount, pokemonUpgradeCost, pokemonUniqueIdentifier);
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/kamron/pogoiv/logic/PokemonShareHandler.java
```diff
@@ -30,19 +33,19 @@ public void spreadResultIntent(Pokefly pokefly, IVScanResult ivScan) {
...
             jsonPokemon.put("Cp", ivScan.scannedCP);
-            jsonPokemon.put("uniquePokemon", ivScan.uniquePokemonID);
+            jsonPokemon.put("uniquePokemon", uniquePokemonID);
             jsonPokemon.put("estimatedPokemonLevel", ivScan.estimatedPokemonLevel);
             PokeInfoCalculator calc = PokeInfoCalculator.getInstance();
             jsonPokemon.put("candyName", calc.getEvolutionLine(ivScan.pokemon).get(0));
 
         } catch (JSONException e) {
-            e.printStackTrace();
+            Log.e(TAG, "Tried to generate jsonPokemon when clicking the share button",e);
```
Blaisorblade: Please use Timber for logging instead, so we get exception logged on Fabric.

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
```diff
@@ -486,14 +486,14 @@ private String getCandyNameFromImg(Bitmap pokemonImage) {
...
-    private String getPokemonIdentifierFromImg(Bitmap pokemonImage) {
+    private Optional<String> getPokemonIdentifierFromImg(Bitmap pokemonImage) {
         Bitmap infoRow = Bitmap.createBitmap(pokemonImage,
                 (int)Math.round(widthPixels * .1f), (int)Math.round(heightPixels / 1.714286f),
                 (int)Math.round(widthPixels * .8f), (int)Math.round(heightPixels / 25.26316f));
         tesseract.setImage(infoRow);
         String uniqueText = tesseract.getUTF8Text();
 
-        return uniqueText;
+        return Optional.fromNullable(uniqueText);
```
NightMadness: You are right we get empty sting for empty ocr. Optional is not better but its consistant with the rest of the code. Either way is fine with me.

nahojjjen:
1. Switch string to Optional<String>

Blaisorblade seems to be discussing this change with you

2.Removed pokemonUniqueID from IVScanResult

Sounds good, I was trying to have the uniqueID follow the same standards as hp and cp and other scanned info, but I suppose this also makes sense.

3.Added log to PokemonShareHandler

Good idea, though as blaisorblade said we probably want it to log to fabric, so we get online crashreports.

4.Make APPLICATION_POKEMON_STATS constant

Nice

app/src/main/java/com/kamron/pogoiv/OcrHelper.java
NightMadness: Sounds resonable, no need for it to be optional

NightMadness:
> 1. Switch string to Optional

> Blaisorblade seems to be discussing this change with you

Removed from this PR.

> 2.Removed pokemonUniqueID from IVScanResult

> Sounds good, I was trying to have the uniqueID follow the same standards as hp and cp and other scanned info, but I suppose this also makes sense.

I think this IVScanResult is mostly useful if we want to use an item with the clipboard or compare previous Scan. but it Its really up to you.

> 3.Added log to PokemonShareHandler
> 
> Good idea, though as blaisorblade said we probably want it to log to fabric, so we get online crashreports.

changed to fabric

> 
> 4.Make APPLICATION_POKEMON_STATS constant
> 
> Nice

Blaisorblade:
Re pokemonUniqueId I just realized: do we use it to check when it makes sense to allow "Refine" on scan results? From a superficial look, it seems we don't but maybe we should. Would that be easier if we keep `pokemonUniqueId` in `IVScanResult`?

But otherwise I agree the ID doesn't belong there too much...

nahojjjen:
We cant use pokemonUniqueId in the situation where the pokemon is evolved, since that can change type, height and weight.

PR #552, opened 2016-11-12
Update icons in notification

nahojjjen:
Switches the icons for play / pause and add in the notification to material design ones.

This changes the "Stop" icon to a "pause" icon, which I motivate with that we've changed the behaviour of the button since the stop icon was picked, and that button actually just pauses GoIV now.


nahojjjen:
image

image

NightMadness:
Nice! This is much better

PR #553, opened 2016-11-13
New Feature to add a toast when text is copied to the clipboard.

### LINKED ISSUE #551: New Feature: Add a message/popup/notification when text is copied to the clipboard.
What do you guys think of adding a new feature to let the user know with a message (could be toast, open to ideas) when clipboard is copied with what was copied.

Something like "`Copied to clipboard: Pid151515.`".
if there is room we can instruct on how to disable the clipboard setting.

I think this will also ease the concerns of #511, at least will let user know we are "taking over" the clipboard.

NightMadness:
Fixes #551
1. Added **short** toast message in "Center" gravity.
2. Changed default setting to `Name (3 char)` + `Min-Max` + `Unicode not filled (MAX IV)`
3. Removed some debug log messages.

unnamed

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
+++ b/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
@@ -3,12 +3,13 @@ package com.kamron.pogoiv;
 import android.content.Context;
 import android.content.SharedPreferences;
 import android.os.Build;
-import android.util.Log;
 
 import com.kamron.pogoiv.clipboard.ClipboardToken;
 import com.kamron.pogoiv.clipboard.tokens.IVPercentageToken;
 import com.kamron.pogoiv.clipboard.tokens.IVPercentageTokenMode;
+import com.kamron.pogoiv.clipboard.tokens.PokemonNameToken;
 import com.kamron.pogoiv.clipboard.tokens.SeparatorToken;
+import com.kamron.pogoiv.clipboard.tokens.UnicodeToken;
 
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


nahojjjen:
("min""-""max") default was implemented so users who upgrade dont suddenly have their clipboard settings altered.

NightMadness:
@nahojjjen - I am not as concerned with this change in the update. I think that the users may find the change useful because we will be letting them know the defense, attack, and HP stats in addition to the average IV and will highlight the cool Unicode feature. Since you created this code, I respect any decision you think is best.

Blaisorblade:
Both make sense — but current users didn't have a choice.
Nevertheless, what about having min-max instead of average in the new default? I might be biased because that's what I use, so of course I do think it's a reasonable default ;-)

NightMadness:
Sounds reasonable: `Name (5 char)`+ `MIN-MAX` + `Unicode not filled.`.

Blaisorblade:
I think you want `name (3 char)` then, so it fits

NightMadness:
Changed to Name (3 char)

Blaisorblade:
Good for me. Assigning to @nahojjjen hoping he's fine with the change of default.

nahojjjen:
I was just explaining my reasoning for why min-max was implemented, I'm certainly open to suggestions for changing it. Your arguments for who it'd be good to change it makes sense. I dont know which default is more valuable though, but I don't think it's very important.

Blaisorblade:
Merging then!

PR #557, opened 2016-11-16
Add Next and Done to keyboard; select all text on Focus in confirmation screen

NightMadness:
1. Clicking on textbox will now have a **working** next and done buttons.
2. All text in EditText will be selected on focus 
3. Renamed enableOrDisablePokeSpamBoxBasedOnSettings to showCandyTextBoxBasedOnSettings

screenshot_20161116-013555
screenshot_20161116-013844

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -39,6 +39,7 @@ import android.view.LayoutInflater;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.WindowManager;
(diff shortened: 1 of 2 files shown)


nahojjjen:
You accidentally added the line 
private void showCandyTextBoxBasedOnSettings() {
Twice, so the code does not compile.

Otherwise this feels like a good addition. A minor problem i noticed is that the CP textbox seems to have focus by default, which means that if you click it when the input window opens, it does not select all characters.

NightMadness:
Thanks, missed a spot.

Yeah android is weird with focus, ideally it should be out of focus on start.

PR #559, opened 2016-11-17
Adds credit activity accessible though settings

### LINKED ISSUE #328: Add a screen where we give credit to the third party tools and libraries the app uses
I believe some of the dependencies we have have licenses that require us to link back to their sources, something which we currently don't do. Even if the license does not require that we give credit, it would be a nice thing to do.

here's a list of the things I think we should mention (some might be missing)

Fabric

Tesseract

okhttp

butterknife

lombok

timber(?)

nahojjjen:
Several terms of open source projects require that you link back to the source somewhere in the app, and it's also the nice thing to do. So while this is not a feature or bugfix, it's still something that's required in the app.

edit: fixes #328

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,89 +1,90 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.kamron.pogoiv">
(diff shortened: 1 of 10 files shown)


NightMadness:
Tested. LGTM.  I prefer to hide with a new link at the bottom of GoIV Main Activity 😋, but either way will work.

PR #562, opened 2016-11-18
Adds custom separator clipboard token and different settings for single result

nahojjjen:
Allows the user to fill in their own separator string in the custom clipboard


NightMadness:
LGTM, tested working 👍

nahojjjen:
Ive started working with different settings for single/multi results on this same branch, because making a separate branch would result in conflicts.

I could have waited with this getting merged before starting on the new feature, but I don't know how long that will take, and I have free time to dev right now, and I don't know if i will later.

sarav:
I haven't read the code, but for single vs multiple IV, can you make it change the color of the circle? Black means single, white means multiple?

nahojjjen:
@sarav With this PR you can configure two completely separate settings for when theres a single or more result, so your scenario would certainly work.

Example settings:
Single result preference: IV% IVUNICODE - POKENAME5
Multi result preference: LowIV%-HighIV% - POKENAME5

PR #564, opened 2016-11-19
Sets auto update to false

nahojjjen:
So when the online version is published to google play, the user does
not get two different ways that update the app.


NightMadness:
Your right autoupdate from google play could be a problen

PR #566, opened 2016-11-20
Testing GoIV with Images Assets (Test PR)

NightMadness:
Hi Team,
This an initial test for unit testing with images.
This allows us to run images through android as instrument and see if the image is processing correctly.

its a good start, and one step close to an API, please let me know what you think.

tests


Blaisorblade:
I've given a *very* quick look, but I'm concerned by the amount of code you had to copy... It's good that at least you added comments there, but wouldn't it be better to separate that code so that it can be reused? Maybe that's the plan already? Maybe #499 would help?

Also, does this need a connected device?
While that makes sense for some sorts of tests (integration tests), ones that can be run in isolation are easier to run (e.g. we can run them easily on Travis).

Blaisorblade:
OTOH sorry I forgot to say: having such tests would be great, since GoIV keeps having regressions.

NightMadness:
I first tried a regular test, the problem is loading the SO library files for tess-two didnt work 

Regarding refactoring, thats the next step.

Blaisorblade:
> I first tried a regular test, the problem is loading the SO library files for tess-two didnt work

Uh, I wonder how pokescreenshot proceeds... But it seems it does the same, and uses mocking to test outside Android...

https://github.com/BraisGabin/pokescreenshot/blob/master/processing/src/test/java/com/braisgabin/pokescreenshot/processing/OcrUnitTest.java
https://github.com/BraisGabin/pokescreenshot/blob/master/processing/src/androidTest/java/com/braisgabin/pokescreenshot/processing/Tess2Test.java

Issue #567, opened 2016-11-20
Translation in calculation screen have to be 1 lined


chrisi51:
please change the german translation in the calculation screen ... i mentioned it in 3.2 and the problem still exist: http://imgur.com/a/EM79b
Please use "Verschiebe den roten Punkt mit dem Slider" instead. I guess, it should be understandable for germans.

NightMadness:
Created PR #569

chrisi51:
hopefully that works... can you make a release with that?

NightMadness:
Hi @chrisi51 can you confirm issues is fixed?

PR #569, opened 2016-11-20
Change slider prompt to be shorter and consume only 1 line.

NightMadness:
Change German translation based on user @chrisi51 at #567

I don't speak German so I used google translate to check.
before: `Use the slider to align the red point`
after:    `Move the red dot with the slider`


nahojjjen:
Looks harmless

PR #581, opened 2016-11-23
Fixes scanning 8 or 5 as S or B in hp

nahojjjen:
In some resolutions tesseract misstakes 8 and 5 as B or S, this fixes
that.


NightMadness:
Tested, working. 👍

Issue #580, opened 2016-11-23
Remove user corrections caching code?


sarav:
Creating this issue so that others can use it to discuss their thoughts on this.

I added the user corrections caching code. I'm wondering with the candy cost based detection, how often do we get the Pokemon line (set of evolutions) wrong? If we get the Pokemon line right all the time or very often, I don't think the user corrections caching is worth it for two reasons:

1. I've had cases where the Pokemon detection was wrong where for a specific Pokemon it got the detection wrong (completely unrelated Pokemon), but got the detection right for another Pokemon of the same species. So, I'm guessing it was due to user corrections caching, but I have no clue how it could have matched a completely unrelated pokemon -- because even with renames, I leave the first few characters of the Pokemon name.

2. It would just be unnecessary code that we could delete.

nahojjjen:
Just to mention the pr, #582 removes this code

NightMadness:
User correction are only usful for eevolutions.
Any similar need for future gen2 pokemon?
After we remove user corrections we can implement code to detect eevee type using candy or pokemon type.

nahojjjen:
It could be worth it to generalize the isNidoranFemale method to a method which runs on any pokemon flagged as "difficult"

NightMadness:
So it seems we need user corrections for eevee and future gen2 pokemon

NightMadness:
So you suggesting we do the candy checking for gen2 pokemon as well
The list seems small enough.

nahojjjen:
User corrections only help with eeveelutions if the user has already renamed them, scannned them and wants to scan them again. I don't consider it a likely scenario.

NightMadness:
Most of my eeveelutions are renamed to 98% water and 96% fire

sarav:
@nahojjjen -- I rescan my pokemon often if I wanted to check the max CP value or max HP value or power up costs to determine which of my different pokemon species I want to power up. So, rescanning after rename is not an uncommon use case.

nahojjjen:
I dont consider it a scenario which occurs often enough that I personally want to implement an image identification method for "troublesome" pokemon as long as it's  just eeveelutions. Nidoran was problematic because it identified the completely wrong candy family, which made it awkward to correct it, eevee is just a tap in the dropdown.

PR #582, opened 2016-11-23
Removes the userCorrection implementation

nahojjjen:
App no longer records manual user corrections for pokemon names

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
+++ b/app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
@@ -17,33 +17,18 @@ import lombok.AllArgsConstructor;
  */
 public class PokemonNameCorrector {
     private final PokeInfoCalculator pokeInfoCalculator;
-    private final HashMap<String, String> userCorrections;
 
     public PokemonNameCorrector(PokeInfoCalculator pokeInfoCalculator, Map<String, String> storedUserCorrections) {
         this.pokeInfoCalculator = pokeInfoCalculator;
-        userCorrections = new HashMap<>(pokeInfoCalculator.getPokedex().size());
-        userCorrections.putAll(storedUserCorrections);
-    }
-
-    /**
-     * Saves the pokemon nickname relation to picked pokemon.
-     *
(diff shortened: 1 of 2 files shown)


NightMadness:
User correction are only usful for eevees. Any simlar need for future gen2 pokemon?

nahojjjen:
Oddish - vileplume & bellossom

slowpoke - slowbro /king

Eevee - espeon / umbreon


(later generation? dont remember clearly)
tyrouge - hitmon lee/chan/top

Poliwag - Poliwrath-politoed

NightMadness:
Ah so maybe we need user corrections in the future but not right now

NightMadness:
Moving discussion to #580

app/src/main/java/com/kamron/pogoiv/logic/PokemonNameCorrector.java
```diff
@@ -17,33 +17,18 @@
  */
 public class PokemonNameCorrector {
     private final PokeInfoCalculator pokeInfoCalculator;
-    private final HashMap<String, String> userCorrections;
 
     public PokemonNameCorrector(PokeInfoCalculator pokeInfoCalculator, Map<String, String> storedUserCorrections) {
```
sarav: You should delete the last parameter. Also, you should be able to delete these lines from Pokefly onCreate()

         sharedPref = getSharedPreferences(PREF_USER_CORRECTIONS, Context.MODE_PRIVATE);
         corrector = initCorrectorFromPrefs(pokeInfoCalculator, sharedPref);

Github isn't allowing me to comment on those lines.

When deleting those line, I'm wondering if you should just force empty "PREF_USER_CORRECTIONS" and store it so that we wipe that data from "disk" for a few versions. Otherwise, that file is going to stay there till the user clears the app data or reinstalls it.

Also check for more code that can be deleted.

nahojjjen:
Ive tried to find more code related to the userCorrections, thanks for telling me to look deeper. :)

I agree that flushing the user settings for old usercorrections could be good, but I'm not going to attempt to do it myself.

nahojjjen:
Merging as I believe I addressed all comments.

PR #583, opened 2016-11-23
Fixes base stat errors that are found from last update

nahojjjen:
Turns out metapod had an error in it in the dataset i used to copy the stat information.

The dataset has been updated now, and Ive double checked with the formula used to generate the base stats from the game stats.

### DIFF
--- a/app/src/main/res/values/integers.xml
+++ b/app/src/main/res/values/integers.xml
@@ -166,7 +166,7 @@
         <item>155</item>
         <item>210</item>
         <item>62</item>
-        <item>64</item>
+        <item>94</item>
         <item>151</item>
         <item>55</item>
         <item>86</item>


Blaisorblade:
LGTM. But didn't people also report Raichu?

NightMadness:
Tested 👍, Metapod fixed.

nahojjjen:
Here's the document I created to double check pokemon stats, https://docs.google.com/spreadsheets/d/1w_SGmSYCKKUgSywMmckw-7MBfXknpG5fiKKvKHSr1XE/edit#gid=0

I think my document got the same stats for raichu as the info that's in our integers.xml?

Blaisorblade:
I just trusted the report on Reddit. That spreadsheet matches (updated) Gamepress there so it's good: https://pokemongo.gamepress.gg/pokemon/26

PR #585, opened 2016-11-23
Improves settings layout and aesthetics

nahojjjen:
Old setting screen was confusing, as new settings have been added they
have not been grouped logically.

This pr groups settings together using android PreferenceCategorys, and
refreshes some setting descriptions.


nahojjjen:
image

Image of new settings

nahojjjen:
I interpret your suggestion to be

-----------------------------

**Settings**
UI 
-    Enable ocr editing dialog
-    Show translated pokemon names

Screen capture 
-    Use screenshots to calculate iv
-    Delete screenshots

Advanced features
-    Copy to clipboard 
-    Edit Clipboard and 
-    Pokespam Calculation

---------------------------

I agree that "Input editing" is confusing and should be changed, but I do not think replacing it with "ocr" will make it more clear.

I dont know where the settings "auto update", "autolaunch pogo" & "credits" would fall into your suggestion. I'm guessing either another setting group or advanced features. If it's placed in advanced features, it would be too bloated and would serve to hide GoIVs more useful optional features.

As for the category name "screen capture", I avoided "screen capture" as it's closely related to a specific implementation, and I believe it can be difficult for a user to interpret. Therefor I went with "GoIV mode". As far as the user can tell, they're changing how they're interacting with the GoIV application, therefor they're changing the mode.

I'd also like to avoid the group name "advanced features", as a lot of users dont look at them because they believe them to be complicated or only useful for specific people. So I went with "optional". 

My reasoning for the category name "App behavior" was not only based on the current settings, but also some settings that will potentially be added soon: 
1.Remember appraisal expansion 
2.Auto select 80% appraisal

My intentions with the "app behavior" category is to separate settings which add new features to goiv, and ones that edit how the user interacts with the application.

>Technically everything is "App behavior"

Technically, all our settings change the UI, so they're UI settings.  :)
Also, all our settings are advanced features, so they also fall under "advanced features".

I'm open for suggestion on changing the "App behavior" name to something better, I agree that it's quite vague.

sarav:
Deleting screen shots and using screenshots is not UI behavior :)

As for "Advanced features" vs "Optional features" -- I thought Optional features is kinda confusing and vague. Any setting is an optional feature. How about Miscellaneous features? Just throwing out suggestions. Don't have a strong opinion.

Input vs OCR -- How about "Enable scan correction dialog"? That seems pretty descriptive and easy to understand. I think most people will understand what "scan" means in this context.

"Screen capture" -- what is implementation specific about it? All of GoIV is based around capturing the screen. These settings allow the user to decide how to capture screens.

nahojjjen:
>As for "Advanced features" vs "Optional features" -- I thought Optional features is kinda confusing and vague. Any setting is an optional feature. How about Miscellaneous features? Just throwing out suggestions. Don't have a strong opinion.¨

What's confusing and vague about it?  They're features you can enable or disable. They're optional.

> Input vs OCR -- How about "Enable scan correction dialog"? That seems pretty descriptive and easy to understand. I think most people will understand what "scan" means in this context.

I agree, it sounds better

> "Screen capture" -- what is implementation specific about it? 

"Screen capture" is not terminology used by common people, only people who deal with screen recording and implementation.

sarav:
I think Screen capture is clear enough. There are pretty common English words. If you aren't using "Screen Capture" what's the alternative? Putting it under optional features just makes that group bloated IMHO.

nahojjjen:
> I think Screen capture is clear enough. There are pretty common English words. If you aren't using "Screen Capture" what's the alternative? Putting it under optional features just makes that group bloated IMHO.

It's currently under the category "GoIV Mode".

PR #586, opened 2016-11-23
Changes the noti-icon depending play/pause

nahojjjen:
shows a filled icon if goiv is playing, and a hollow icon if its paused,
so user can see if goiv is running without opening the notification
tray.


NightMadness:
Although I cannot test in all DPI's, Tested, LGTM.

PR #587, opened 2016-11-23
Bump version for 3.3.1

Blaisorblade:
We should have done this earlier. Maybe we should update to the next version code directly.


NightMadness:
LGTM

nahojjjen:
yea, we should probably update this to version 3.4

PR #593, opened 2016-11-27
Allow refining Too Many possibilities (mon with 10CP/10HP) with overall appraisal phrase. 

### LINKED ISSUE #568: The ?%-?%-?% result does not factor refining the possibilities with overall appraisal phrase.
While using the appraisal refining feature, the minimum and maximum IV range associated with overall phrase should replace the ?%-?%-?% (min-avg-max) summary when there are "too many IV combinations".  Each team's overall appraisal phrase will break down each Pokemon's minimum to maximum IV range into four separate possibilities.  This should be reflected after using the appraisal refining feature even if the IV calculations were skipped in the getIVPossibilities routine.
.

NightMadness:
Fixes #568 by allowing of refining "Too Many possibilities" (mon with 10CP/10HP) with both appraisal phrases. 
-- Refactored **some** Appraisal calculation to AppraisalHelper
-- Tried to keep much of the original code undisturbed.
-- Added calculations for Appraisal texts
-- Added AppraisalHelper parameterized Tests
-- Added Appraisal checkbox calculations

**Known issue:** 
-- Clipboard module doesn't use new calculation, this can be fixed in adding Appraisal selected variable to _ivScanResult_ which I can work on adding at a later time.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -58,7 +58,7 @@ import android.widget.Toast;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.kamron.pogoiv.clipboard.ClipboardTokenHandler;
-import com.kamron.pogoiv.logic.AppraisalHelper;
+import com.kamron.pogoiv.logic.AppraisalHelper.AppraisalPercentPrefect;
 import com.kamron.pogoiv.logic.CPRange;
 import com.kamron.pogoiv.logic.Data;
 import com.kamron.pogoiv.logic.IVCombination;
@@ -77,6 +77,8 @@ import com.kamron.pogoiv.widgets.PokemonSpinnerAdapter;
 import java.io.File;
 import java.text.DecimalFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


nahojjjen:
The last time we discussed the issue with 10cp/10hp pokemon in discord we ended up saying that we can probably remove the limit which keeps the app from trying to calculate their ivs, since most phones can handle the worst case limit of 4096 maximum iv possibilities. Removing the limit would work with the existing "multiple result" logic and decrease our code complexity.

This PR is a little complicated, but I think it has the potential of using less cpu time than allowing all brute-force attempts for all cp/hp combinations.  Despite this, I would personally prefer to just remove the limit which causes "tooManyCombinations" . We could alternatively have the "tooMany" option as before, but allow the user to ignore it and force goiv to calculate it anyways in settings.


Please dont see this as critique to your code style or anything, I haven't thoroughly evaluated it, I'm just questioning if this method of  solving the problem is really the way to go.

NightMadness:
Hi @nahojjjen, thanks for responding I can take a look at that as well, 
I see what your saying, so if we have 10CP/10HP we simply fill the entire table with all 4096 combinations and let current code do its magic, that seems kinda lazy 😃 but efficient.

Blaisorblade:
> Despite this, I would personally prefer to just remove the limit which causes "tooManyCombinations" .

👍

nahojjjen:
Ive tried to do the "simpler" solution in pr #604

Blaisorblade:
BTW: I've just looked at the code briefly. #604 looks good at first sight, but I guess that adding these tests (and the needed refactorings) would be useful anyway, if you @NightMadness have time!

Blaisorblade:
IIUC this should be closed since #604 was merged, doing that, sorry if I should have checked more closely.

Issue #366, opened 2016-09-04
We waste storage with an update.apk and maybe more


Blaisorblade:
I and @sarav talked about it in https://github.com/farkam135/GoIV/issues/317#issuecomment-243728926.

This is for data in `SDCARD_PATH/Android/data/com.kamron.goiv/files/`. And for caches of unknown provenance (and stored not sure where).

Just remembered 20 MB should be for the OCR training data that we copy out of the archive at startup (src in `./app/src/main/assets/tessdata/eng.traineddata`). I assume that's necessary but haven't investigated.

Not sure why we use so much more than that though. Right now the relevant code (in Pokefly) is:

```
        if (!new File(extdir + "/tessdata/eng.traineddata").exists()) {
            CopyUtils.copyAssetFolder(getAssets(), "tessdata", extdir + "/tessdata");
        }
```

But that's fine.

However, at some point I found an `update.apk` of 28.9MB in there, probably a GoIV apk (from unknown sources). That doesn't seem saved by the downloader, maybe by the Android installer code?

DJCrapsody:
I can confirm the following:
I checked the data folder and only found tessdata but in the settings, it says that GoIV is holding onto ~79MB total data (~35MB app + ~44MB data) and ~16MB cache.
Wipe app data, run app, start GoIV.
Tessdata copied into GoIV folder. ~59MB total data (~35MB app and ~24MB data) and ~16MB cache. There's a phantom 20MB data drop.
This is based solely on using offlineDebug so no update or anything.

Blaisorblade:
> ~59MB total data (~35MB app and ~24MB data) and ~16MB cache. 

24MB is close to the size of tessdata, so let's say that's fine. But the rest is indeed weird. (And we see that even using releases).

DJCrapsody:
Okay the update apks will not affect the data storage. Somehow there's this phantom 20MB, and I'm unable to find it because I didn't root my phone. We only have this solution to work with: http://stackoverflow.com/questions/13006315/how-to-access-data-data-folder-in-android-device

oskomonzon:
Debug offline version is around 80MB on my device too, same as @DJCrapsody.
I'm rooted, so I'll see if I find something.

DJCrapsody:
Here's my research so far.
3.1.0 Online Release Build
LG V10 - Stock Android 6.0

http://imgur.com/a/brh3v

sarav:
[...]

PR #598, opened 2016-12-01
Delete previously downloaded apk file

rhari991:
Partially solves #366 
Checking during startup if an apk file exists, and deleting it.

### DIFF
--- a/app/src/online/java/com/kamron/pogoiv/updater/AppUpdateUtil.java
+++ b/app/src/online/java/com/kamron/pogoiv/updater/AppUpdateUtil.java
@@ -9,7 +9,6 @@ import android.os.Environment;
 import android.support.v4.content.ContextCompat;
 import android.support.v4.content.LocalBroadcastManager;
 import android.support.v7.app.AlertDialog;
-import android.util.Log;
 
 import com.kamron.pogoiv.BuildConfig;
 import com.kamron.pogoiv.MainActivity;
@@ -27,8 +26,6 @@ import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
 import timber.log.Timber;
-
-import static com.kamron.pogoiv.updater.DownloadUpdateService.FILE_NAME;
 
 
 public class AppUpdateUtil {
@@ -107,7 +104,7 @@ public class AppUpdateUtil {
     public static void deletePreviousApkFile(Context context) {
         if (ContextCompat.checkSelfPermission(context,
                 Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED){
-            String newApkFilePath = context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) + "/" + FILE_NAME;
+            String newApkFilePath = context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) + "/" + DownloadUpdateService.FILE_NAME;
             final File newApkFile = new File(newApkFilePath);
             if(newApkFile.exists() && !MainActivity.isGoIVBeingUpdated(context)) {
                 newApkFile.delete();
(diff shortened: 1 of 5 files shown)


nahojjjen:
Looks good to me, but I dont have the competence to do a good review concerning the auto download code. I'll merge if someone else also checks the code for errors.

nahojjjen:
@sarav or @Blaisorblade could you give an opinion or an okay so I can merge?

app/src/online/java/com/kamron/pogoiv/updater/AppUpdateUtil.java
```diff
@@ -94,4 +103,15 @@ public void onClick(DialogInterface dialogInterface, int i) {
         AlertDialog dialog = builder.create();
         return dialog;
     }
+
+    public static void deletePreviousApkFile(Context context) {
+        if (ContextCompat.checkSelfPermission(context,
+                Manifest.permission.WRITE_EXTERNAL_STORAGE) == PackageManager.PERMISSION_GRANTED){
+            String newApkFilePath = context.getExternalFilesDir(Environment.DIRECTORY_DOWNLOADS) + "/" + FILE_NAME;
```
sarav: Going by vague memory of discussion about the app's size bloat, I thought the downloaded apk was stored in GoIV's private data section.  Is it not? Shouldn't that be the one that's deleted? Also, now that the app is on the play store, do we even need this autoupdater?

sarav:
@nahojjjen -- if we merge this, do a squash merge and throw away the commit text about deleting unrelated files.

app/src/online/java/com/kamron/pogoiv/updater/AppUpdateUtil.java
sarav: Re-read some of the comments. At least @DJCrapsody was reporting that a copy of the APK was ending up in the Downloads folder. So, this seems okay. What about the update.apk that people were talking about? Is that why you call this a "partial" fix?

sarav: Actually, why isn't the delete in the Download service onStartCommand() not sufficient to delete this file? Isn't that service started all the time?

I'm not sure I understand the reason for splitting up the code into AppUpdateUtil and DownloadUpdateService, but is there any reason you consider that AppUpdateUtil is a better location for this code instead of DownloadUpdateService? Just a thought. I'm not sure what the right answer is to this part.

Blaisorblade:
@rhari991 Please also take a look at the Travis failures. I was confused by https://travis-ci.org/farkam135/GoIV/builds/180424954#L1008-L1018, but it turns out it fails compiling the offline version because `deletePreviousApkFile ` is missing there. I'd add an empty stub to the offline version.

nahojjjen:
Rhari asked me about the travis failures on discord, but I told him to ignore it because it seemed to only be caused by travis building the offline version.

As for whether or not the auto updated is still needed, Ive had a sorta- discussion about this with stopyourshenanigains, though that discussion was only related to the play-store version.

I think an auto updater is still good to have, for one- some people still download the github version, and they cant use the google play auto update feature. Another thing that scares me is the risk of getting kicked from play store again. If we do get kicked, having the auto update left in the code feels reassuring.

I did however toggle auto-update setting  to be defaulted to "false", as requested by stopyourshenanigains. This was to prevent users getting updates from both the play-store and the app itself.

Blaisorblade:
> Rhari asked me about the travis failures on discord, but I told him to ignore it because it seemed to only be caused by travis building the offline version.

We're still building the offline version, no at release time, so it better compile, no? Also, now all builds fail, and one can't (easily) tell whether it's an interesting failure or not: https://github.com/farkam135/GoIV/commits/master

If you don't want to build the offline version anymore, so you don't care about testing it, I recommend somebody remove it from the Travis test.

nahojjjen:
@Blaisorblade Yes, sorry I forgot to do that before merging the PR. I think it's good to keep the Travis build, so I'll push a method stub. PR here: #603

sarav:
I still compile and use the offline version.

PR #603, opened 2016-12-04
Fixes compile error for offline version

nahojjjen:
Adds a method stub for deletePreviousApkFile so offline build can
compile


Blaisorblade:
Compiles and LGTM. Thanks!

PR #604, opened 2016-12-04
Removes limit of not scanning 10cp hp lvl 1 pokemon

nahojjjen:


Blaisorblade:
At a first look, this seems pretty nice! Kudos on deleting the dead code! I can't look at it too closely but maybe @NightMadness can?

NightMadness:
Iv calc for 10cp/hp will not be accurate because pogo rounds up to 10.

app/src/main/java/com/kamron/pogoiv/logic/PokeInfoCalculator.java
```diff
@@ -190,32 +190,25 @@ public IVScanResult getIVPossibilities(Pokemon selectedPokemon, double estimated
...
-                            }
+        IVScanResult returner = ScanContainer.createIVScanResult(selectedPokemon, estimatedPokemonLevel, pokemonCP);
+        for (int staminaIV = 0; staminaIV < 16; staminaIV++) {
+            int hp = (int) Math.max(Math.floor((baseStamina + staminaIV) * lvlScalar), 10);
+            if (hp == pokemonHP) {
+                double lvlScalarStamina = Math.sqrt(baseStamina + staminaIV) * lvlScalarPow2;
+                for (int defenseIV = 0; defenseIV < 16; defenseIV++) {
+                    for (int attackIV = 0; attackIV < 16; attackIV++) {
+                        int cp = (int) Math.floor(
+                                (baseAttack + attackIV) * Math.sqrt(baseDefense + defenseIV) * lvlScalarStamina);
```
Blaisorblade: @NightMadness So if I understand you correctly, instead of `int cp = (int) Math.floor` we need to do like for `hp` above, right?
```java
int cp = (int) Math.max(10, Math.floor(...));
```

NightMadness: At one glace, Yes that should work. Some debugging is needed to make sure.

NightMadness:
Looks good to me

Blaisorblade:
Squash-merging on @NightMadness's review (the code also makes sense to me).

Issue #601, opened 2016-12-03
Screenshot method consistently fails to identify Pokemon and HP


Nincodedo:
Using screenshots, the Pokemon and their HP are consistently identified wrong. Using screen capture works perfect every time.

http://imgur.com/a/oMVvx

First two are screen capture. Last 3 are screenshot. I am using GoIV 3.3.1 and a Nexus 5X on Android 7.0.

nahojjjen:
should be fixed in #605

PR #605, opened 2016-12-06
Fixes #563 - screen shot scan error

### LINKED ISSUE #563: 3.3.0 stopped scanning correctly
3.2.0 correctly detected species, new version doesn't. I'm not sure how I can help debuggin (screen captureof my device?).

PoGo: 0.4.5
Nexus 4 running CyanogenMod 13, configured to 320 DPI (default).

UPDATE: this is when using manual screen capture, screen recording mode works better, but has number recognition problems mentioned in other issues.

UPDATE 2: level 30 pokemon are recognized as level 30.5 too.

Also, I love the perfect IV CP% feature, would it be possible (if it fits) to print the actual CP value or the difference? I.e. a 20% loss meaning 50 CP isn't the same as a 15% loss that means 75 CP loss.

EdwarDDay:
Screen shot images are scaled with the image ratio unchanged and additional
height cropped. This way the navigation bar is cropped and the OcrHelper
can find the pokémon information.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -18,6 +18,7 @@ import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Color;
+import android.graphics.Matrix;
 import android.graphics.PixelFormat;
 import android.graphics.Point;
 import android.graphics.drawable.Drawable;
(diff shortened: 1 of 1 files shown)


nahojjjen:
Ive heard that there's a problem with screenshot scanning, but havent spent time understanding it. Could you explain from the beginning what the problem is (was), and how you solved it? I unfortunately dont completely understand based on the description of your comment or by reading the implementation.

I'd like to understand the pr before I merge it.

EdwarDDay:
The problem on my Nexus 4 and, according to #601, on the Nexus 5X was, that the screen shot from the device captures the navigation bar at the bottom of the screen and than the images got scaled to the deviceMetrics values. But the deviceMetrics height matches only the screen height without the navigation bar.
So when the screen shot image is scaled, the complete ratio of the image is wrong. The image get compressed. So the OrcHelper uses wrong values for the image cropping.

So I first scaled the image ratio constant, to match the deviceMetrics width. This is done with the `android.graphics.Matrix` object. Than I cropped the additional height with the `android.graphics. Bitmap.createBitmap()` method.

nahojjjen:
Sounds good

PR #607, opened 2016-12-12
Add settings to optionally disable clipboard toast

MarcLandis:
The toast message gets annoying when checking IVs after a day of catching.

The default is still to show the toast like it used to be.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
+++ b/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
@@ -31,6 +31,7 @@ public class GoIVSettings {
     public static final String GOIV_CLIPBOARDSINGLESETTINGS = "GoIV_ClipboardSingleSettings";
     public static final String SHOW_TRANSLATED_POKEMON_NAME = "showTranslatedPokemonName";
     public static final String HAS_WARNED_USER_NO_SCREENREC = "GOIV_hasWarnedUserNoScreenRec";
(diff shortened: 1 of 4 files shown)


Blaisorblade:
Thanks for the PR! Not sure about the others, but this makes sense to me. I know we've been often for "not having too many settings" but this seems reasonable to me (plus, there's already a PR).

Regarding the actual code I'm not 100% sure (no testing done), but the code seems very consistent with what we have.

MarcLandis:
Thanks for the comment. After opening the PR I was later thinking if it would be a little bit cleaner to only get the GoIVSettings instance once and use it directly in the if statements.

Blaisorblade:
> After opening the PR I was later thinking if it would be a little bit cleaner to only get the GoIVSettings instance once and use it directly in the if statements.

The performance impact is minimal, so IMHO it doesn't make a big difference either way. But yeah, if you do that, it's more obvious there's no reason for concern for potential future readers.

nahojjjen:
I like the idea and the implementation of this, but I'd like to alter the string description for the setting.

I do not think a normal user understands what "show a (...) **toast**" means. "Toast" is just the name of the android implementation class. I think it'd be more clear if we described to the user something like "Show clipboard text popup" with the description "Show a small preview of what's placed in the clipboard with a popup"

MarcLandis:
I will change it as I have to fix the code anyway after #610.

PR #608, opened 2016-12-12
Adds possibility to evaluate gen 2 pokemon

nahojjjen:
Adds all pre-computed data about gen 2 pokemon except devolution
numbering - this is intentional, fixing so pokemon are identified
correctly will need to be done separately when more information is
available.

The patch currently has all pokemon names in english, even for other
languages. These need to be translated.


Blaisorblade:
Thanks for the quick reaction! No regressions should happen, so merging this quickly.

PR #611, opened 2016-12-15
Move hardcoded strings to resources to be able to translate them

MarcLandis:


nahojjjen:
One of the reasons I didnt make them resources is that I dont know if they're placeholders: the clipboard ui is "acceptable", but not very good, so I didn't know if I was going to change it... And if someone takes the effort to translate them, then changing it would feel more difficult.

MarcLandis:
No problem. So I will close this until the UI is "finalized".

PR #612, opened 2016-12-17
Fixes some faulty gen 2 stats

nahojjjen:
Previous stats were based on spreadsheet which had wrong info for some
pokemon, now updated based on bulbpedia info.


nahojjjen:
Heres a quick pdf of the calculations:
[GoIV Data - Stat sheet without mega.pdf](https://github.com/farkam135/GoIV/files/659071/GoIV.Data.-.Stat.sheet.without.mega.pdf)

You can ignore the second column, I just used it to find which pokemon had mega/alola evolutions - you can see some of them marked if you scroll into gen 3+, where i didnt bother to delete them. Notice that lapras is still using old stats, since niantic did not want to alter his stats while there was still an ongoing lapras event in japan.

PR #623, opened 2017-01-02
Auto appraisal scanning

MarcLandis:
- skip first two touches
- cache scan results
- remove not found phrases from cache


nahojjjen:
Have tested the pr and it's good, but waiting to merge this until appraisal dropdown to checkboxes #622 is merged, and so this can be amended to work with the checkboxes.

app/src/main/res/values/appraisals.xml
```diff
@@ -0,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="highest_stat_att">attack</string>
+    <string name="highest_stat_def">defence</string>
```
FoxFireX: In en-us, this will be "defense" instead.

nahojjjen: You're right, its "defense" in my pogo. Is it really "defence" in yours @MarcLandis ?

MarcLandis:
I guess that is just a typo. I fixed it now.

FoxFireX:
I was just assuming it was an en-uk translation; I know that word does differ in spelling between the dialects.  Which may mean that it will become an issue again regardless of which is in the default list.

MarcLandis:
I just set my phone to en-UK and it is"defense"

nahojjjen:
Okay, so reviewing the changes since the amend:

1. We need to fix the "todo" strings for  the other teams before this is merged, to make the feature work for other teams

2. I think the scan delay is too long - the text animation finished after 300 milliseconds on my phone at the longest text, and the current delay seems to be set to 800 ms?

MarcLandis:
I will go and add the  other phrases.

While the 300ms delay for one of my phones is ok, it isn't for one that is generally slower (CPU and GPU wise). I am not sure if the delay can be somehow calculated based on the phone information.

nahojjjen:
This seems to be working well, although I'd still like to change the scan delay, that can be fixed in a future patch

PR #625, opened 2017-01-04
Modified level detection algorithm to better handle L30+

FoxFireX:
Implemented algorithm starts at the center point of each candidate level, then searches for consecutive white pixels up, down, left, and right. By comparing how far we can travel in all directions with only white pixels, we can more accurately determine if the proposed center is the true center of the onscreen dot.  This helps significantly with level 30.0+ detection.


MarcLandis:
I did some tests with low and high level Pokemon and all were scanned correctly.

nahojjjen:
People seem to have tested it and says its working, and the code is clean.

PR #628, opened 2017-01-04
Avoid checking Nidoran gender twice

MarcLandis:
Using an Optional<boolean> to store the result of the first check in getPokemonNameFromImg
and reuse it in getCandyNameFromImg.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -30,6 +30,7 @@ public class OcrHelper {
     private final boolean candyWordFirst;
     private final String nidoFemale;
     private final String nidoMale;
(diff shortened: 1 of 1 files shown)


nahojjjen:
I dont know if this is a good idea - while its good to avoid repeated scannings too much, adding static variables such as isFemale to ocrHelper is kinda unintuitive - and if we change the app behavior in the future, it might be difficult to understand. I'd be more willing to merge this if isFemale was a heavy method, but test running it on my s7 says that it only takes around 6 milliseconds.

TripSixes:
How does the new gender icon affect this functionality?

nahojjjen:
@TripSixes good question, I could make a method which scans for ~~dicks~~ the gender symbol, and we could avoid the color scan method for nidoran. I know one guy asked for male/female as a token for the automatic naming, which would also need the feature.

PR #630, opened 2017-01-06
Refinement of the proposed appraisal checkbox dialog

TripSixes:


nahojjjen:
I havent tested it on my device yet, and I cant as not one of my 11 eggs ive hatched has resulted in a gen 2 baby pokemon, but the  changes look good. Could someone who actually has a gen 2 pokemon test & confirm before I merge?

Also, I didnt check and see if you overwrote the "fix some gen1 stats" commit by adding the poke names, I assume you didnt?

TripSixes:
I'm pretty sure that I did not lose any of those changes.  I was extra
careful to fetch and sync and merge from upstream before committing my
changes.

I'm pretty fluent in SVN version control. And trying to come up to speed
quickly on git.  I think this project is extremely interesting and am
excited to be able to contribute back.

I actually didn't try to submit the new evolution links in that commit.  It
was mainly focused on the appraisal input check boxes.  But wasn't sure how
to split it since I had already committed the evolution links to my own
master.

Andrew

nahojjjen:
So I've tested it on my device now, these are my comments:

We need to add evolution candy cost to integers.xml when we add the devolution numbers, so the PokemonNameCorrector can identify the pokemon. An example problem right now is caused by that pichu is listed as having "-1" upgrade cost right now. The object which identifies which pokemon is which, looks at the candy name, and the upgrade cost. For example, if the candy name is "pikachu", and the evolution cost is "-1", the identification module will find all evolutions in the "pikachu" evolution line, and check which has "-1" cost. This is currently BOTH raichu and pichu, which means we dont know what the method will return.

The second comment I have: 
The text wrapping for the checkboxes look nice although the added information does make it look a little more cluttered. I think this is an okay tradeoff though. One thing I'd like to undo in this commit is the un-alignment of the checkboxes. 




Images above show the difference - first image has the second row out of phase, while the second image has all checkboxes aligned.

Nincodedo:
Shouldn't the first and third row be radio buttons since you can only really have one of those at a time?

nahojjjen:
We discussed this in chat, but people didnt like the design of radio boxes.


But from a design pattern perspective, yes.

TripSixes:
Hmm, I thought that I did have the pichu evolution cost in there.. maybe I
changed it for testing a forgot to change it back.  Thanks for finding
that.

I was attempting to equally space the middle row... And I too thought that
it made it appear a bit "off".  I even had the hidden checkbox on that row
for a bit before switching back to this.  The proposed screenshot looks
great.

I too looked into radio buttons to prevent users from submitting more than
one.  Since I'm still coming up to speed on Java, I didn't feel confident
enough to attempt that change at this time.

One thing that I couldn't figure it was why the 1st and 3rd row check boxes
in the app are appearing in reverse order than the XML layout and studio
layout preview... Any thoughts?

nahojjjen:
>I was attempting to equally space the middle row... And I too thought that
it made it appear a bit "off".  I even had the hidden checkbox on that row
for a bit before switching back to this.  The proposed screenshot looks
great.

This is why I had 4 linear-layouts with vertical alignments next to each other instead of 3 horizontal layouts - that guarantees that they all line up.


>One thing that I couldn't figure it was why the 1st and 3rd row check boxes
in the app are appearing in reverse order than the XML layout and studio
layout preview... Any thoughts?

I wrote the hard-coded string previews in the wrong order probably - they actually read the strings from strings.xml to get the correct lines for the different teams.

TripSixes:
> We need to add evolution candy cost to integers.xml when we add the devolution numbers, so the PokemonNameCorrector can identify the pokemon. An example problem right now is caused by that pichu is listed as having "-1" upgrade cost right now. The object which identifies which pokemon is which, looks at the candy name, and the upgrade cost. For example, if the candy name is "pikachu", and the evolution cost is "-1", the identification module will find all evolutions in the "pikachu" evolution line, and check which has "-1" cost. This is currently BOTH raichu and pichu, which means we dont know what the method will return.

I've reviewed the integers.xml submitted as part of this PR and the Pichu->Pikachu->Raichu evolution line and their cost is already set to the expected "25".  Could you please check again?

I've also updated the alignment of row 2 and will attempt to update this PR with that change.

TripSixes:
All.  My apologies.  It would seem I'm accidentally pulling in all sorts of commits to this pull request.  Seems I need to be careful about what I merge into master if that is where I generate my pull request from.  

Please let me know if I need to try and clean this up.. either closing and recreating the pull request, Or by some other process?... 

Thanks in advance for your patience and guidance as I learn the ropes.

nahojjjen:
Sorry for the slow review, I  mentioned in our discord chat that I had test-week in university, so I've been quite busy...

PR #632, opened 2017-01-15
New MixedUnicodeToken for copy to clipboard

TripSixes:
The new MixedUnicodeToken is intended to be have exactly as the
original UnicodeToken, except that the filled vs. empty unicode
characters are now used to indicate exact or multiple values for each
 stat.

 For example, if CheckIV returns multiple results, this token will
 look at each stat individually to determine if each one only has 1
 possible value or multiple. (ie: Attack 11-12, Defense 9-11, and HP
 only has 13 for all combinations... this would result in Attack and
 Defense being one unicode character set and HP being the other.

 When modifying the clipboard settings, the trainer may select whether
 they wish for the "empty" to represent exact values or the "filled"
 to represent exact values.

 This provides a bit more clarity to the UIV value without adding any
  extra characters in the process.
(cherry picked from commit ecd2658)

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/clipboard/ClipboardTokenCollection.java
+++ b/app/src/main/java/com/kamron/pogoiv/clipboard/ClipboardTokenCollection.java
@@ -14,6 +14,7 @@ import com.kamron.pogoiv.clipboard.tokens.PokemonNameToken;
 import com.kamron.pogoiv.clipboard.tokens.PowerupsToMaxToken;
 import com.kamron.pogoiv.clipboard.tokens.SeparatorToken;
 import com.kamron.pogoiv.clipboard.tokens.UnicodeToken;
+import com.kamron.pogoiv.clipboard.tokens.MixedUnicodeToken;
 
 import java.util.ArrayList;
 
@@ -85,6 +86,8 @@ public class ClipboardTokenCollection {
         //Unicode iv representations
         tokens.add(new UnicodeToken(false)); //Unicode iv circled numbers not filled in ex ⑦⑦⑦
         tokens.add(new UnicodeToken(true));//Unicode iv circled numbers  filled in black ex ⓿⓿⓿
+        tokens.add(new MixedUnicodeToken(false)); //Mixed Unicode IV circled numbers, empty exact, filled multiple ex ⑦⓿⑦
+        tokens.add(new MixedUnicodeToken(true)); //Mixed Unicode IV circled numbers, filled exact, empty multiple ex ⓿⓿⑦
         tokens.add(new HexIVToken()); //hex representation of iv (ex A4B)
         /////////////////////////////////////////////////////////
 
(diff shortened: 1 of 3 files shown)


nahojjjen:
Hey, the checkstyle thing failed for this commit, because there's some small errors, like the line-length being 121 characters instead of the max 120, and that one of your curly brackets is on the wrong indentation level.

The full error log for the "errors" is here:
```
/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/ClipboardTokenCollection.java:89: Line is longer than 120 characters (found 121). [LineLength]
[ant:checkstyle] [ERROR] 

/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/MixedUnicodeToken.java:20: Line is longer than 120 characters (found 123). [LineLength]
[ant:checkstyle] [ERROR] 

/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/MixedUnicodeToken.java:58:9: '}' at column 9 should be on the same line as the next part of a multi-block statement (one that directly contains multiple blocks: if/else-if/else or try/catch/finally). [RightCurly]
[ant:checkstyle] [ERROR] 

/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/MixedUnicodeToken.java:73:50: '+' should be on a new line. [OperatorWrap]
[ant:checkstyle] [ERROR] 

/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/MixedUnicodeToken.java:74:51: '+' should be on a new line. [OperatorWrap]
```


Besides that, I think the idea for the token is pretty good, but there are things I'm concerned about.

I'm concerned about how the high and low (like "highestAttack") have been added to scanresults - I see that they're different from the existing "highAttack",. I'd prefer if you made the token actually loop through the ivCombinations in the ivScanResult, to prevent having these very similar instance variables. This would also ensure that the small performance cost is only added for those who actually use this token, and the code-complexity of IVScanresult stays low.

TripSixes:
Thanks for the detailed feedback.  I'll fix the check styles and look into moving the highest and lowest as you suggest.

TripSixes:
> I'm concerned about how the high and low (like "highestAttack") have been added to scanresults - I see that they're different from the existing "highAttack",. I'd prefer if you made the token actually loop through the ivCombinations in the ivScanResult, to prevent having these very similar instance variables. This would also ensure that the small performance cost is only added for those who actually use this token, and the code-complexity of IVScanresult stays low.

After review, I agree that pre-calculating all of the lowest and highest of each stat during "addIVCombination" is unnecessary.

I have refactored this to be contained within MixedUnicodeToken.java, however, code inspection is now stating that .getValue is too complex to analyze by data flow algorithm.

The middle ground here would be to move the lowest/highest calculations into a dedicated method within either IVCombination or IVScanResult.  This way it can be re-used (if ever needed), while also aiming to reduce complexity of getValue and only being used during token generation.

Thoughts?

nahojjjen:
> code inspection is now stating that .getValue is too complex to analyze by data flow algorithm.


That's not a problem, as long as we can just test and see that it works ourselves. :)


The changes look good 👍 

There's currently one line of code which checkstyle is complaining about:

/home/travis/build/farkam135/GoIV/app/src/main/java/com/kamron/pogoiv/clipboard/ClipboardTokenCollection.java:89: Line is longer than 120 characters (found 121). [LineLength]


After that, I think it's safe to merge.

Issue #635, opened 2017-01-20
IV button not appearing on Samsung S3 CyanogenMod13 (6.0.1)


adam-uk:
Hello, I use this great app on another device, but I'd really like to get it working on the S3 too.
Happy to provide logs etc if you tell me how to do it. The screenshot/share option works, so that's a start.

Samsung S3 GT-I9305 CyanogenMod13 (6.0.1) 
GoIV 3.3.2

Thanks.

TripSixes:
You might try taking a screenshot and uploading that here, unedited.   From what I understand, the app has many screen profiles that help it detect the expected pixels in the expected locations.. maybe it's just a bit off.

From a fresh install, do you get all the expected permission pop-ups?  Screen overlay, storage, draw over other apps.. 

And lastly, do you get the notification indicating that the service is started?.. even without button appearing?

adam-uk:
I just reinstalled to double check. Yes to 'Allow photos media etc' and allow
drawing over other apps. Press start - 'GoIV will start capturing everything that's displayed on your screen'. Press 'Start Now'...starting...running. The service does start, I can see it in the status bar and notifications.

If PoGo is already running, it turns the screen black like it does on my other device (known issue, but an indication of what's going on)

I've hopefully got the many display optimisations disabled.

![screenshot_20170122-213108](https://cloud.githubusercontent.com/assets/7096262/22186258/bbc52b4a-e0ea-11e6-86d3-4ad344402360.png)

TripSixes:
And just to be sure.. even if you're on the screen, then touch again (not on a button) .. GoIV still doesn't make the button appear?  It scans the screen once per second for 3 seconds to look for matching pixel colors. .. I guess that's assuming that the screen scan service is working as expected to begin with.

adam-uk:
Yep, unfortunately it still doesn't appear.

I have found an app that is able to do screen recording on the play store 'Game Screen Recorder', it shows the familiar message:

'Game Screen Recorder will start capturing everything that's displayed on your screen'. 

That at least suggests the 'display api' is sort of functional (I think).

TripSixes:
[...]

PR #640, opened 2017-01-24
Add alternate "white" for pixels[0] location

TripSixes:
A screenshot submitted in issue #635 seemed to indicate that some
phones might display the white area of the Pokemon screen in more than
one color of white.  Specifically, that screenshot had #fafafa and #f9f9f9.

Adding the alternative color observed, rgb(249,249,249).


nahojjjen:
I wonder why this phone does not display the right white - it does not seem to be running any screen tinting apps.

I'm gonna merge this fix as I dont think there's any noticeable performance impact, and the fix is quite straightforward.

PR #644, opened 2017-01-25
Refines logic for matching appraisal phrases and updates for Valor/Instinct

TripSixes:
Instinct appraisal phrases use the word "best" quite often.  Due to
this, addInfoFromAppraisalText is updated to only try to match
IVRange when appraisalRangeGroup is not already checked. Also change
"best" to "best of" in instinct_percentage1_phrase2.
-- Tested with Instinct account and all matches seem to be accurate now.

In addition, some match-words for Valor are subsets of other words
(strong/strongest, great/greatness).  This change adds a space after
strong and great so as to prevent matching strongest/greatness.
-- Has not actually been tested against a live Valor account yet.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/AutoAppraisal.java
+++ b/app/src/main/java/com/kamron/pogoiv/AutoAppraisal.java
@@ -148,10 +148,7 @@ public class AutoAppraisal {
      * @param hash the hash of the bitmap used by ocr
      */
     private void addInfoFromAppraiseText(String appraiseText, String hash) {
-        boolean match = false;
-        if (appraisalRangeGroup.getCheckedRadioButtonId() == -1) {
-            match = setIVRangeWith(appraiseText);
-        }
+        boolean match = setIVRangeWith(appraiseText);
         if (!match) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/kamron/pogoiv/AutoAppraisal.java
```diff
@@ -148,7 +148,10 @@ public void reset() {
      * @param hash the hash of the bitmap used by ocr
      */
     private void addInfoFromAppraiseText(String appraiseText, String hash) {
-        boolean match = setIVRangeWith(appraiseText);
```
nahojjjen: What does this change do?

TripSixes: I think something's missing.. need to review on computer

nahojjjen: Yea I think its best to either revert, or make clear comments so we know whats going on.

nahojjjen:
Is this ready for merge now?

TripSixes:
As far as I can tell..

PR #645, opened 2017-01-25
Adding "not" to "out of the norm" to match phrase from Blanche

TripSixes:
Re-adds the word "not" in the "not out of the norm" phrase.


nahojjjen:
I'm guessing this is inspired by the reddit comment

I dont feel like this is a needed change, but I'm not entirely against it - does the text fit though?

TripSixes:
Worked on my screens...

nahojjjen:
Why not

Issue #631, opened 2017-01-11
Should swiping GoIV from Recent Tasks stop the Pokefly service?


TripSixes:
As mentioned by @voltaek in issue #615 

> It seems if you swipe GoIV out of your running app tray, the notification doesn't go away and GoIV continues to run. I believe I'd read that swiping the GoIV app away in the tray was supposed to stop the service/notification and all related GoIV things, so this might be a bug we should look into, and might also be somehow related to this weirdness I'm seeing.

Opening up this issue for discussion.

Should the Pokefly service (and related notification) be stopped when the user swipes away the GoIV MainActivity from the Recent Tasks?

I dug around a little bit last night to familiarize myself with how it's working today.  Since we have different activities (MainActivity vs Pokefly), the answer to @voltaek's statement above is that the GoIV that shows up in the Recent Tasks is not directly related to the Pokefly service that is started.  Thus, when you swipe away the MainActivity, it is merely calling onDestroy for that activity and not for Pokefly.

Reading many posts and discussions about Android's implementation of the Recent Tasks, swiping things away, apps being Stopped vs apps being Destroyed, .... seems there's no real guarantee that your "onDestroy" will be called if a user swipes away an app from Recent Tasks.

With that in mind, I am testing the addition of some Pokefly service cleanup within the onDestroy of the MainActivity.  So far the change appears to be performing as expected.

    public void onDestroy() {
        LocalBroadcastManager.getInstance(this).unregisterReceiver(pokeflyStateChanged);
        LocalBroadcastManager.getInstance(this).unregisterReceiver(showUpdateDialog);
        LocalBroadcastManager.getInstance(this).unregisterReceiver(restartPokeFly);

        if (Pokefly.isRunning()) {
            stopService(new Intent(MainActivity.this, Pokefly.class));
            if (screen != null) {
                screen.exit();
            }
        }
        TODO: What if !Pokefly.isRunning()?
        super.onDestroy();
    }

If I swipe away GoIV in recent tasks while Pokefly is running, it seems to successfully stop the service (and casting), and clear the notification.  Subsequent launch of GoIV shows the correct state with "Start" being the text of the launch button.

As indicated by the TODO, I'll need to add an else for what to do if Pokefly is not running. ie: clear the Notification.

Comments? thoughts? areas of concern? better way of achieving this?

Thanks!

nahojjjen:
[...]

PR #646, opened 2017-02-17
Service lives

sarav:
I don't think the Pokefly service should be killed when the activity is killed. It doesn't work well for mid/low tier phones. It also goes against the general app behavior expectations.


TripSixes:
Just to link.. this is a revert of #631

PR #648, opened 2017-02-17
Link all remaining Gen1->Gen2 evolutions

TripSixes:
This update links all remaining Gen1->Gen2 evolutions that were previously unlinked due to unknown candy cost to evolve to their Gen2 forms.  Evolution links and candy costs are now updated.


TripSixes:
Gen2->Gen2 are already linked as far as their "devolutionNumber" is concerned.  But, I believe that GoIV also utilizes the evolutionCandyCost when determining the evolution chain.  ie: if the cost to evolve is -1, then GoIV sees it as a final evolution.  Once the Gen2 Candy Cost update is submitted, these should all start behaving as expected.

PR #653, opened 2017-02-19
Strings in spanish up to date

MARTINI1:
Updated strings of spanish in "values-es" and updating with new strings added of the original "strings.xml" file.


nahojjjen:
I dont speak spanish, but google translate tells me you haven't hidden any dick jokes in there, so I think it's alright.

PR #656, opened 2017-02-19
Adddefaultbigappraisal

nahojjjen:


nahojjjen:
This adds the option to default the appraisal box to be expanded when the user opens the goiv input screen. The strings are hardcoded in this commit, and needs to be amended. Merge is rushed to make it into next release.

PR #660, opened 2017-02-21
Update French translation

Remiscan:
I translated the new strings and the appraisals.xml file. I tested auto appraisal in French for team Instinct on multiple Pokémon, and it seems to be working. I couldn't test it for the other teams, but I checked that the checkboxes text was displayed ok.

All the token-related strings still appear in English in the app, so I guess they're still hardcoded, but I translated them anyway.

### DIFF
--- a/app/src/main/res/values-fr/appraisals.xml
+++ b/app/src/main/res/values-fr/appraisals.xml
@@ -1,0 +1,60 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="highest_stat_att">attaque</string>
+    <string name="highest_stat_def">fense</string><!-- "défense" isn't recognized -->
+    <string name="highest_stat_hp">pv</string>
+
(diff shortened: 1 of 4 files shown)


Remiscan:
A few words weren't detected by GoIV, mainly those with accents. With the new commit I just pushed, I couldn't find (yet) a Pokémon where GoIV was wrong when auto-detecting appraisal for team Instinct. It should probably be tested for the other two teams too.

nahojjjen:
So, does this make auto appraisal work in french?

Remiscan:
Yes it does, at least for team Instinct. There may be some adjustments to make for other teams, but I don't think I can test that myself.

nahojjjen:
Could you post a screenshot showing the appraisal/input screen, so I can see stuff align correctly?

Remiscan:
Sure, here are some screenshots: http://imgur.com/a/TwB7O

nahojjjen:
Looks good 👍

Issue #667, opened 2017-03-09
Incorrect detection of some "baby" evolution lines


TripSixes:
Pokemon within the following evolution lines are not being correctly identified when GoIV scans and attempts to figure out which Pokemon it is looking at.

Magby -> Magmar
Smoochum -> Jynx
Elekid -> Electabuzz

The other baby evolutions (Pichu, Cleffa, Igglybuff) do not seem to be affected.

The root cause seems to be that the "getBestGuessForEvolutionLine" method is using the list of "BasePokemons" when attempting to "getNicknameGuess".

    private ArrayList<Pokemon> getBestGuessForEvolutionLine(String input) {
        //candy name will only ever match the base evolution, so search in getBasePokemons().
        PokeDist bestMatch = getNicknameGuess(input, pokeInfoCalculator.getBasePokemons());
        return pokeInfoCalculator.getEvolutionLine(bestMatch.pokemon);
    }

The comment in the code predates Gen2, and thus at that time, it was safe to assume that the candy name would always be the base (or lowest) pokemon in a given evolution line.

With Gen2, Niantic has introduced babies and has not yet updated the candies associated with them (and may never update them).

This issue needs to fix the list of Pokemon that is sent to getNicknameGuess so that it includes all valid "candy" name Pokemons.

TripSixes:
Resolved by #668 

APK with these changes: https://goo.gl/6KsyCo

nahojjjen:
Since #668 was merged Ill mark this as fixed and close

PR #668, opened 2017-03-09
Fix getBestGuessForEvolutionLine to correctly identify using Candy Name

TripSixes:
Prior to this update, the getBestGuessForEvolutionLine method assumed
that only "basePokemon" are used for the Candy names within PoGo.
With the addition of Gen2 Pokemon, specifically the babies, Niantic
has so far decided to keep the Candy Names set to their Gen1
evolutions.  This means that we must now track exactly which Pokemon
name is used by Niantic for Candy naming.

A new integer-array has been added to track this relationship.
Similar to the devolutionNumber, this array maps which Pokemon name
is used for the Candy Name for that evolution line. For the vast
majority, they will point back to themselves (ie: Bulbasaur = 0).
However, for cases like the babies, they will point to the actual
Pokemon in their evolution line that the candy is named after. (ie:
Pichu = 24 (which is Pikachu's array index number)).


TripSixes:
APK with these changes: https://goo.gl/6KsyCo

TripSixes:
resolves #667

PR #672, opened 2017-03-30
Experimental double-letter CP tier

DanySK:
This pull request adds an experimental CP tier clipboard token, which produces an extremely fine grained evaluation for the Pokémon in the AA (worse) to ZZ (best) range, based on the best estimation among the possible IVs.

Evaluations are weighted automatically against the strongest Pokémon available at its best IV. Such upper bound is computed (once per application execution at most), so it should add no maintenance burden.

The system is meant for those who want to use the Pokémon name as a sorting mean to prioritize training the most promising Pokémons CP-wise. The (maybe counterintuitive) choice of assigning ZZ to the best pokémons is due to the IV measure being lexicographically sorting in the same way (with the notable exception of perfect IV and of single-digit IVs).

[Build Status](https://travis-ci.org/DanySK/GoIV)


nahojjjen:
Looks good, and your explanation of your reasoning is great.

DanySK:
Thanks.

PR #674, opened 2017-04-05
Add new IVSum Token

DanySK:
This very trivial token returns an IV evaluation in the most natural format: two digit number in 00-45. It just operates the sum of the single IV values. Solves the lexicographic sorting issues of the classic IV evaluation in 0-100 (single digit numbers and iv perfect get sorted badly).


nahojjjen:
Tested and seems to be working well, although I wonder why you choose to have it sum the highest possible IV combination instead of the average?

DanySK:
Thank you @nahojjjen.
It currently supports both highest and lowest - just load it with `false` in `ClipboardTokenCollection`. I personally favour the highest, since I want to spend the stardust into the most promising Pokémons - and sort accordingly. Of course also the average makes sense, I believe this token can be easily improved by using a `IVPercentageTokenMode` in place of the current `boolean`.

PR #682, opened 2017-05-04
Use the evolution with higher CP

DanySK:
This pull request changes the behavior of the ExtendedCpTierToken in case multiple evolution lines are available for a selected Pokémon. In accordance with the spirit of the initial proposal of ordering Pokémon by how promising in terms of CP they are, the system now computes the best possible outcome out of all the possible evolution branches.
For instance, with this change, Eevee's projected CP at maximum evolution will use Vaporeon's stats (rather than Umbreon's), Poliwag will match with Poliwrath (not Politoed), Oddish with Vileplume, and so on.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/ExtendedCpTierToken.java
+++ b/app/src/main/java/com/kamron/pogoiv/clipboard/tokens/ExtendedCpTierToken.java
@@ -8,6 +8,10 @@ import com.kamron.pogoiv.logic.IVScanResult;
 import com.kamron.pogoiv.logic.PokeInfoCalculator;
 import com.kamron.pogoiv.logic.Pokemon;
 
+import java.util.Deque;
+import java.util.LinkedList;
+import java.util.Objects;
+
 /**
  * Created by Danilo Pianini.
  * A token which returns a "tier" based on the pokemon max cp, in the AA-ZZ range.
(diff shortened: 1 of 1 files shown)


TripSixes:
Sounds a great update.  Will have to review and test it later.

nahojjjen:
Sounds like a good update. I noticed you code very differently from me, whats your code background (favorite language)? (pure curiosity, the code is understandable, just not used to it.)

DanySK:
Thanks for the appreciation :)

@nahojjjen - my coding abilities were forged on Java. I am quite a maniac of code quality (as you might see looking at the build configuration for PMD / FindBugs / Checkstyle of some of my projects). However, I find Java very verbose, so I decided to use Scala wherever possible from now on, and Xtend whenever I need pure Java code to be generated.
I like designing the system architecture thinking in OOP (in the large), but whenever it's easy to read and compact I favor functional programming (mostly in the small). For some specific uses, I also like logic programming (Prolog). I teach OOP (also with some focus on good engineering practices) at my university.
Other languages I use are C (rarely, low level stuff), Python (mostly as a Matlab replacement), C# (rare, mostly because of Unity), Groovy (mostly to write Gradle plugins), bash (well, a daily tool)... and probably something else that does not come to my mind right now :)

A question: am I allowed to use Java 8 novelties in GoIV, like lambdas and streams? I am writing pure Java 7 right now because that looked like how most of the stuff there is written.

nahojjjen:
I dont think we've discussed Java 8+ code, but I think most people dont write it in their commits because Java ~7 what most people are used to, and so it's easier to understand.

DanySK:
Thanks

PR #687, opened 2017-07-01
Fix Auto-Update for SDK Version 24

TripSixes:
[History, Explanation and Solution](https://inthecheesefactory.com/blog/how-to-share-access-to-file-with-fileprovider-on-android-nougat/en)

"Summarily, file:// is not allowed to attach with Intent anymore or it will throw FileUriExposedException which may cause your app crash immediately called."

GoIV's Auto Updater code has been updated as shown from this posting, 
implementing FileProvider to work for version 24 and above (and still backwards
compatible with prior versions).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -62,7 +62,7 @@ android {
      * Update (both Gradle and local SDK) whenever possible
      */
     compileSdkVersion 24
-    buildToolsVersion "24.0.2"
+    buildToolsVersion '25.0.0'
 
     /**
      * Build configurations for APK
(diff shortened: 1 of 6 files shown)


TripSixes:
My tests show that the crash is no longer occurring.  Unfortunately, I can't fully test the auto-update since I don't have the same signing key as the .apk posted on github as latest release.  My tests end up with a "APK is corrupt" message.

TripSixes:
Sorry for the mess.. going to see about cleaning up this PR and re-submitting tomorrow.

PR #697, opened 2017-07-11
A lot of stuff

nahojjjen:
This branch bot waaay bloated, and I'm sorry.

This refactors some stuff from pokefly, changes the look dynamically on the iv preview button, fixes some crashbugs, tweaks the pokemon corrector, and adds a clipboard token.


TripSixes:
I've been testing/using this branch since yesterday and here are my observations/suggestions:

I still like this clean look of just expanding the circle to an oval for the IV %.  One of the other suggestions might work too, but this is really good and I like the border coloring.

The IV Button and quickIV preview button still seem a bit large on my screen.

I think we should either remove the button or change it back to "IV" immediately when the user touches the screen again.  When swiping between pokemon, some folks might not realize that GoIV is still scanning/OCR'ing and that the numbers are from the previous scan.  Mine can take a few seconds for the results to be diplayed on the button.

I don't think there's any need for a Toast of the quickIV results if the button is working as expected.

Also, can we have GoIV "not" rescan again when clicking the IV to refine via appraisals or corrections?  in other words.. if the quickIV scan was successful (not failed), then just use that information to populate the GoIV screens if the user clicks on the button.

And lastly, I'm wondering if there is a way to prevent the quickIV scan/results if the user is simply closing the GoIV edit/input/details dialog.  .. the main issue I see here is that refining the results and then closing the GoIV window can result in quickIV rescan and update button to original range.  (Yes, the copy-to-clipboard is still accurate, but users may not realize that).
 - maybe the logic could be:
  - if user closes GoIV dialog, don't scan.
  - if user touches the pokemon screen but the Pokemon's uniqueID is the same, don't scan.
 - only rescan if a screen touch is detect AND the uniqueID is different.

All-in-all.. this update is a huge leap forward!  If user's were impressed by the quickIV Toast, then this will blow them away.

nahojjjen:
>I think we should either remove the button or change it back to "IV" immediately when the user touches the screen again. When swiping between pokemon, some folks might not realize that GoIV is still scanning/OCR'ing and that the numbers are from the previous scan. Mine can take a few seconds for the results to be diplayed on the button.

I agree, I also think there should be an indicator for which pokemon the quickpreview is currently showing a result for.

> I don't think there's any need for a Toast of the quickIV results if the button is working as expected

Currently the IV preview button will show the preview, and if you have the setting active for showing your custom clipboard, that will Toast.


>And lastly, I'm wondering if there is a way to prevent the quickIV scan/results if the user is simply closing the GoIV edit/input/details dialog. .. the main issue I see here is that refining the results and then closing the GoIV window can result in quickIV rescan and update button to original range. (Yes, the copy-to-clipboard is still accurate, but users may not realize that).
> 
> maybe the logic could be:
> if user closes GoIV dialog, don't scan.
> if user touches the pokemon screen but the Pokemon's uniqueID is the same, don't scan.
> only rescan if a screen touch is detect AND the uniqueID is different.


The preview button doesnt update if the user just closes the GoIV dialog, since the screenwatcher object checks (event.getActionMasked() == MotionEvent.ACTION_OUTSIDE), and doesnt perform the screentouched if it's false. 

Logic to prevent duplicate scans of the same pokemon would be nice, however during normal usage it doesnt occur that often in my experience, and we'd need to be careful to not cause unexpected scanning behaviour or increase code complexity too much if we try and fix it. If we attempt to use the pokemon unique ID to determine if we want to do a full rescan, we'd need to split up the scan into two broadcast intents instead of one, and even that would not prevent duplicate scanning as the scan would still occur if the user scrolls down slightly, since that changes the unique id.

PR #700, opened 2017-07-19
Refactor pokefly attempt - Andrew's contributions

TripSixes:
Good morning @nahojjjen,

This PR is intended to be a merge into your refactorPokeflyAttempt branch.  It is based off of that branch and should merge cleanly.

Major updates here are:
 - Fix the IV Button "big" issue.  Rewrote the iv_button.xml file to utilize rectangle shapes rather than ovals.  I took a real hard look at the exact pixel/dp alignment of the Pokemon Go buttons so that I could recreate the button accurately utilizing "dp" sizing.  I had to remove the width/height from many of the layers so that it would size correctly.  The width/height is now only defined on the "shadow" and then all other layers  resize accordingly.
 - Created a new "iv_button_background.xml" to be used for the quick preview functionality (ie: empty button with no "IV" characters/shapes).
 - dpToPix modified slightly... During my testing I found that the screen density formula in GoIV was not accurately calculating the actual density of my screen.  I think this was one of the factors leading to the incorrect size of the IV button and it being in the wrong place.
 - I tweaked the text coloring to be a scale of "red-orange-yellow-green", to hopefully make it easier to tell at a glance.  Also created 2 sets of colors, a lighter set for text and a darker set for the left and right shaded regions.

Feature still in progress:
 - I'm trying to get an additional coloring to only show up in the far left and far right of the IV quick preview button.  See screenshots below.   The only issue is that I'm having a hard time getting the code to work to be able to replace the Gradient startColor and endColor dynamically during run-time.

At this time, the below screenshots were created by hard-coding the color values in there just for the purpose of taking the screenshot.

screenshot_20170719-015849
screenshot_20170719-020422
screenshot_20170719-020945

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPopupButton.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPopupButton.java
@@ -3,7 +3,6 @@ package com.kamron.pogoiv.pokeflycomponents;
 import android.content.Context;
 import android.graphics.PixelFormat;
 import android.support.v4.content.res.ResourcesCompat;
-import android.util.DisplayMetrics;
 import android.view.Gravity;
 import android.view.MotionEvent;
 import android.view.View;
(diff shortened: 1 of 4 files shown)


TripSixes:
Closing this PR.

PR #709, opened 2017-07-30
Implement a "Worth Training" token

DanySK:
This PR implements a new token whose goal is to evaluate how worth is spending stardust on a Pokémon, considering its base stats and its individual values.

Motivation: with the recent updates, CP is much less important than it was before, as gyms are no longer based on it. At the same time, the removal of the 5000 stardust/day bonus for gym holders made stardust an extremely valuable resource, that should be invested very carefully.

The idea is that an average Pokémon with very high IV is a better long-term stardust investment than a strong Pokémon with very low IV. This token weighs the base stats of the Pokémon based on IV, then normalizes them considering the best possible Pokemon (lazely computed at runtime once per execution, no maintenance required in case of new Pokémons or changes in stats), computes the product of such stats (arguably a much better metric than the sum), uses a non linearity (cube root) to prevent the values from flattening, and finally normalizes in the 00-99 range.

A few examples:
* a 96% Lugia scores 87
* a 100% Dragonite scores 82
* a 100% Gyarados scores 80
* an 87% Lugia scores 79
* a 100% Meganium scores 68
* a 51% Houndoom scors 23


nahojjjen:
Looks like a good token to have

DanySK:
Thanks

PR #710, opened 2017-07-30
Fix recognition of portuguese candy names

onilton:
Portuguese has "doces" (candy) coming first in candy name.


nahojjjen:
Just to prove this change is correct, could you show / link a screenshot where the candy comes first? It looks like a good PR, I just want to make sure. :)

onilton:
Screenshot

onilton:
I think in the future we will have to split portuguese BR and portuguese (Portugal)

onilton:
Since portugal portuguese doesn't seem to have translation but portuguese BR does.

Issue #708, opened 2017-07-28
Failed to read CP value: the higher digit gets trimmed more often than not


thearaks:
I am using a Xiaomi Mi 6 (1080x1920 resolution) and the game is set to italian language.
So CP is translated as PL.

`OcrHelper.getPokemonCPFromImg` method fails to recognize the CP value correctly.
Since it always trims the first two characters of the OCR'd string, it erroneously removes the higher digit of the CP value if the CP acronym is detected as a single character.
As a consequence, it's often impossible to have a working quick scan, while the normal scans require a manual correction to add the missing digit.
This might have been introduced when the `tess-two` library was updated to v7.0.0.

Here are some examples:

![image](https://user-images.githubusercontent.com/5188791/28739230-90df0b4c-73f9-11e7-97de-9629bba6d69d.png)
Gets OCR'd as **H2976**

![image](https://user-images.githubusercontent.com/5188791/28739134-9df3d318-73f8-11e7-9283-e7a71d8b6eea.png)
Gets OCR'd as **H1193**

![image](https://user-images.githubusercontent.com/5188791/28739218-603a9f42-73f9-11e7-9637-8cca7841df07.png)
Gets OCR'd as **H654**

Sometimes, instead for the character "H", the acronym PL gets mistaken for an "m"!

We need to find some smarter way to clear out the CP acronym from the OCR output since we can't rely on the assumption that tesseract always recognizes the correct amount of characters anymore.

TripSixes:
I wonder if we can just do a regex removal of any non-numerical 0-9 characters.. ?

String value = string.replaceAll("[^0-9]","");

thearaks:
In that method, there's this comment:
```
* Always remove the two first characters instead of non-numbers: the "CP" text is 
* sometimes OCR'ed to something containing numbers (e.g. cp, cP, Cp, c3, s3, 73, 53 etc),
* depending on backgrounds/screen sizes, but it's always OCRed as two characters.
```

I don't know how much of that is still true, since I never had the chance to see the CP letters erroneously read as numbers, but for sure is not true the last sentence: in my case PL is OCR'd as a single character, not 2.

nahojjjen:
I havent checked what things OCR as right now, but I remember that we used to have a method that removed numbers, and it consistently produced errorous result on my S7 as the CP was OCR'd as C7.

Could we maybe make an algorithm that checks the "height" of the characters?
[...]

PR #712, opened 2017-08-03
Improve the CP value detection

thearaks:
Sometimes the CP acronym get scanned as one character instead of two.
This can can lead to the trimming of the first digit of the 'mon CP
value since the code always strips the first two starting characters
from the OCR'd text in order to remove the CP acronym.

This commit introduces the ability to detect the container rectangle of
each character inside the CP cropped image. After that, it ignores all
the characters whose height is lower than the average.
This method effectively detects the CP acronym, regardless of the user
locale and it being suffix or postfix to the value, since it's always
written in a smaller font size.

Resolves: #708

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -1,5 +1,6 @@
 package com.kamron.pogoiv;
 
+import android.content.res.Resources;
 import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.graphics.Rect;
@@ -551,8 +552,10 @@ public class OcrHelper {
         // Every chunk will contain a character
         ArrayList<Rect> chunks = new ArrayList<>(6);
         Rect currentChunk = null;
-        for (int x = 0; x < width; x++) {
-            for (int y = 0; y < height; y++) {
+        // On devices denser than XHDPI (2x) we can skip a pixel every two to increase performances
+        int increment = Resources.getSystem().getDisplayMetrics().density > 2 ? 2 : 1;
+        for (int x = 0; x < width; x += increment) {
+            for (int y = 0; y < height; y += increment) {
                 final int pxColor = cp.getPixel(x, y);
 
                 if (currentChunk == null) {
--- a/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/OcrHelper.java
@@ -2,6 +2,7 @@ package com.kamron.pogoiv;
 
 import android.graphics.Bitmap;
 import android.graphics.Color;
+import android.graphics.Rect;
 import android.support.annotation.NonNull;
 import android.util.LruCache;
 
@@ -10,8 +11,10 @@ import com.googlecode.tesseract.android.TessBaseAPI;
 import com.kamron.pogoiv.logic.Data;
 import com.kamron.pogoiv.logic.ScanResult;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


nahojjjen:
I dont have access to a computer at the moment, I read through the code with my phone, the idea seems nice,

I've got a few questions though.

Im not familiar with the chunks implementation, and I dont understand how you're detecting where each chunk should be - could you give an overview of how it works?

Does creating the chunks / looking at the image for where the chunks should go take a lot of proccessing power? Have you checked the difference in speed with this version compared to the previous implementation? It's fine if the method takes a bit longer than the previous if it's more accurate, but it'd be good to know a ballpark for the performance loss.

thearaks:
First of all, let's clarify what I my by chunk:
A "chunk" of the cropped CP image is a bounding box that identifies a single character.
detect


The algorithm to detect them is based on two simple assumptions:

1. The CP acronym always use a smaller font size
2. Each character is separated from the other with (at least) 1 full column of black pixels

It start from the top left pixel (0;0) of the cropped image and scans the first column from the top to the bottom.
If it reach the bottom of the column without finding any non-black pixel it goes 1px to the right and scans the next column.

As soon as it finds the first non-black pixel it infers that a **chunk** must **start** at the current _X_ coordinate, and is at least as high as the current _Y_ coordinate.
When a non-black pixel is found, the algorithm always skip immediately to the next column without reaching the bottom of the current column.

The algorithm keeps scanning, proceeding to the right, and each time it encounters a non-black pixel it updates the height of the current chunk if it is lower than the _Y_ coordinate of this non-black pixel. When a non-black pixel is found it skips to the next column.

As soon as a column full of black pixel is found, it infers that the current **chunk** is **ended**.

Then, the algorithm keeps going to the right until it finds another chunk or reaches the CP image end.

Once all the chunks are detected (we've reached the CP image end), it computes the average height of the chunks and discards every chunk that is below this value.
average

Then all the chunks are merged in a single bounding box that is passed to tesseract so it scans only the desired area instead of the entire CP image.
merge


Regarding the performances, I'll post a benchmark later on!
Anyway I can say that on a Xiaomi Mi6 it took **50ms** to execute this algorithm!
I'm also studying a possible optimization, I'll implement it on my spare time...

nahojjjen:
That looks like a very clever implementation, the only thing I could figure out would break it would be background noise, but that'd cause the ocr to fail anyways.

Do you want me to merge this now, or wait until you've checked if your "possible optimization" will be added?

thearaks:
Please wait a little more, I'm benchmarking 2 optimizations.
And for the background noise we could ignore chunks whose width is too small (less than a couple of pixels).
Writing code in this very moment!
I'll post an update later this evening!

thearaks:
**Performance analisys**
Durations are computed obtaining timestamp at first line of `getPokemonCPFromImg` method and before its return statement.
All tests executed on a Nexus 5, a fairly old device.

Four scenarios:

1. Only tesseract on entire CP image
2. New algorithm
3. New algorithm with _skip-a-pixel_ optimization
4. New algorithm with _skip-a-pixel_ and _skip-a-column_ optimization

schermata 2017-08-07 alle 22 40 00

schermata 2017-08-07 alle 22 39 48

schermata 2017-08-07 alle 22 39 37

thearaks:
In the last commit I also implemented a strategy to avoid the chunks that might be generated by background noise. I simply check their width and see if they're too small!

Now the feature set is complete, this PR can be reviewed.
Don't be too harsh on me! ;)

udnp:
Hi @thearaks,

I seems that current `OcrHelper#get**S8**PokemonCPFromImg()` also needs this changes.

https://github.com/thearaks/GoIV/blob/857358bcd5e6518704bf8d67efa0951fd479fc17/app/src/main/java/com/kamron/pogoiv/OcrHelper.java#L911-L926

Thanks.

thearaks:
Thank you @udnp for your observation.
Yes, the code should be replicated also for S8.
But, since the S8 support is made with a dirty patch, I prefer to generalize how the 'mon screen scales instead of copy-pasting more code.
I think I can handle the issue in the next days, but not earlier than next week!

udnp:
@thearaks, ok,  no problem.
I'm sorry I dont know this reason.

thearaks:
Don't worry, no problem for me!

PR #716, opened 2017-08-05
Fix hard coded Nidoran name for OCR multilingualization

udnp:
Hi, I try and test to use GoIV for Japanese Pokémon GO game app.
https://github.com/udnp/GoIV_JP

This PR is one of some tips I noticed for OCR multilingualization,
used Nidoran name from resource string instead of hard coded.

Could you check this? 
Thanks.

PR #721, opened 2017-08-09
Fix hard coded Nidoran name for OCR multilingualization (updated)

udnp:
Hi, GoIV project team.
I try and test to use GoIV for Japanese Pokémon GO game app.
https://github.com/udnp/GoIV_JP

This PR is one of some tips I noticed for OCR multilingualization,
used Nidoran name from resource string instead of hard coded.
(see also https://github.com/farkam135/GoIV/pull/716)

If judging Nidoran name with hard coded value "nidora",
it is failed for other language Nidoran name , like Japanese name "ニドラン".  
So this PR changed to use Nidoran name from resource string, instead of hard coded value.

Could you check this?
Thanks.


thearaks:
Looks good to me.
Built and ran on my Nexus 5 with PoGO localized in Italian an worked flawlessly on a Nidoran M!
Also the other 'mons are detected correctly.

udnp:
@thearaks, thanks for your check.

And I have finished to check that this PR causes no problem to the GoIV setting "Show translated monster name".

PR #725, opened 2017-08-17
Improve clipboard modifier activity

thearaks:
This PR introduce various improvement to the UI and UX of ClipboardModifierActivity.
Since it is a huge change, I'm open to opinions and suggestions from everyone.
I'd also need some testing on KitKat devices!

A quick glance to the UI:

image

Pressing the "star" floating action button let the user switch between standard and max evolution variants:

image

Selecting a token highlights it and shows the description in a wider TextView, then it can be added using the "plus" floating action button:

image

Enabling "copy different on single result" in settings adds a tabbed navigation:

image

Adding a custom separator is now easier:

image

image

image


thearaks:
Update 18/8:

- Add drag'n'drop to preview list so the user can rearrange its tokens and add swipe up or down gesture to delete a token.
- Increase tokens size in showcase grid to simplify selection.
- The user is again able to delete all tokens.
- Add a Dialog to tell the user that has unsaved changes on exit.

Updated screenshot:
image

PR #727, opened 2017-08-25
Fix hard coded pokemon type names for eeveelution correction

udnp:
Hi GoIV team.

This PR fixes hard coded pokemon type names for eeveelution correction.
So eeveelution correction is able to work on non English language also.
This is tested on English and Italiano (and my local Japanese fork build).

But now, to work this fix on other languages(eg de, fr,...),
you have to add localized pokemon type names to your locale `pokemons.xml`,
[like Italiano(it) in this patch](https://github.com/farkam135/GoIV/compare/master...udnp:pr/m17n-eeveelution-correction?expand=1#diff-5c3da1c320986525d8e5971ffde60920).

It seems that eeveelution correction sometimes fails in my test emvironment.
But it is because pokemon type OCR result is wrong(eg `PS**Y**CHIC` is `PS**V**CHIC`).
So I think this problem is caused by OCR error, not by my this PR.

Could you check this?
Thanks.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -454,7 +454,8 @@ public class Pokefly extends Service {
                 getResources().getIntArray(R.array.stamina),
                 getResources().getIntArray(R.array.devolutionNumber),
                 getResources().getIntArray(R.array.evolutionCandyCost),
-                getResources().getIntArray(R.array.candyNames));
+                getResources().getIntArray(R.array.candyNames),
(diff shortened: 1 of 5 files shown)


nahojjjen:
I tested it on my samsung s7, and all the current eeveelutions still work for english on mine. Thanks for the PR :)

udnp:
@nahojjjen, Thanks for your test and merge. 😃

thearaks:
@udnp Thank you for including the Italian strings, they are all correct!

udnp:
@thearaks Thanks for your checks.
Italiano choiced as official locale other than English in my this tests.😃

PR #728, opened 2017-08-25
Alters input screen and adds raidEgg switch

nahojjjen:


nahojjjen:
Changes the alignment of the Species text, so there isnt a lot of whitespace.

Makes the normal input area become invisible when the user enters the additional refinement section

Adds a switch for egg/raid catches, that removes combinations where an iv value is below 10.


(The second screenshot has a misspelling that Ive fixed after taking the screenshot, and I also made the switch text the same color as the checkboxes)

nahojjjen:
And I also added a text-indicator next to the level input that shows a numeric value of the current choice.

PR #733, opened 2017-08-31
Update gradle to last version in Android Studio

MARTINI1:
when using android studio with this project all the time I get the notice to update the version of gradle and I have decided to modify it so that it is updated


thearaks:
Ok for me, this solves the outdated Gradle plugin version in Android Studio 2.3.
This will fix the annoyng warning until Android Studio 3.0 stable will be released!

PR #743, opened 2017-09-08
More hardcode text to strings.xml (now with many more)

MARTINI1:
extracted a very beast amount of hardcoded text and tested to see if everything works fine

if you want to test I share the apk with the changes made in this PR:
https://drive.google.com/file/d/0B8YGb9eCs7lQdEoxMVFQSU1qWGs/view?usp=sharing


thearaks:
Even if I'd correct some minor things here and there, I'd merge this since it's a huge improvement from the current situation of recalibration and clipboard functionalities localization!

PR #746, opened 2017-09-12
Corrections and other tweaks in tokens and strings

MARTINI1:
- Revert change in tools:text L43 from activity_ocr_calibration_result.xml
- Extract names tokens to strings.xml
- Fix english tips


thearaks:
Very nice work.
+1 for me!

PR #749, opened 2017-09-17
Improved the OcrCalibrationResult screen and others fixes

MARTINI1:
now it is easier to understand if the calibration has gone well or there is something wrong
also have made some arrangements like improve strings, improve the status bar, add a button back to the game when the calibration goes wrong and have been fixed 2 warnings

# **v2 of OcrCalibrationResult**


nahojjjen:
Looks great

PR #787, opened 2017-10-11
GoIV tutorial

MARTINI1:


nahojjjen:
This needs more work before its merged into master, but I'm merging into the addTutorial branch, so more people can work with it.

PR #788, opened 2017-10-12
optimized GoIV with some fixes and improvements

MARTINI1:
apart from it said in the title, I solved the disappearance of the share button

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -767,7 +767,7 @@ public class Pokefly extends Service {
     }
 
     /**
-     * Creates and initializes the level seekbarr in the evolution and powerup prediction section in the results
+     * Creates and initializes the level seekbar in the evolution and powerup prediction section in the results
      * screen.
      */
     private void createExtendedResultLevelSeekbar() {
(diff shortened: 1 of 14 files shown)


nahojjjen:
There are several good small changes to the code here, but Ive got some questions

You've added checks against null for the low and high values in ivScanResult in several places, but I dont quite understand why - are there any documented crashes caused by missing low-high values?


You've added assert statements before the bitmap.recycle methods, whats your reasoning for this?

MARTINI1:
@nahojjjen yes, in ivresultscan had already located those values ​​that cause a null exception, I have sometimes seen this error (which causes closure GoIV), when playing on the street and had prepared the solution but when making updates and being busy I left it for later

and the other, asserts, I do not use the screenshot but I have taken a look and I have noticed that the Android studio itself warns of null exception and I prefer to prevent errors

thearaks:
There are many many changes in this PR... The changes should be, at least, split in different commit otherwise it's very difficult to review the PR!
Also, asserts should be avoided for many reasons (Google suggest to avoid them because the have an inconsistent behavior between different devices) and because they make the app crash when the assertion fails. So, from a user point of view, it doesn't change the end result: the app crashes with or without the assert! :)

nahojjjen:
I'm going to have to agree with thearaks, please split this PR up into more segregated groups. It'll be easier to review & in case we have to revert something, it becomes difficult if things are bundled in the same PR. I'm hesitant to accept all of these changes.

MARTINI1:
close this PR because I have already divided it into other PRs

PR #798, opened 2017-10-17
fix correct detection as nexus6P in ocrhelper

MARTINI1:


nahojjjen:
This certainly looks like an error, can @thearaks comment on if this is intentional? (At least I believe you wrote this?)

MARTINI1:
I suppose not because otherwise it would not exist in Device.java this:

`GOOGLE_NEXUS_6P("Google Nexus 6P", 3.5f, "testinfoscreens/nexus6P",
            makeRect(588, 1109, 256, 67),               // Name
makeRect(572, 1508, 306, 30), // Type`

thearaks:
You are right guys, this is not intentional!
Just a copy-paste error.

Please @MARTINI1, correct also `Device.java` at line 27 to:
`makeRect(960, 1690, 97, 45),                // Candy amount`

Thank you for the review!

thearaks:
The next time it would be nice to squash the two commits with a `rebase -i`, it keeps the commit log more clean ;)

PR #801, opened 2017-10-18
fixed the infinite list of DS_Store files

MARTINI1:
Finally ignored DS_Store files when doing git status

currently with git status

captura de pantalla 2017-10-18 a las 21 46 02

now with this change

captura de pantalla 2017-10-18 a las 21 56 20

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ captures/
 **/*.jks
 keystore.properties
 fabric.properties
+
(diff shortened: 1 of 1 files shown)


thearaks:
It should have been done before; even if I worked on GoIV with a Mac before I never realized that .DS_Store files weren't ignored!
This PR can be merged immediately!

PR #806, opened 2017-10-20
add android O channel notification

MARTINI1:


app/src/main/java/com/kamron/pogoiv/pokeflycomponents/GoIVNotificationManager.java
```diff
@@ -139,10 +144,31 @@ public void showRunningNotification() {
...
+
+    /**
+     * The class which create channel notification for oreo.
+     */
+    private void initNotificationChannel(NotificationManager notificationManager) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && notificationManager
+                .getNotificationChannel(NOTIFICATION_CHANNEL_ID) == null) {
+            // Create notification channel
+            NotificationChannel channel = new NotificationChannel(NOTIFICATION_CHANNEL_ID,
+                    "GoIV", NotificationManager.IMPORTANCE_LOW);
```
thearaks: You should put a translatable string here instead of "GoIV".
Also, the channel name must be more explicative.
I'd set it to "GoIV status" or something similar, so the user knows that disabling this channel in settings will cause the GoIV status notification (running / paused) to disappear.

MARTINI1: the user see this in the notification settings

captura de pantalla 2017-10-20 a las 14 47 57

Do you still want to change it to status ?

thearaks:
Yes, there should be something more than just "GoIV" or the user won't understand what that channel is used for.
Given the reason behind the notification channel feature was added in Android 8.0, itotally no-sense to create a notification channel whose name equals the app name...!

MARTINI1:
@thearaks does not change the name when putting the "goiv status", I think it works if we have more than one channel but if there is only one is the name of the app by logic so as not to confuse the user

thearaks:
It doesn't work because the channel isn't recreated every time.
Uninstall the app and reinstall it and you'll see the channel with the new
name :)

Il 20 ott 2017 3:24 PM, "MARTINI1" <notifications@github.com> ha scritto:

MARTINI1:
@thearaks tested and you're right 👍 

@nahojjjen ready for merge

PR #810, opened 2017-10-26
update graddle tool and others libraries (gradle 4.3.1, lombok 1.16.18, butterknife 8.8.1)

MARTINI1:
update gradle tool to 3.0.1

update gradle to 4.3.1
https://docs.gradle.org/4.3.1/release-notes.html

update google maven repo to google() as oficial announcement
https://developer.android.com/studio/build/dependencies.html#google-maven

change apt to annotationProcessor (android-apt deprectated)
https://bitbucket.org/hvisser/android-apt/wiki/Migration

update lombok and change the dependency as
`dependencies {
	compileOnly "org.projectlombok:lombok:1.16.18"
}`
https://projectlombok.org/setup/android

update butterknife to 8.8.1 version
https://github.com/JakeWharton/butterknife

update guava to 23.5 version
https://github.com/google/guava/releases

add flavors dimens in ProductFlavor (now is required)
https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#flavor_dimensions

PR #811, opened 2017-10-27
Compatibility with Android Studio 3.0.0

thearaks:
Update build.gradle files
Update some dependencies versions

Please merge this when we decide to move to Android Studio 3.0.0


thearaks:
I fear this is very similar to #810

nahojjjen:
since #810 and this were basically the same, and 810 came before, I merged 810.

PR #813, opened 2017-11-06
Correct the size of dot arc

MARTINI1:
BEFORE

AFTER

### DIFF
--- a/app/src/main/res/drawable/dot.xml
+++ b/app/src/main/res/drawable/dot.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <shape xmlns:android="http://schemas.android.com/apk/res/android"
     android:shape="oval">
-    <size android:width="10dp" android:height="10dp"/>
+    <size android:width="8dp" android:height="8dp" />
     <solid android:color="@color/colorAccent" />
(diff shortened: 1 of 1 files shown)


thearaks:
I don't understand the change.
If you want achieve a smaller dot, just change the size to 8dp in _drawable/dot.xml_.
Adding different dp sizes for different densities is **wrong**, dp already scales following the screen density.
In addition, this PR would make GoIV **crash** on mdpi, hdpi and xxxhdpi devices (and every other screen density excluding xhdpi and xxhdpi).
Before posting a change like this, you should test it on all the different possible configurations to ensure it doesn't crashes... :)

MARTINI1:
@thearaks 
but you put the size at 10dp because it looks correct in your resolution (xxhdpi since you use a nexus 5 if I'm not wrong)

thearaks:
@MARTINI1 I didn't set it to 10dp because it looked correct but just because I added some white stroke around it (that measure 1dp on both sides) so, to keep the colored part of the same dimension as before (8dp) I went for a total dimension of 10dp.
It's not related to the screen density :)
Anyway, any UI/UX change should be discussed in Discord #dev channel to avoid confusion and errors!
I'm positive with the change to 8dp if everyone else thinks the same!

PR #814, opened 2017-11-06
Extract hardcode text updating goiv

MARTINI1:
and I have taken advantage of adding the translation of the string into Spanish

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/updater/DownloadUpdateService.java
+++ b/app/src/main/java/com/kamron/pogoiv/updater/DownloadUpdateService.java
@@ -14,13 +14,13 @@ import android.os.IBinder;
 import android.support.v4.content.FileProvider;
 
 import com.kamron.pogoiv.BuildConfig;
+import com.kamron.pogoiv.R;
 
 import java.io.File;
 
 public class DownloadUpdateService extends Service {
 
     public static final String FILE_NAME = "GoIV_new.apk";
-    public static final String DOWNLOAD_UPDATE_TITLE = "Updating GoIV";
     public static final String KEY_DOWNLOAD_URL = "downloadURL";
 
     @Override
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/kamron/pogoiv/updater/DownloadUpdateService.java
```diff
@@ -36,7 +36,7 @@ public int onStartCommand(Intent intent, int flags, int startId) {
                 newApkFile.delete();
             }
             DownloadManager.Request request = new DownloadManager.Request(Uri.parse(downloadUrl));
-            request.setTitle(DOWNLOAD_UPDATE_TITLE);
+            request.setTitle(getBaseContext().getString(R.string.notification_updating));
```
thearaks: _getBaseContext()_ is useless, you could just use _getString(R.string.notification_updating)_ since the Service class extends Context

app/src/main/java/com/kamron/pogoiv/updater/DownloadUpdateService.java
```diff
@@ -14,13 +14,13 @@
...
 
 import com.kamron.pogoiv.BuildConfig;
+import com.kamron.pogoiv.R;
 
 import java.io.File;
 
 public class DownloadUpdateService extends Service {
 
     public static final String FILE_NAME = "GoIV_new.apk";
-    public static final String DOWNLOAD_UPDATE_TITLE = "Updating GoIV";
```
thearaks: Since you removed this constant, you should update also MainActivity.java:141 to:
`if (fileName.equals(getString(R.string.notification_updating))) {`

MARTINI1:
@thearaks this change is difficult to test, i don't know to create a fake update xD

nahojjjen:
You can test updating by building the online version, and setting a lower version in the gradle files. Since the debug will think it's a previous version, it'll try to update to whatever's currently available.

MARTINI1:
thanks for the info @nahojjjen 👍 

tested the online debug v4.1.1 and works correct the "updating" notification

once finished it does not do anything but I imagine it is because it does not find the new apk of online debug

nahojjjen:
It should actually successfully download a downgrade when using the previous testing method, if it doesnt, something might be wrong...

PR #824, opened 2017-12-11
Changes !"quickscan failed" toast to button look

nahojjjen:
Button now shows "?" instead of toasting an error message when pokemon could not be scanned.

Also fixed some formating issues with the button caused by the buttonreset not covering all aspects.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPreviewPrinter.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPreviewPrinter.java
@@ -6,7 +6,6 @@ import android.widget.Toast;
 
 import com.kamron.pogoiv.GoIVSettings;
 import com.kamron.pogoiv.Pokefly;
-import com.kamron.pogoiv.R;
 import com.kamron.pogoiv.ScreenGrabber;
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPreviewPrinter.java
```diff
@@ -73,7 +73,8 @@ public void run() {
             boolean succeeded = runQuickScan();
             Pokefly pokefly = pokeflyRef.get();
             if (!succeeded && pokefly != null) {
-                Toast.makeText(pokefly, R.string.touch_screen_quick_iv_scan, Toast.LENGTH_SHORT).show();
+                ivButtonRef.get().showError();
```
thearaks: To avoid a null pointer exception when the IV button gets garbage collected, ivButtonRef.get() return value should be checked, also Pokefly instance isn't needed anymore since we disabled the Toast and thus there's no need for the Context instance:

```
boolean succeeded = runQuickScan();
IvButton ivButton = ivButtonRef.get();
if (!succeeded && ivButton != null) {
    ivButton.showError();
}
```

PR #841, opened 2018-01-10
OcrHelper v2

thearaks:
This PR is a wide maintenance work on OcrHelper.
It cleans and make more readable its functions and remove legacy code.

The main change is the replacement of all the custom image processing functions with equivalent "standard" implementation from OpenCV library.

Other highlights:
- General: avoid String.replace() calls to increase performances
- Tesseract: fine tune whitelist to improve text detection
- Tesseract: update tess-two library from version 7.0.0 to 8.0.0
- Image hashing: do not hash every px, hash every dp instead
- Image cropping: reuse bitmaps created for each ScanArea to avoid frequent GCs
- Scanning: improve level detection with finer level dot radius estimation
- Unit testing: add LevelDetectionTest test suite


thearaks:
This must be kept on hold until some performance issues are addressed!

PR #855, opened 2018-01-19
Add extra copy to clipboard format for monsters with perfect IV's

Jaap-Jan:
I've added an extra clipboard copy format for monsters which have perfect IV's, as I find showing both the percentage and the individual IV's a bit redundant.

- Add a ClipboardResultMode enum to differentiate between clipboard copy
  modes. This replaces the singleResultMode boolean
- Add extra tab so the user can configure a format for monsters with
  perfect IV's
- Change the logic which determines which clipboard result mode to use
- Removed maxLineLength property from LeftCurly Checkstyle rule, it is
  no longer valid
- Small improvements in JavaDoc as suggested by Checkstyle

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
+++ b/app/src/main/java/com/kamron/pogoiv/GoIVSettings.java
@@ -36,6 +36,7 @@ public class GoIVSettings {
     public static final String DELETE_SCREENSHOTS = "deleteScreenshots";
     public static final String COPY_TO_CLIPBOARD = "copyToClipboard";
     public static final String COPY_TO_CLIPBOARD_SINGLE = "copyToClipboardSingle";
+    public static final String COPY_TO_CLIPBOARD_PERFECTIV = "copyToClipboardPerfectIv";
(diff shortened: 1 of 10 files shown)


thearaks:
Hi!
Thank you for your contribution!

I've not tested yet this branch but, overall, the feature is ok for me. I'll have to test if the clipboard modifier activity tabbar doesn't get too cluttered and all the labels are visible on smaller devices.

I have a suggestion to make: you posted all the changes in a single commit. It's preferable to split the changes into smaller atomic commits so the repository maintenance and workflow is kept easy and lean.

For example, this PR could be split in the following commits:

1) add enum and remove singleResult Boolean

2) change logic to determine which mode to use

3) update clipboard modifier UI

This last change should be in a totally different PR, since it's not related to the topic or, at least, in a separate commit :

4) remove maxLineLenght from LeftCurly rule

By the way, you say you removed it since it's no longer valid, but how we can replace it to keep our code style that wraps lines at 120 chars?

Thank you again,
I'll review the PR as soon as I'll be on my PC!
Good job!

PR #856, opened 2018-01-26
Extract hardcode strings of activity_main

MARTINI1:


thearaks:
Unfortunately these two strings have been removed in #859!
#859 and #860 will make this PR useless, but there will be many new strings that needs to be localized in _es_ so I'll be waiting for your contribution after those two PRs are merged! :+1:

PR #860, opened 2018-01-27
Latest italian translations

thearaks:
Add localizations for recalibration and copy-to-clipboard texts.

Please merge only after #859.


MARTINI1:
please delete the changes in the values-es because I have already corrected the strings and added the new ones of the new design

app/src/main/res/xml/settings.xml
```diff
@@ -35,7 +35,7 @@
         android:title="@string/settings_title_optional">
 
         <SwitchPreference
-            android:defaultValue="false"
+            android:defaultValue="true"
```
MARTINI1: I think I would have to keep on being "false" so that it activates or has activated it only people who want to use this

thearaks: This change isn't introduced by this PR, but instead from #859.
It is needed due to the new position of the clipboard configurator: now it is accessible in the main screen, if it would be disabled and the user started to configure it without knowing it's disabled than he would be disappointed when it would see it not working (Before you could configure clipboard only after enabling it).
So, let it enabled by default so users can test and configure it, than let anyone who dislikes it or doesn't use it free to disable it from settings.
IMHO there's nobody who that after scanning a 'mon doesn't need to "save" the scan results using the clipboard renaming feature...

PR #865, opened 2018-02-06
[Suggestion] Pokedex number view in scan result

udnp:
Hi GoIV pj team.

I have a small suggestion for GoIV.
I think it is useful that the pokedex number is displayed in a scan result.

In Pokemon Go game app, as you know, 
we can use pokedex numbers for filtering pokemons, like following:
device-2018-02-06-002643

If GoIV displays the pokedex number simply and quickly in a scan result,
we can select specific pokemon by this number without going back from Pokedex menu
or switching app and visiting other pokedex info web site.
That's useful, I think.

So with this PR, I suggest the way to display the pokedex number in a scan result.
In my implementation, pokedex number is placed before the pokemon name.

What do you think about this suggestion?
Would you tell me your good idea?

Screen images are following:

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -225,6 +225,8 @@ public class Pokefly extends Service {
     TextView exResCandy;
     @BindView(R.id.exResLevel)
     TextView exResLevel;
+    @BindView(R.id.resultsPokedexNumber)
+    TextView resultPokedexNumber;
     @BindView(R.id.resultsPokemonName)
     TextView resultsPokemonName;
     @BindView(R.id.resultsCombinations)
@@ -1336,6 +1338,7 @@ public class Pokefly extends Service {
      * Shows the name and level of the pokemon in the results dialog.
      */
     private void populateResultsHeader(IVScanResult ivScanResult) {
+        resultPokedexNumber.setText("#" + (ivScanResult.pokemon.number + 1));
         resultsPokemonName.setText(ivScanResult.pokemon.toString());
         resultsPokemonLevel.setText(getString(R.string.level_num, ivScanResult.estimatedPokemonLevel.toString()));
     }
--- a/app/src/main/java/com/kamron/pogoiv/Pokefly.java
+++ b/app/src/main/java/com/kamron/pogoiv/Pokefly.java
@@ -1338,7 +1338,7 @@ public class Pokefly extends Service {
      * Shows the name and level of the pokemon in the results dialog.
      */
     private void populateResultsHeader(IVScanResult ivScanResult) {
-        resultPokedexNumber.setText("#" + (ivScanResult.pokemon.pokedexNumber));
+        resultPokedexNumber.setText("#" + ivScanResult.pokemon.pokedexNumber);
         resultsPokemonName.setText(ivScanResult.pokemon.toString());
         resultsPokemonLevel.setText(getString(R.string.level_num, ivScanResult.estimatedPokemonLevel.toString()));
     }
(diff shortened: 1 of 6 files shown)


udnp:
quick IV preview with pokedex nubmer(Left: Without Pokedex# / **Right: With Pokedex#**)
vs-pokedex-number-view

udnp:
scan result dialog with pokedex nubmer(Left: Without Pokedex# / **Right: With Pokedex#**)
vs-pokedex-number-view-in-scan-result

udnp:
Hi @nahojjjen, Thanks for your comment in my issue.

https://github.com/udnp/GoIV_JP/issues/15#issuecomment-363564059
> Hey, I'm not sure I understand when knowing this number would be useful, could you clarify the use-case a bit?

Examples of use-cases when knowing pokedex number would be useful.
Do these answer your comment?

 - selecting specific pokemons as battle members on the starting battle
 - restore specific pokemons by using potions/revives
 - comparing caught pokemon immediately with the same species caught previously(like following movie)
   device-2018-02-09-043018

These operations become more complicated according to the count of collected pokemons.
So filtering by pokedex number is useful,
but it is hard to memorize a lot of pokedex numbers.

If GoIV could display this number, users are able to know and remember it more easily
than searching in deep Pokedex Menu of Pokemon Go game app.

In these use-cases, using pokemon name also causes the same result.
But I feel using pokedex number can do more quickly and simply.

nahojjjen:
I think you've implemented this well, and I think I can understand your usecases, however I dont think the information should be presented in the quick-iv button. The quick iv button is only for the most essential information. 

By the way, unless I misunderstand, all your usecases are for searching / filtering pokemon, why not write the pokemon name instead?


As a sidenote, you can currently check the pokedex number of a pokemon in GoIV by pressing the iv button, and then pressing the pokemon species name.

udnp:
@nahojjjen thanks for your comment and sharing your thinkings.

> however I dont think the information should be presented in the quick-iv button. The quick iv button is only for the most essential information.

OK, I had not recognized this important UI policy.

I feel this quick-iv preview very useful because information appears without user operations.
So in order to know pokedex number by as few user operations as possible,
I experimented with the quick-iv button.

In that case I could remove this view from the quick-iv button and remake my PR.

> By the way, unless I misunderstand, all your usecases are for searching / filtering pokemon, why not write the pokemon name instead?

Because filtering with pokedex numbers would be faster than with names 
(especially on non alphabetic language Japanese IME 😢). 
Everyday I play many pokemon gym battles, 
and I feel this is the best way to assign my ハピナス(Blissey) to the gym faster than anyone, 
as soon as possible after a battle. 😄

> As a sidenote, you can currently check the pokedex number of a pokemon in GoIV by pressing the iv button, and then pressing the pokemon species name.

Yes I know.
In my GoIV configuration, [Enable scan correction dialog] is turned off 
and [Expanded appraisal opened by default] is turned on.
Therefore in my understanding, to check this number by your way, 
I need one more operation, tapping the [BACK] button before tapping the pokemon species name.

So I wish I could check this number more quickly and simply because of some use-cases I listed in my previous comment.

What do you think the plan to add pokedex number view to only result dialog like following?
If you have any other thinkings or questions, feel free to ask.

device-2018-02-06-003424

nahojjjen:
The UI code is currently being rewritten in preparation for GoIV 5.0.0, so we cant merge a change in that class right now, however the suggestion you sent with the dex# in the resultview in front of the pokemon name looks pretty good. I think it'd be best if we revisited that idea once the 5.0.0 UI code has been merged.

udnp:
Thanks for your positive comment. 😃 
Untill then I'll use my private GoIV build for checking pokedex number.

I have two question.

1) Could I check new UI codes progressing toward version 5.0.0?
   Is the branch "movesetEvakl"?

2) Could this PR#865 be stay open until the 5.0.0 UI code merged?
Or should this be closed once and new other #number PR/issue needed?

nahojjjen:
The (poorly named) branch movesetEvalk was merged to master yesterday

nahojjjen:
This PR can stay up as a reminder for the time being

udnp:
@nahojjjen OK. Thanks for your answer.

PR #873, opened 2018-02-14
Rebalances input UI

nahojjjen:
Adds more whitespace and removes some info, see screenshots


nahojjjen:
Old:
screenshot_20180213-180043

New:
device-2018-02-14-211544

nahojjjen:
If this design is accepted, we need to alter the navigation tabs in the other fractions to match.

PR #874, opened 2018-02-14
Hardcode default scan areas to work with s7

nahojjjen:
alters the existing hardcoded values to work with the s7


nahojjjen:
Do not merge - currently ongoing discussion with making it work with phones like the pixel 1.

Potential problem : onscreen vs no onscreen navigation

nahojjjen:
This would only really make the situation better for myself, and not the average player - it'd be better to generalize the default scan areas instead.

PR #879, opened 2018-02-18
Hard-Codes in a test moveset UI

nahojjjen:
This branch is not currently meant for merging, but so that people can look at the design.


nahojjjen:
image

Several things are missing:

The data is not real

You cannot order the atk/def

The moveset the pokemon has is not marked

The moves which are legacy are not marked

app/src/main/java/com/kamron/pogoiv/pokeflycomponents/fractions/MovesetData.java
```diff
@@ -0,0 +1,94 @@
+package com.kamron.pogoiv.pokeflycomponents.fractions;
```
thearaks: This class should probably be moved into `com.kamron.pogoiv.scanlogic`

app/src/main/java/com/kamron/pogoiv/pokeflycomponents/fractions/PowerTableDataAdapter.java
```diff
@@ -0,0 +1,95 @@
+package com.kamron.pogoiv.pokeflycomponents.fractions;
```
thearaks: While this class should be probably put inside `com.kamron.pogoiv.widgets`

nahojjjen:
@the class placement in packages @thearaks .
Yea, I didnt know exactly which classes was needed for the sortabletableview at first, so I placed them in the same folder, and they morphed quite a bit while I was fixing/implementing the table. At their current shape, they should definitely be in the folders you suggested.

PR #881, opened 2018-02-23
Fix Lvlx2 output when level has .5

pfsmorigo:
First multiply by two and then round it.


thearaks:
Hi @pfsmorigo,
Could you please elaborate the issue? Have you experienced a wrong output from this token?

The (int) cast is not used for rounding, but to make sure that the  String.valueOf call doesn't print any decimal digit. It's just a string formatting thing.
Since the level will always be .0 or .5 multiplying it by 2 always makes it an integer number. So there is no rounding in place.

PR #886, opened 2018-03-03
Initial pokebattler integration

celandro:
This is not quite the right way to do it. It should be a one time
script. It should work if someone wires in the assetmanager and loads
the pokemonMovesetData file.

The script is a bit slow to run so it should really only be run on release of a new version of pokebattler.


celandro:
Updated to push to the correct file location.

Note that it is set to run every time you run unit tests right now. You may want to change it to an actual script in some other module

celandro:
Updated new pokemon names.

Note that the constants files for other languages only have entries if there is a translation for the move. I fall back to english if it is missing

Issue #898, opened 2018-04-23
Back button missing on results screen (to go back to scan correction dialog)


jamesl1001:
I downloaded the latest 5.0.0 update and it looks great! The inclusion of the move sets is a nice addition. Good job!

The only issue I have is that the Back button has gone from the results screen to go back to scan correction dialog.

I have the "Enable scan correction dialog" and "Expanded appraisal opened by default" options switched off so that the results screen shows immediately.

In 4.3.3 I can then tap the Back button to go back to the scan correction dialog and/or appraisal dialog, however in 5.0.0 I cannot.

Thanks!

thearaks:
Fixed in 5.0.1!

PR #899, opened 2018-04-23
Fix navigation between overlay Fractions

thearaks:
Fix navigation in overlay when "should show confirmation" preference is set to off


thearaks:
Fixes #898

nahojjjen:
Seems to work 👍

PR #917, opened 2018-05-04
Replace very slow Bitmap.getPixel with array lookup to improve scan s…

Agilo3:
…peed

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/ocrhelper/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/ocrhelper/OcrHelper.java
@@ -881,10 +881,16 @@ public class OcrHelper {
         Rect currentChunk = null;
         // On devices denser than XHDPI (2x) we can skip a pixel every two (or more) to increase performances
         int increment = (int) Math.max(1, Math.ceil(Resources.getSystem().getDisplayMetrics().density / 2));
+
+        // Get all pixels from the bitmap
+        int[] pixels = new int[cp.getWidth() * cp.getHeight()];
+        cp.getPixels(pixels, 0, cp.getWidth(), 0, 0, cp.getWidth(), cp.getHeight());
+
         // When we're over a chunk check every pixel instead of skipping so we're sure to find the blank space after it
         for (int x = 0; x < width; x += (currentChunk != null) ? 1 : increment) {
             for (int y = 0; y < height; y += increment) {
-                final int pxColor = cp.getPixel(x, y);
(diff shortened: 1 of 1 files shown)


thearaks:
Hi Alessandro,
Thanks for your contribution!

I have just 1 question and 1 change request:

- Have you measured the performance difference between the `getPixel` version and your proposed `getPixels` one?
- For every execution of `getPokemonCPFromImg` you are allocating a new int[] as large as the CP cropped image. This means that, especially for high-density screens, heavy memory allocations are done too frequently, causing unneeded GCs.
In example, on a Galaxy S7 (4x density) the CP crop area is around 296*88px. At 32bpp it means you are allocating 296*88*4bytes = 101,75KB.
This happens for every scan but especially for each quickscan (and could be retried trice for each of this scan if it fails at first attempts).

This has a much greater performance improvement instead of just relying on the `getPixel` calls.
Don't forget that the getPixel implementation is made at native level, and since it's just a RAM access operation it's not that slow.

Given that this allocate-and-copy-memory is all around OcrHelper, in the past I proposed a solution for this.
Unfortunately the commit that implements a sort of caching for these croppings is parked inside a greater PR that, at the moment, is in stand-by because it needed further performances investigations.

The PR I'm speaking about is #841 
And the specific commit that mitigates the useless memory allocations is da2058edd2e0542f3ea0220feeafd40ea0cd98a9 where I introduce the `private static LruCache<ScanArea, Bitmap> cropBitmapCache;`

Anyway, we should understand if copying all the CP crop bitmap memory to a int[] with `getPixels` (even without reallocating it each time) is faster that a bunch of `getPixel` calls or not.
So, please, post some performance analysis on this matter so we can decide if apply this methodology to all OcrHelper code (or not).

Thank you again!
Mattia

Agilo3:
Hey Mattia,

What I did was use the profiler in Android Studio and saw that `Bitmap.getPixel()` was being called a lot. It is taking up around 25% of the time it takes to scan a Pokemon. When I looked at how I could improve this alone I found various sources ([1](https://stackoverflow.com/a/19454882), [2](https://stackoverflow.com/a/30559570), [3](https://stackoverflow.com/a/4716662), [4](https://stackoverflow.com/a/24811635), etc) which state that it's faster to get all the pixels at once and reference them in the int array. This is actually also what has been done in other  places such as `replaceColors()`. The tradeof is of course a slightly larger memory footprint, but I was under the impression that the garbage collector would take care of this efficiently enough.

Initially I used only the emulator. As it turns out, comparing the profiling data from before and after this change and running it on actual hardware (Galaxy S7), there is some improvement but not as much as I initially perceived.

I have attached the screenshots of the profiling data:
_before
_after

Based on the data I would be at peace closing this pull request.

Issue #908, opened 2018-04-24
Can't turn off movesets tab


Agilo3:
Hi,

I don't see an option for disabling the Movesets tab. Could this be added?

Thanks!

nahojjjen:
Hey, why do you want this option?

Agilo3:
It's not useful information for me, as I mainly use Gamepress for moveset info when I need it. It's just bloat to me. Being able to turn it off saves an extra tab space on top too.

We're also able to turn off other features like evolve count, or are able to adjust behavior like automatically opening the appraisal screen by default. An extra toggle for turning off the moveset info tab doesn't seem like an overly big request. 😄

thearaks:
Unfortunately it's harder then it might look from the outside.
It would require a change to the settings (update the key-values definitions, update the translations for the preference title and description), adding the logic to avoid loading the moveset data and avoiding using it everywhere it is used at the moment to avoid nasty crashes (scan result UI, clipboard rename string generation, ...), and the scan result UIs shoud be changed accordingly to remove the tab.
In my opinion it's a medium-difficulty task, with lower priority compared to other tasks we have to implement...

nahojjjen:
Im closing this issue, I understand why you want it, but unfortunatelly the downsides are bigger than the pros.

The freed up tab-space would not give you anything but more blank space, which is aestetically pleasing, but not valuable enough to warrant the added code complexity for handling another possible UI combination.

PR #919, opened 2018-05-04
Allow users to turn off moveset info

Agilo3:
This will give users the option to hide moveset info. Saves a bit of resources and allows people to simply turn it off if they don't care about the integration in GoIV.


thearaks:
There are a couple of things I'd like to review, but I'm busy at the moment.
There's a change in the clipboard code that probably brakes the configuration saving/restoring to preferences.
I'll post a review later,
in the meanwhile thank you!

Agilo3:
Did a quick rebase onto master just to keep things tidy.

Also, I did test the clipboard functionality when turning the option on/off. I'll be interested to hear your review.

This pull request is the first Android development I've done so far so I value your feedback; thanks! I'm also available on the Discord channel for any direct messages (either PM or in #dev).

app/src/main/java/com/kamron/pogoiv/Pokefly.java
```diff
@@ -323,6 +325,10 @@ public int onStartCommand(Intent intent, int flags, int startId) {
         return START_REDELIVER_INTENT;
     }
 
+    private boolean isMovesetEnabled() {
```
thearaks: This new method can be avoided if possible, I'd prefer using `GoIVSettings.getInstance(this).isMovesetEnabled()` directly!

app/src/main/java/com/kamron/pogoiv/activities/ClipboardModifierParentFragment.java
```diff
@@ -59,6 +59,21 @@ public View onCreateView(@NonNull LayoutInflater inflater,
         return view;
     }
 
+    @Override
+    //Pressed return button - in the case when coming back from settings screen after movesets have been enable/disabled
+    public void onResume() {
+        super.onResume();
+
+        if (viewPager != null)
+        {
```
thearaks: This curly bracket should be on the previous line ;)

app/src/main/java/com/kamron/pogoiv/activities/ClipboardModifierParentFragment.java
```diff
@@ -59,6 +59,21 @@ public View onCreateView(@NonNull LayoutInflater inflater,
         return view;
     }
 
+    @Override
+    //Pressed return button - in the case when coming back from settings screen after movesets have been enable/disabled
+    public void onResume() {
```
thearaks: Avoid all this behavior, all this added complexity just to remove the moveset clipboard tokens from the list is overkill.
If a user have deactivated the moveset functionality in the settings, he would probably just ignore the movesets tokens knowing that they won't be effective.
Also, moveset tokens should already work when the movesets info isn't present (i.e. when the moveset scan fails).

app/src/main/java/com/kamron/pogoiv/clipboardlogic/ClipboardTokenCollection.java
```diff
@@ -124,11 +127,12 @@
         tokens.add(new HexIVToken()); //hex representation of iv (ex A4B)
         /////////////////////////////////////////////////////////
 
-
-        //Moveset//////////////////////////////////////
-        tokens.add(new MovesetInitialsToken(1));
-        tokens.add(new MovesetInitialsToken(2));
-
+        //enable/disable visibility based on setting
+        if (GoIVSettings.getInstance(context).isMovesetEnabled()) {
```
thearaks: This method and the list it returns shouldn't be changed by the user's preferences because this same list is used to read/write tokens configuration into sharedpreferences.
This list should never change its contents (aside from new token additions when a new app version is released) or the token configuration logic would fail.

app/src/main/java/com/kamron/pogoiv/pokeflycomponents/ocrhelper/OcrHelper.java
```diff
@@ -1130,15 +1132,18 @@ public ScanData scanPokemon(@NonNull GoIVSettings settings,
...
-                estimatedLevelRange,
-                ScanArea.calibratedFromSettings(POKEMON_POWER_UP_CANDY_COST, settings),
-                ScanArea.calibratedFromSettings(POKEMON_EVOLUTION_COST_AREA, settings));
-        String moveFast = null;
-        String moveCharge = null;
-        if (moveset != null) {
-            moveFast = moveset.first;
-            moveCharge = moveset.second;
+
+        String moveFast = "";
```
thearaks: If the moveset scan fails `moveFast` and `moveCharge` should be null, otherwise `ScanData` / `ScanResult` data structures would behave like the moveset scan was successful. Please initialize this variables to null instead of empty String

app/src/main/res/values-cs/strings.xml
```diff
@@ -59,4 +59,6 @@
     <string name="token_pokemonname">Jméno Pokémona</string>
     <string name="clipboard_modifier_preview">Náhled</string>
     <string name="position_handler_toast">Přemísti okno táhnutím tohoto držáku.</string>
+    <string name="moveset_enabled_setting_summary" />
```
thearaks: Please do not declare translation keys without values

thearaks:
Hi @Agilo3,
Even if it's your Android development I see that you're very familiar with Java coding and gitflow.
Your code is definitely high-quality!
In fact, my review is mainly based on "how we use to make things here" since there's almost anything wrong or not working in your PR!

I'm not very active on Discord lately but I'd like to meet you there soon :)

nahojjjen:
I'd like to point out that, while the code is high quality, its for a feature that I've rejected in #908 discussion. Thearaks seems to have started a discussion about changes for this PR with you,  but in the future it'd be highly preferred if you discussed in Github issue threads or in the discord dev channel before creating PRs.

I'm however impressed that you managed to understand the GoIV codebase well enough to actually make this change.

PR #926, opened 2018-06-01
patch for  Alola Exeggutor and other Alolans

udnp:
Hi GoIV pj. team.

I made a patch for Alola Exeggutor and other forms.
Could you check this PR's patch?

device-2018-06-01-173602

device-2018-06-01-173653

device-2018-06-01-173719

P.S.
I think it is better in the future to do refactoring Pokemon data mode like discussed in the pr #878.
On the other hand, this patch's way is similar with current implementation for evolution line pokemons,
therefore easy to add quickly other Alola form pokemons. 

Thanks.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPopupButton.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/IVPopupButton.java
@@ -117,12 +117,16 @@ public class IVPopupButton extends android.support.v7.widget.AppCompatButton {
             int high = highest.percentPerfect;
 
             final StringBuilder text = new StringBuilder();
+            String pokemonName = scanResult.pokemon.name;
+            if (pokemonName.contains(" - ")) { // check including form name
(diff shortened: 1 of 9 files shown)


nahojjjen:
Ive hard-coded in alolan forms which Ive discussed in the discord channel, which means merging this branch directly would cause conflicts. However this way of adding alolan forms seems to be more stable in the long run, so I think its worth discussing the merits of having a forms column added to the data as you've done here. Please do contribute to discussion for how to store the alolan data in the discord.

Another suggestion mentioned would be to turn the 'pokedex' list in PokeInfoCalculator into a list containing lists of pokemon, with different forms.

udnp:
@nahojjjen, Thanks for your information.
I've not known the Discord channel, sorry.
I'll try to join into this channel.

nahojjjen:
Sorry, Ive added a link to the discord channel in the contributions.md file now.

udnp:
Because #931 merged instead of this, I closed this.

PR #931, opened 2018-06-05
Changes the structure for alolan database

nahojjjen:
This is #926 , but fixed to avoid merge conflicts. Sorry udnp


udnp:
Don't worry, @nahojjjen 
Thanks for your fixing and merging my patch.
I've done checking that this commit works well on my local debug build.

PR #937, opened 2018-06-09
autocomplete view with alolan forms for user input

udnp:
Hi GoIV pj. team.

I created a patch for autocomplete view with aloan forms.
Would you check this patch?

autocomplete-view

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/fractions/InputFraction.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/fractions/InputFraction.java
@@ -244,7 +244,7 @@ public class InputFraction extends Fraction {
      * Initialises the autocompletetextview which allows people to search for pokemon names.
      */
     private void initializePokemonAutoCompleteTextView() {
-        String[] pokeList = pokeInfoCalculator.getPokemonNamesWithFormsArray();
+        String[] pokeList = pokeInfoCalculator.getPokemonNamesWithFormArray();
         ArrayAdapter<String> adapter = new ArrayAdapter<>(pokefly, R.layout.autocomplete_pokemon_list_item, pokeList);
         autoCompleteTextView1.setAdapter(adapter);
         autoCompleteTextView1.setThreshold(1);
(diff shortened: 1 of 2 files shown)


udnp:
NOTE:
On my local build, I've confirmed that GoIV always stops on selecting autocompeled pokemon name and tapping [CHECK IV] button with following exception.
But this exception occurs before v5.0.3 alolan structure added.
So I think this exception is other issue for this patch.

```
    java.lang.NullPointerException: Attempt to invoke interface method 'boolean android.view.inputmethod.InputConnection.finishComposingText()' on a null object reference
        at android.view.inputmethod.InputConnectionWrapper.finishComposingText(InputConnectionWrapper.java:150)
        at com.android.internal.view.IInputConnectionWrapper.executeMessage(IInputConnectionWrapper.java:431)
        at com.android.internal.view.IInputConnectionWrapper$MyHandler.handleMessage(IInputConnectionWrapper.java:93)
        at android.os.Handler.dispatchMessage(Handler.java:102)
        at android.os.Looper.loop(Looper.java:241)
        at android.app.ActivityThread.main(ActivityThread.java:6274)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:886)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:776)
```

nahojjjen:
I'm not at home, and cant access my dev environment right now, so I'll test building it later, but reading the code it looks like a reasonable fix. Although, I'd like if you split up the assignment/return.
>  return pokeNamesWithForm = pokemonNamesArray.toArray(new String[pokemonNamesArray.size()]);

Would be more clear as
> pokeNamesWithForm = pokemonNamesArray.toArray(new String[pokemonNamesArray.size()]);
return pokeNamesWithForm;



I can't reproduce the crash issue you encountered. Could you describe the steps to recreate the crash in more detail, or is the issue specific to your current locale settings?

udnp:
@nahojjjen Thanks for your advice.
I've improved and updated this PR, according to your advice.

udnp:
> I can't reproduce the crash issue you encountered. Could you describe the steps to recreate the crash in more detail, or is the issue specific to your current locale settings?

It seems that this exception issue occurs **only on the debug build**, does not occur on the release build.
I've not still the reason why so.
I'm checking this on English locale system.

udnp:
It seems that this exception issue is caused by following Android Studio 3.x bug.
https://issuetracker.google.com/issues/77764953
https://stackoverflow.com/questions/49512975/android-studio-3-1-edittext-stackoverflowerror

So now I think this exception issue is not caused by GoIV.

udnp:
I've fixed the problem that toast indicating "was not found" appears when user selects a pokemon name in the autocomplete view.

nahojjjen:
Tested, and looks good to me.

Issue #947, opened 2018-06-24
Still can't scan Meowth - v5.0.8


Netfreak25:
Version: 5.0.8
Android: 8.0.0
Device: Samsung S7 edge
Language: German

Failures: 15 of 15

Problem:
Can't scan Meowth, not manually and not automatically.
![screenshot_20180624-114107_pokmon go](https://user-images.githubusercontent.com/6755265/41817885-0fd547aa-77a4-11e8-8b61-2641e4328c15.jpg)
![screenshot_20180624-114044_pokmon go](https://user-images.githubusercontent.com/6755265/41817886-10056aa2-77a4-11e8-8817-81d2df410e2d.jpg)
![screenshot_20180624-114114_pokmon go](https://user-images.githubusercontent.com/6755265/41817887-10353f84-77a4-11e8-8fb0-0f3009d3ed25.jpg)

Values are detected properly! Still no result!

Second Problem:
Wrong IV calculation. App says 91% but ingame info suggests its <82%
![pokemon go_2018-06-24-11-48-43](https://user-images.githubusercontent.com/6755265/41817922-b2ca6ae4-77a4-11e8-8a16-c39f973825d7.png)

If you need more screenshots or information, just let me know

canavan:
You need to pick the Alola form manually. The Alolan forms have different base stats, and therefore, even if a matching solution for the given CP/HP/Level is found, it'll be incorrect. In the topmost screenshot, tap the pen icon to the right of "Mauzi", type alola and pick (the german equivalent of ) "Meouth - Alola Form"

udnp:
@Netfreak25 thanks for your issue report.

I think in German locale the words for pokemon types have not been localized yet,
therefore these can not be detected automatically.

So could you let us know the localized/translated words for German pokemon types, like default English DARK to German UNLICHT?

Hi, GoIV pj team,
pokemons.xml files in the v5.0.8 are following.
So there is not pokemon `typeName` resource for German locale.

In German locale
https://github.com/farkam135/GoIV/blob/8ff422b7b15a51ae3d14b83d95c9872f97fe2caa/app/src/main/res/values-de/pokemons.xml#L391-L393

In default English locale
https://github.com/farkam135/GoIV/blob/8ff422b7b15a51ae3d14b83d95c9872f97fe2caa/app/src/main/res/values/pokemons.xml#L392-L414

Netfreak25:
Canavan was right, that way it worked!

[...]

PR #958, opened 2018-07-16
German Pokemon Types

udnp:
for issue #947.
I've checked it can detect eevees evolutions and some alolan forms on German locale system.


nahojjjen:
👍

PR #959, opened 2018-07-17
add in udnps comments

nahojjjen:


udnp:
Thanks for your work.

And I have a tiny FYI.
I think in this case instead of creating a new PR, you could just merge directly commit e2d2d5e to the master.

PR #961, opened 2018-07-27
Support for changing scan areas for Lucky Pokemon

FoxFireX:
Added an alternate version of the region scan computation which can offset vertically for lucky pokemon. Retained the existing signature for backward compatibility, as well as for use in fields unaffected by lucky status. When the first scannable area that would be affected by Lucky is encountered, if no result is returned, it will retry with the lucky offset; if this returns a value, the pokemon is assumed to be lucky, and further scans below the line will be similarly offset. Using an offset of 120% of the power up candy scan height, just as a "feels right" value. Needs testing against other devices. Works spot on for Pixel.

### DIFF
--- a/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/ocrhelper/OcrHelper.java
+++ b/app/src/main/java/com/kamron/pogoiv/pokeflycomponents/ocrhelper/OcrHelper.java
@@ -1103,8 +1103,12 @@ public class OcrHelper {
         // If no power up cost was found, check by offsetting down by the height of the
         // "LUCKY POKEMON" string, which is slightly higher than the power up candy cost field
         if (!powerUpCandyCost.isPresent()) {
-            int tempLuckyOffset =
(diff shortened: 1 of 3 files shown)


thearaks:
Hi @FoxFireX!
Looks good to me!
I'll try to test it ASAP on my phones. Have you checked if something bad happens if a ScanArea gets pushed too much to the bottom and gets out of the Bitmap? (is it even possible?)
There's a checkstyle issue, could you fix that?

FoxFireX:
@thearaks Was trying to find where the checkstyle issue is reported; can you help me track it down? As for the scan area dropping too low, I hadn't considered that initially, but then looked into it after your comment. The getImageCrop method actually sanity checks that the areas are within the bounds of the screen image, and if not looks like it sends an exception/debug message through Timber (which I'm not familiar with) before returning null.  I believe it should be safe.

thearaks:
Thank you @FoxFireX for your check!
I remember something like that but your double-check reassured me.
I'll try to take a look to the checkstyle later to see what's wrong! Thank you!

PR #987, opened 2018-10-15
Updates poke stats

nahojjjen:


nahojjjen:
We'll probably use the generator again once this change goes through.

PR #988, opened 2018-10-16
Fixed Deoxys form array entries

FoxFireX:
Formes are handled using separate arrays of values. Deoxys had its four formes added into the main list of pokemon information which should only have had a single base instance, relying on the forme arrays for replacement information as necessary.

Note that this likely will still pose a problem when the additional formes are added in the game, as I don't believe the logic will be able to distinguish between them, unless the Pokemon name gives us a hint. It will likely be necessary to refactor the logic to allow multiple pokemon guesses to be evaluated by the IV engine; this should allow the correct forme to be determined in this instance.


nahojjjen:
this change was included in the gen 4 stat addition, thanks

PR #995, opened 2018-10-20
Speed optimizations

nahojjjen:
On my samsung s7, quickIV scan time went from 1.2 seconds to 0.6 seconds, and pressing the IV button went from ~8 seconds to ~2


udnp:
Sorry for this performance issue caused by my previous changes.
I had not checked that the init of `InputFraction` is called everytime quick scan runs.
Thanks for your commit c4e30c5 to solve this serious performance issue.

I've investigated this performance on my device Xperia (worked with Snapdragon 650 CPU) with ARM 64 bit build APK.
Its duration is about 0.6 sec. on my device.

PR #996, opened 2018-10-20
Fix incorrect values and add missed in PokeSpam

MARTINI1:
Fix wrong values (zubat or horsea for example) and add missed values in the 3 & 4 generation


nahojjjen:
Hey, didn't notice you made this PR, I just made something very similar in #997  ^^

PR #999, opened 2018-10-22
change repository definition order so 'common' resolves from google before jcenter is checked

harkin:
When I pull master and try to build I get this gradle error: 
```
Failed to resolve: common
Open File
```
Changing the order that the repositories are specified lets it check google first for that dependency and find it. Curious who else sees this; I don't have anything in my local caches (which it would check first).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -2,8 +2,8 @@
 
 buildscript {
     repositories {
-        jcenter()
         google()
+        jcenter()
     }
     dependencies {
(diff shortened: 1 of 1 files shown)


xZise:
This has been already merged as 3215aca, when I was trying to fix my builds.

thearaks:
Closing since already applied through @xZise PR #1046!
Thank you @harkin anyway for reporting this problem

PR #1000, opened 2018-10-22
update gradle build tools and gradle version used

harkin:
`3.2.1` requires gradle 4.6 or above and adds a bunch of build errors and warnings.

It has a default `buildToolsVersion` now so specifying one is no longer necessary. it also errors on sdk versions being specified in the manifest

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -72,7 +72,6 @@ android {
      * Update (both Gradle and local SDK) whenever possible
      */
     compileSdkVersion 27
-    buildToolsVersion '27.0.3'
 
     /**
      * Build configurations for APK
--- a/openCVLibrary330/build.gradle
+++ b/openCVLibrary330/build.gradle
@@ -2,7 +2,6 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion "27.0.3"
 
     defaultConfig {
         minSdkVersion 19
--- a/openCVLibrary330/src/main/AndroidManifest.xml
+++ b/openCVLibrary330/src/main/AndroidManifest.xml
@@ -2,7 +2,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
       package="org.opencv"
       android:versionCode="3300"
-      android:versionName="3.3.0">
-
-    <uses-sdk android:minSdkVersion="8" android:targetSdkVersion="21" />
-</manifest>
+      android:versionName="3.3.0" />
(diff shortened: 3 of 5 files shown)


app/build.gradle
```diff
@@ -72,7 +72,6 @@ android {
      * Update (both Gradle and local SDK) whenever possible
      */
     compileSdkVersion 27
-    buildToolsVersion '27.0.3'
```
thearaks: This shouldn't be set to `buildToolsVersion "28.0.3"`?

openCVLibrary330/build.gradle
```diff
@@ -2,7 +2,6 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion "27.0.3"
```
thearaks: This shouldn't be set to `buildToolsVersion "28.0.3"`?

openCVLibrary330/src/main/AndroidManifest.xml
```diff
@@ -2,7 +2,4 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
       package="org.opencv"
       android:versionCode="3300"
-      android:versionName="3.3.0">
-
-    <uses-sdk android:minSdkVersion="8" android:targetSdkVersion="21" />
-</manifest>
+      android:versionName="3.3.0" />
```
thearaks: This change is a bit unrelated to the topic of this PR but it's OK.
Next time I'd put it, at least, in a separate commit.

harkin: It's not, as mentioned in the PR description build tools `3.2.1` added a bunch of warnings and errors. One was an error when SDK versions are specified in the manifest. If I remember right it actually only errors on `targetSdkVersion` and `minSdkVersion` was a warning but I figured removing one would be a less clear change

openCVLibrary330/build.gradle
harkin: Nope, as mentioned in the PR description the new gradle build tools have their own internal default for `android.buildToolsVersion`. For gradle build tools `3.2.1` that default is `28.0.*`. Not specifying this is fine and reduces the overhead to keeping an app up to date - keep gradle build tools up to date and you'll be up to date with android build tools too

app/build.gradle
harkin: See comment below 👍

openCVLibrary330/src/main/AndroidManifest.xml
harkin: I split it out anyone into https://github.com/farkam135/GoIV/pull/1006, if that gets merged first I'll rebase this branch 👍

PR #1005, opened 2018-10-29
Fix NPE while making the screenshot bitmap mutable

nkming2:
sCalibrationImage is null at this point, causing NPE when invoking getConfig()

stack trace:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'android.graphics.Bitmap$Config android.graphics.Bitmap.getConfig()' on a null object reference
	at com.kamron.pogoiv.activities.OcrCalibrationResultActivity.startCalibration(OcrCalibrationResultActivity.java:108)
	at com.kamron.pogoiv.ScreenShotHelper$1.onChange(ScreenShotHelper.java:81)
	at android.database.ContentObserver.onChange(ContentObserver.java:145)
	at android.database.ContentObserver$NotificationRunnable.run(ContentObserver.java:216)
	at android.os.Handler.handleCallback(Handler.java:789)
	at android.os.Handler.dispatchMessage(Handler.java:98)
	at android.os.Looper.loop(Looper.java:164)
	at android.app.ActivityThread.main(ActivityThread.java:6809)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.Zygote$MethodAndArgsCaller.run(Zygote.java:240)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:767)
```


app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
```diff
@@ -105,7 +105,11 @@ public static void startCalibration(@NonNull Context context,
             sCalibrationImage = bitmap;
         } else {
             // Make a mutable copy of the bitmap so we can draw on it with a Canvas
-            sCalibrationImage = bitmap.copy(sCalibrationImage.getConfig() ,true);
```
thearaks: Please replace this line with this instead:

`sCalibrationImage = bitmap.copy(bitmap.getConfig(), true);`

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
```diff
@@ -105,7 +105,11 @@ public static void startCalibration(@NonNull Context context,
...
+            sCalibrationImage = bitmap.copy(Bitmap.Config.ARGB_8888 ,true);
```
thearaks: Using `Bitmap.Config.ARGB_8888` might be a waste of space if the screenshot bitmap used a lower color configuration.
See the previous suggestion to copy the configuration from the screenshot bitmap.

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
```diff
@@ -105,7 +105,11 @@ public static void startCalibration(@NonNull Context context,
...
+            if (sCalibrationImage == null) {
```
thearaks: This check is useless since the copy can't return a null bitmap.
Please remove this added lines since they're not helpful.

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
thearaks: Please replace this line with this instead:

`sCalibrationImage = bitmap.copy(bitmap.getConfig(), true);`

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
thearaks: Using `Bitmap.Config.ARGB_8888` might be a waste of space if the screenshot bitmap used a lower color configuration.
See the previous suggestion to copy the configuration from the screenshot bitmap.

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
nkming2: It could be NULL according to the reference:
```
If the conversion is not supported, or the allocator fails, then this returns NULL
```
but yea, I think it's a pretty rare case and maybe just doesn't worth bothering at all

app/src/main/java/com/kamron/pogoiv/activities/OcrCalibrationResultActivity.java
nkming2: That's also what I thought originally but I checked the reference and realized getConfig() could actually return NULL.
```
If the bitmap's internal config is in one of the public formats, return that config, otherwise return null.
```
Maybe I just make an if there

nkming2:
Amended

PR #1007, opened 2018-10-30
fix checkstyle issues in OcrManualCalibrationActivity

harkin:
The Ci step in Travis is failing on checkstyle all the time.

The biggest culprit is `src/main/java/com/kamron/pogoiv/thirdparty/pokebattler/PokemonId.java` with 2464 style violations. `OcrManualCalibrationActivity` was in at #2, I've fixed them all. 

I'll clean up the other files with style violations, thinking it might be worth adding a checkstyle suppression for PokemonId.java so that CI turns green 

It was mostly whitespace issues, one line length and two for each variable declaration going on it's own line.


nahojjjen:
Adding a checkstyle suppression for pokemonID would make a lot of sense. It's a copy of a file given from Pokebattler, so whenever it needs updating the file gets copy-pasted from pokebattler files, re-adding the checkstyle errors.

PR #1015, opened 2018-11-16
detection for pokemons Meltan and Melmetal (quick hardcoded patch)

udnp:
This is a quick hardcoded patch.
but dose not need to change `integers.xml` and its generator.

tested with a pokemon renamed as Meltan,
because I still don't have these new pokemons.

pr-meltan


udnp:
This patch has a minor issue that their pokedex numbers appears in the following picture are not correct.
Meltan pokedex number is 808, and Melmetal 809.
pr-meltan-minor-issue

Issue #1016, opened 2018-11-27
Values in tag 'UIV-mixed-07' are showing incorrect values


BigCabbage:
There is a problem in this calculation (within module MixedUnicodeToken.java). The results do not match the description, as they currently show the IV values of the lowest IV combination, instead of the individual lowest atk/def/hp IVs.

When a stat, like HP has for all combinations a value of 15, that number is shown as black 15, which is fine. The defense has different numbers, so it should be white and show the lowest number, for example 6. However, while I noticed my attack has a range of 11-15, it should be white 11, not white 15 as currently assigned.

I created this issue so I can link it when presenting the solution through a PR.

The return string currently returns the IV values of the lowest IV combination, instead of the individual lowest atk/def/hp IVs (MixedUnicodeToken.java:102-105)

The code:
```
return attToUse[lowestIVCombination.att]
   + defToUse[lowestIVCombination.def]
   + staToUse[lowestIVCombination.sta];
```

should become:
```
return attToUse[lowestAttackStat]
   + defToUse[lowestDefenseStat]
   + staToUse[lowestStaminaStat];

```

PR #1017, opened 2018-11-27
MixedUnicodeToken: Returnvalue now matches token description in GoIV

BigCabbage:
This PR addresses https://github.com/farkam135/GoIV/issues/1016, which has also been discussed on Reddit: https://www.reddit.com/r/GoIV/comments/a0u2yo/problem_with_uivmixed07_display_of_calculated/

It fixes the returnvalue to be in line with the description in the app GoIV, which states that it returns the lowest value for each stat. Instead, it currently returns the lowest IV combination.


BigCabbage:
I checked this failure, it has absolutely nothing to do with my patch.

nahojjjen:
Seems to be working

BigCabbage:
Thanks a lot, removed the no-longer needed branch. 
Will see if I contribute more.

PR #1024, opened 2018-12-16
Updates pokemon stats

nahojjjen:


udnp:
Thanks for your working.
I've checked out this branch and confirmed to build ok.

but it seems it is failed when I've checked to detect IV for Ninjask and Alolan Diglett.

If reverted it changes success, so PoGo server has not distributed new data yet.  
When reverted it changes success to detect, so it seems PoGo server has not distributed new data yet.

P.S. I can not check for Meltan, because my meltan has not returned yet from Gym...

udnp:
As discussed in Discord,
Meltan and etc which have already existed in users pokemon box are not updated to their new stats until they are powered up or evolved.
So it seems we need to judge whether to update xml files for new stats or not.

xZise:
> but it seems it is failed when I've checked to detect IV for Ninjask and Alolan Diglett.

This is interesting because as I mentioned in #1027, I actually think that the numbers for teh Alolan Diglett here are actually correct. If I assume attack 14, defense 15 and stamia 14, and manually calculate the CP using the formula given on [pokemongo.gamepress.gg](https://pokemongo.gamepress.gg/pokemon-stats-advanced#calculating-cp) I get:

    CP = ((108 + 14) * (81 + 15)^0.5 * (67 + 14)^0.5 * 0.59740001^2) / 10 = 383.944462
         ((108 + 14) * (82 + 15)^0.5 * (67 + 14)^0.5 * 0.59740001^2) / 10 = 385.938992

And my Diglett has CP 383, so it looks to me the fixed value here is actually correct (if it gets round down).

udnp:
@xZise thanks for your feedback.

> > but it seems it is failed when I've checked to detect IV for Ninjask and Alolan Diglett.
>
> And my Diglett has CP 383, so it looks to me the fixed value here is actually correct (if it gets round down).

My that Alolan Diglett had been caught before its base stats was updated.

> Meltan and etc which have already existed in users pokemon box are not updated to their new stats until they are powered up or evolved.

Alolan Diglett is also the same.

udnp:
Currently we don't have solutions because current GoIV can have just one set of stats values for a pokemon.
So now GoIV can not have histories of pokemons stats.

PR #1044, opened 2019-04-20
Adds a CP% token.

kingpin003:
This token is used to show what percentage between 0% and 100% IV the Pokemon's CP will be at L40.


nahojjjen:
This PR adds a CP% token, but also adds an appraisal toggle to filter weather boosted pokemon, and also adds a CP indicator for the next powerup in the CP list.

I'd personally accept the CP% commit, but I'd need to have a discussion about the pros/cons of the two other changes before merging.

kingpin003:
Sorry about that. Didn't realize the other commits I made would combine into this PR. Closing and re-submitting just the token commits.

PR #1076, opened 2019-07-19
Support new Pokemon Go version

nahojjjen:
As a first quick patch, it only modifies how it reads the appraisal.

Makes a PR so this change can be discussed and reviewed


Ershany-IBM:
Any update on when this will be merged?

xZise:
I think it could be merged in this state now.

PR #1080, opened 2019-09-23
Add Gen 5

### LINKED ISSUE #1079: Doesn't work with Unova pokémon
These were added recently and do not work with GoIV

KuSh:
I didn't test it ATM so I can't say if it works or not. I'll try to test it during the week.

Should fix  #1079.


KuSh:
Ok. Fixed a few things and I can now scan Gen 5 IVs !
Enjoy

PR #1084, opened 2019-12-09
Don't ANR on Q

luca020400:


nahojjjen:
Hey,  does your title refer to an "application not responding" android error?

  I don't personally have an android Q phone, so I cannot test / verify that the PR solves the issue. Could you briefly describe the issue this PR solves, and how?

luca020400:
It's a more deeper issue within the media projection stack, seems like Android doesn't really like creating virtual displays on the main thread.
Google is currently tracking this bug internally but I have no real answer yet.

luca020400:
Android itself deadlocks leading to an ANR ( Android not responding )

PR #1094, opened 2020-12-24
Fix GAME_MASTER parsing

lramati:
Updates the gradle script to support V2_GAME_MASTER.json
Cleans up the code for the generation by a lot, way fewer loops.

Couple other small things too

NOTE: No Idea how to get the style stuff to happen, maybe this needs a style pass before committing 🤷‍♂️ 

Supercedes #1089


app/src/main/res/values/forms.xml
```diff
@@ -1,1000 +1,966 @@
...
-  <!--Genesect-->
-  <item>252</item> <!-- Normal Form -->
-  <item>252</item> <!-- Douse Form -->
-  <item>252</item> <!-- Shock Form -->
-  <item>252</item> <!-- Burn Form -->
-  <item>252</item> <!-- Chill Form -->
- </integer-array>
+ <!-- Rattata -->
+ <!-- Rattata -->
+ <!-- Rattata -->
```
xZise, line 138: Why does the name appear three times?

app/build.gradle
```diff
@@ -204,6 +204,10 @@ dependencies {
     implementation 'com.android.support:palette-v7:27.1.1'
     implementation project(':openCVLibrary330')
 
+    /* for devMethods.thirdparty.pokebattler, temporally set for All Builds */
+    implementation 'org.json:json:20160810' //used to generate moveset list
```
xZise, line 208: I needed to comment this to actually be able to build the "offlineRelease" build. It mentioned that Android comes with it built in. It also uses GSON, so maybe use that instead?

app/src/main/res/values/forms.xml
lramati: Yep, probably forgot to change the object name when copy pasting things around

app/build.gradle
lramati: I have no idea whats going on here, so I guess do whatever you need to make it work. This is all old stuff from the original PR, and I've not really gotten a grasp on Gradle yet.

Are you recommending we use gson or switch to json? I'm having a bit of trouble parsing your comment 😅

xZise: I wanted to say, that this project already uses GSON which is also a JSON parser. So I'm wondering why it is also requiring JSON. Maybe I'm wrong and they don't do the same.

Additionally it didn't compile for me because it was a duplicated requirement. I'll try to post the exact message later.

lramati:
Will continue work on xZise's branch

PR #1095, opened 2021-01-27
Version 5.5.2!

lramati:
* Mega Energy, XL Candy support, including new layout
* Lv50 support
* Gen6 pokemon
* re-targeted to Android 10 to allow continued Play Store updates

Unfortunately PokeSpam had to be disabled and PokeBattler is still horribly out of date. Recalibration is also extremely finicky right now, hoping to fix that in v5.5.3


lramati:
Contains and closes PR #1089

lramati:
I think the build failed due to mismatched gradle versions?

